PR #8, opened 2019-07-31
fix(eslint-plugin): expose new rules and use type guards

JamesHenry:
Hey @sfabriece I'm sorry I forgot on my PR reviews to point out that the new rules need to be exported in order to be used, so they need to be added to: `packages/eslint-plugin/src/index.ts`

I also did a couple of minor refactorings that were not worth blocking the original PRs over - they now use the type guards available in utils rather than type assertions where possible


sfabriece:
I'll keep that in mind for future PRs, although I might some type guards sometimes. @JamesHenry

PR #12, opened 2019-08-01
feat(eslint-plugin): add contextual-lifecyle rule

wKoza:


packages/eslint-plugin/src/utils/utils.ts
```diff
@@ -173,9 +190,8 @@ export const getDecorator = (
   return node.decorators.find(
     decorator =>
       isCallExpression(decorator.expression) &&
-      decorator.expression.arguments &&
-      decorator.expression.arguments.length > 0 &&
-      getDecoratorName(decorator) === decoratorName,
+      isIdentifier(decorator.expression.callee) &&
```
wKoza: A decorator may not have parameters ( `@Injectable()` for instance)

packages/eslint-plugin/src/rules/contextual-lifecycle.ts
```diff
@@ -0,0 +1,103 @@
...
+    },
+    schema: [],
+    messages: {
+      contextuaLifecycle: `The method {{methodName}} is not allowed for class {{className}} because it is decorated with {{decoratorName}}`,
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      ClassDeclaration(node: TSESTree.ClassDeclaration) {
```
JamesHenry: Rather than accessing a single node in the tree and then manually asserting things based on what "kind" it is in a single handler function, usually we want to leverage the fact that we can provide any number of node selectors to ESLint and it will call the dedicated function appropriately:

```ts
import { PIPE_CLASS_DECORATOR, DIRECTIVE_CLASS_DECORATOR } from '../utils/selectors';

// ...etc

return {
  [PIPE_CLASS_DECORATOR]: pipeHandler,
  [DIRECTIVE_CLASS_DECORATOR]: directiveHandler,
  // ... etc
```

They can of course share any number of functions between them so code does not need to be repeated any more than it does in your example

packages/eslint-plugin/src/rules/contextual-lifecycle.ts
```diff
@@ -0,0 +1,87 @@
...
+            methodName,
+            className,
+            decorator,
+          },
+        });
+      }
+    }
+
+    return {
+      [PIPE_CLASS_DECORATOR](pipeHandler: TSESTree.Decorator) {
```
JamesHenry: I think maybe just a copy paste issue - the whole function is the "handler", what you have called handlers here in each case are just nodes.

So L73 + L74 would be:

```ts
[PIPE_CLASS_DECORATOR](node: TSESTree.Decorator) {
  checkContext(node, AngularClassDecorators.Pipe);
```

The same for each example below

JamesHenry:
Nice! It's awesome how clean the implementation became after making those changes! Just the one minor naming point left I think

JamesHenry:
Thanks @wKoza!

PR #22, opened 2019-10-06
feat(eslint-plugin): add no-lifecycle-call rule

rafaelss95:


packages/eslint-plugin/src/rules/no-lifecycle-call.ts
```diff
@@ -0,0 +1,46 @@
+import { createESLintRule } from '../utils/create-eslint-rule';
+import { isAngularLifecycleMethod, isIdentifier } from '../utils/utils';
+
+type Options = [];
+export type MessageIds = 'noLifecycleCall';
+export const RULE_NAME = 'no-lifecycle-call';
+
+export default createESLintRule<Options, MessageIds>({
+  create(context) {
```
JamesHenry: Sorry that it's a fairly trivial thing, but in all other cases the `create()` method comes last, so that all the metadata is at the top, please could we follow that here as well?

packages/eslint-plugin/tests/rules/no-lifecycle-call.test.ts
```diff
@@ -0,0 +1,671 @@
...
+      annotatedSource: `@Component({
+        selector: 'test'
+      })
+      class Test {
+        test() {
+          this.ngAfterContentChecked();
+          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+        }
+      }`,
+      description: 'it should fail if explicitly calling ngAfterContentChecked',
```
JamesHenry: Similarly could we keep the `description`s as the first things here so that it is more readable overall?

rafaelss95: Oops, sry. I've committed the requested changes.

<h6>I have a habit of follow `object-literal-sort-keys` rule :stuck_out_tongue_closed_eyes:</h6>

rafaelss95:
Friendly bump @JamesHenry :)

PR #27, opened 2019-12-11
fix(builder): remove `rule` section before running CLI engine

Cammisuli:
Some users of nrwl/nx (nrwl/nx#2051) are requesting that `semi` not be a hard error while using eslint. 

This PR removes that rule before running the CLI Engine


JamesHenry:
D'oh! 😄 

I think I must have had this in there to prove inline rules worked and then just forgot to remove it.

Thanks a lot for finding and fixing!

JamesHenry:
I published 0.0.1-alpha.18 with this included

PR #31, opened 2020-01-04
test: use codecov bash uploader instead of js uploader

JamesHenry:
Copied from: https://github.com/typescript-eslint/typescript-eslint/pull/1353

The js uploader keeps on timing out which causes coverage to not report, causing PR status to get stuck in pending.

The bash reporter has a retry mechanism, so it is supposed to be be more resilient.


JamesHenry:
All green, looks promising

PR #34, opened 2020-01-20
Include install instructions

BO41:
is this the correct way of including angular-eslint?

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,11 +43,25 @@ If you are looking for general help in migrating from TSLint to ESLint, you can 
 
 For Angular project's specifically, the migration involves a few different aspects:
 
-1. Replacing the builder the Angular CLI will use when you run `ng lint`
+1. Installing dev-dependencies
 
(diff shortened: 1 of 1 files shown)


BO41:
currently it is not possible to use `angular-eslint` with `typescript-eslint` (with required type checking), right?
Otherwise it would be good if it was documented as well

JamesHenry:
Thanks a lot for this, I’ll take a look this weekend!

JamesHenry:
Thanks again @BO41!

Check out the new Angular CLI integration test for how things are setup and configured in a way that matches the existing TSLint setup.

> currently it is not possible to use angular-eslint with typescript-eslint (with required type checking), right?

You can see how the two projects are used together in that integration test - you absolutely do want and need to use them together, angular-eslint builds on top of typescript-eslint and couldn't exist without it.

BO41:
awesome, will take a look at it.
Thanks :)

PR #39, opened 2020-02-14
feat(eslint-plugin): add component-max-inline-declarations rule

JamesHenry:


JamesHenry:
Thanks again @kimamula!

PR #46, opened 2020-03-02
fix(eslint-plugin-template): missing rule export "no-negated-async" 

### LINKED ISSUE #45: "no-negated-async" rule is not exported
Hi, it seems the `no-negated-async` rule is not being exported from `eslint-plugin-template`. This causes the following error:

> Definition for rule '@angular-eslint/template/no-negated-async' was not found.eslint(@angular-eslint/template/no-negated-async)

It appears the fix could be to just add the export to `index.ts`

```diff
  import processors from './processors';
  import bananaInABox, {
    RULE_NAME as bananaInABoxRuleName,
  } from './rules/banana-in-a-box';
+ import noNegatedAsync, {
+  RULE_NAME as noNegatedAsyncRuleName
+ } from './rules/no-negated-async';

  export default {
    processors,
    rules: {
      [bananaInABoxRuleName]: bananaInABox,
+     [noNegatedAsyncRuleName]: noNegatedAsync
    },
  };
```

I've rebuilt the project with the above additions and the error is gone.

**Edit**: Link to file https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/index.ts

BO41:
Fixes #45 , https://github.com/angular-eslint/angular-eslint/commit/a7fc74ad01c0796adb2456e7c49532e7536b0f2b#commitcomment-37537377

### DIFF
--- a/packages/eslint-plugin-template/src/index.ts
+++ b/packages/eslint-plugin-template/src/index.ts
@@ -2,10 +2,14 @@ import processors from './processors';
 import bananaInABox, {
   RULE_NAME as bananaInABoxRuleName,
 } from './rules/banana-in-a-box';
+import noNegatedAsync, {
+  RULE_NAME as noNegatedAsyncRuleName
+} from './rules/no-negated-async';
 
 export default {
   processors,
   rules: {
     [bananaInABoxRuleName]: bananaInABox,
+    [noNegatedAsyncRuleName]: noNegatedAsync,
   },
 };
--- a/packages/eslint-plugin-template/src/index.ts
+++ b/packages/eslint-plugin-template/src/index.ts
@@ -3,7 +3,7 @@ import bananaInABox, {
   RULE_NAME as bananaInABoxRuleName,
 } from './rules/banana-in-a-box';
 import noNegatedAsync, {
-  RULE_NAME as noNegatedAsyncRuleName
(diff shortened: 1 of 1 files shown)


davelsan:
Hey, thanks for the PR! That was fast.

I got a question: Is there a particular reason for adding the trailing colon to the last named import (or the only one in this case)? A style guide that I haven't seen maybe?

If so, why not add it to the last element in the `rules` object too? I'm not nitpicking, just genuinely curious, since you pushed an extra commit just for that.

BO41:
Like many JS/TS projects, this one uses [prettier](https://github.com/prettier/prettier) as an code formatter. The [config file](https://github.com/angular-eslint/angular-eslint/blob/master/.prettierrc#L3) states to add a [trailing comma](https://prettier.io/docs/en/options.html#trailing-commas) everywhere.   
This is often preferred for a less noisy code diff.

The CI did fail on the prettier check, so I added the missing colon.

>why not add it to the last element in the `rules` object too?

Isn't there one too? Or what do you mean.

davelsan:
Ah nice, that makes sense. Thanks for taking the time to answer, I know it was a nooby question.

> Isn't there one too? Or what do you mean?

Yes, there is one there too. My bad.

PR #47, opened 2020-03-14
fix(*): improve support for inline and .html templates

JamesHenry:
See the updated integration test eslint config and the updated snapshot from the integration test to understand the improvement.

Builder still needs some more refactoring now that the usage is becoming much clearer, but this gets us to a great milestone


JamesHenry:
🤔build didn't trigger

JamesHenry:
Triggered build manually from the Azure DevOps side, no idea what the issue is...

PR #50, opened 2020-03-14
fix(eslint-plugin): add missing rule exports

### LINKED ISSUE #49: Add missing rules to index
The following rules have already been completed but are missing from the index:

- directive-class-suffix
- no-forward-ref
- relative-url-prefix
- use-injectable-provided-in

I will add these and update the readme.

nzacca:
This PR adds missing rule exports for the following rules:

- directive-class-suffix
- no-forward-ref
- relative-url-prefix
- use-injectable-provided-in

Fixes #49


nzacca:
Closing. I'll redo this so that the commit history isn't so dirty.

PR #55, opened 2020-03-16
feat(eslint-plugin): add no-input-prefix rule

paulotokimatu:
Two small changes from the rule on Codelyzer:
- To be consistent with the message on the other rules, the error message is:
`@Inputs should not be prefixed by foo, bar` instead of `@Inputs should not be prefixed by "foo", or "bar"`
- The prefixes are configured on a `prefixes` object:
```
        "@angular-eslint/no-input-prefix": ["error", {
            prefixes: ['test'],
        }],
```

Instead of being an array like on Codelyzer.

Please tell me if I should change something. This is my first PR so I am not sure of the preferred ways to do things here. 😄 

**References**
https://github.com/mgechev/codelyzer/blob/master/docs/rules/no-input-prefix/index.html
https://github.com/mgechev/codelyzer/blob/master/src/noInputPrefixRule.ts

---

Also, I think we may have an issue with Windows development. When I run `yarn format:check` on my Win10 I get that the *.json and *.md files are not formatted correctly.

My VSCode says that these files have the line ending as `CRLF`, and after running `yarn format` they change to `LF`.

I skipped this validation to be able to push, but should I run prettier on my machine and commit the changes to the repo?

### DIFF
--- a/packages/eslint-plugin/src/rules/no-input-prefix.ts
+++ b/packages/eslint-plugin/src/rules/no-input-prefix.ts
@@ -1,0 +1,85 @@
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { createESLintRule } from '../utils/create-eslint-rule';
+import { getClassPropertyName, isImportedFrom } from '../utils/utils';
+
+type Options = [
+  {
+    prefixes: string[];
+  },
+];
+export type MessageIds = 'noInputPrefix';
+export const RULE_NAME = 'no-input-prefix';
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Input names should not be prefixed by the configured disallowed prefixes.',
+      category: 'Best Practices',
+      recommended: false,
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          prefixes: {
+            type: 'array',
+            items: {
+              type: 'string',
+            },
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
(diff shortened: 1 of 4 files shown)


JamesHenry:
Thanks a lot for this! It's looking good.

> Also, I think we may have an issue with Windows development. When I run yarn format:check on my Win10 I get that the *.json and *.md files are not formatted correctly.

> My VSCode says that these files have the line ending as CRLF, and after running yarn format they change to LF.

That's surprising, we've had the gitattributes set since the very beginning... Maybe you could run the format and commit the result and I can see what it's like when I pull it down on my Mac?

packages/eslint-plugin/src/rules/no-input-prefix.ts
```diff
@@ -0,0 +1,85 @@
...
+            items: {
+              type: 'string',
+            },
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      noInputPrefix: `@Inputs should not be prefixed by {{blacklistedPrefixes}}`,
```
JamesHenry: Please could we change this language to match what you say in the feedback - i.e. use `disallowed` everywhere, instead of `blacklist`. I know some people feel strongly about this terminology and it's very easy to change to match with the feedback language so we may as well

paulotokimatu: Sounds good, thanks!

paulotokimatu:
> That's surprising, we've had the gitattributes set since the very beginning... Maybe you could run the format and commit the result and I can see what it's like when I pull it down on my Mac?

Hmm, sorry I think it was something wrong locally. When I tried to commit the results of `yarn format` git told that there was nothing to commit and the issue was solved and the checks now pass without problems.

PR #64, opened 2020-03-28
feat(eslint-plugin-template): add cyclomatic-complexity rule

paulotokimatu:


packages/integration-tests/fixtures/angular-cli-workspace/.eslintrc.js
```diff
@@ -193,6 +193,9 @@ module.exports = {
 
         // ORIGINAL tslint.json -> "template-no-negated-async": true,
         '@angular-eslint/template/no-negated-async': 'error',
+
+        // ORIGINAL tslint.json -> "template-cyclomatic-complexity": false,
```
JamesHenry: Was this the case? I thought I had captured all the rules that were included in the original tslint.json...

JamesHenry:
Great stuff thanks, the template piece was very complex to figure out and I'm delighted it's in a place now where others can contribute!

packages/integration-tests/fixtures/angular-cli-workspace/.eslintrc.js
paulotokimatu: You are right, there is just no mention of the rule. I will remove it.

paulotokimatu:
> 
> 
> Great stuff thanks, the template piece was very complex to figure out and I'm delighted it's in a place now where others can contribute!

Yes, when you get a bit familiar it gets straightforward to work with, great work with the template parsing!

PR #65, opened 2020-04-02
feat(eslint-plugin-template): add no-call-expression rule

paulotokimatu:


paulotokimatu:
@JamesHenry I am having an issue regarding the start position of a method call.

If I have a template with:
```
<div>{{ getInfo1() }}</div>
```

I get the correct position:
`1:9  error  Avoid calling expressions in templates  @angular-eslint/template/no-call-expression`

But if the interpolation is on a new line, like:
```
<div>
  {{ getInfo2() }}
</div>
```

I get:
`2:7  error  Avoid calling expressions in templates  @angular-eslint/template/no-call-expression`

But it should actually be `2:6`, right? Can you point me where can I investigate that?

paulotokimatu:
@JamesHenry Ping 😄

JamesHenry:
Have you attempted to see what the underlying angular compiler thinks? It's possible this is just how it is currently applying the range/node information...

paulotokimatu:
> 
> 
> Have you attempted to see what the underlying angular compiler thinks? It's possible this is just how it is currently applying the range/node information...

Yeah you are right, looks like it is just the way it comes from Angular compiler's `parseTemplate` method.

JamesHenry:
Cool, @paulotokimatu would you like to take it up with the Angular team, or press ahead with the PR for now?

paulotokimatu:
> 
> 
> Cool, @paulotokimatu would you like to take it up with the Angular team, or press ahead with the PR for now?

I will try taking a better look at that later, but anyway that would be outside the scope of this PR, so we can go ahead I guess. 😄

JamesHenry:
Thanks!

Issue #60, opened 2020-03-17
Crash when running ng lint


dieend:
version: "0.0.1-alpha.23"

```
> cat /private/var/folders/qx/l6dzbht15fj_s89bxrv7x7740000gp/T/ng-JaEvnr/angular-errors.log

[error] TypeError: Cannot read property 'range' of undefined
    at /path/to/angular_project/node_modules/@angular-eslint/eslint-plugin-template/dist/index.js:1:1353
    at Array.map (<anonymous>)
    at postprocess (/path/to/angular_project/node_modules/@angular-eslint/eslint-plugin-template/dist/index.js:1:1156)
    at Linter._verifyWithProcessor (/path/to/angular_project/node_modules/eslint/lib/linter/linter.js:1311:16)
    at Linter._verifyWithConfigArray (/path/to/angular_project/node_modules/eslint/lib/linter/linter.js:1248:25)
    at Linter.verify (/path/to/angular_project/node_modules/eslint/lib/linter/linter.js:1210:25)
    at Linter.verifyAndFix (/path/to/angular_project/node_modules/eslint/lib/linter/linter.js:1400:29)
    at verifyText (/path/to/angular_project/node_modules/eslint/lib/cli-engine/cli-engine.js:230:48)
    at CLIEngine.executeOnFiles (/path/to/angular_project/node_modules/eslint/lib/cli-engine/cli-engine.js:798:28)
    at _lint (/path/to/angular_project/node_modules/@angular-eslint/builder/dist/index.js:1:4477)
```

paulotokimatu:
Hello, I tried to do a quick reproduction but didn't get this error. Can you please give us more details like the Angular version, the content of `.eslintrc.js` (or equivalent) and `angular.json`?

Thanks.

dieend:
Hi, this crash was happened when I included the following config in eslintrc.js (copied from https://github.com/angular-eslint/angular-eslint/blob/master/packages/integration-tests/fixtures/angular-cli-workspace/.eslintrc.js)
```
{
    override: [
        ....
        {
            files: ['*.component.ts'],
            parser: '@typescript-eslint/parser',
            parserOptions: {
                ecmaVersion: 2020,
                sourceType: 'module',
            },
            plugins: ['@angular-eslint/template'],
            processor: '@angular-eslint/template/extract-inline-html',
        },
    ]
}
```
[...]

PR #69, opened 2020-04-17
fix(eslint-plugin): improve regular expression

### LINKED ISSUE #68: extract-inline-html parser can hang due to regexp catastrophic backtracking
If you configure an override for `*.component.ts` files to use the `extract-inline-html` parser like so:
```
{
    "files": [
        "*.component.ts"
    ],
    "parser": "@typescript-eslint/parser",
    "parserOptions": {
        "sourceType": "module"
    },
    "plugins": [
        "@angular-eslint/template"
    ],
    "processor": "@angular-eslint/template/extract-inline-html"
}
```

And if it encounters a component with some poor formatting (try adding a couple spaces or a tab before the ending of the `@Component` attribute like this):
```
@Component({
    selector: 'widget',
    templateUrl: './widget.component.html',
    styleUrls: ['./widget.component.scss']
  }) <-- space before the curly brace here
```

The linting will never complete because the regex [here](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/processors.ts#L4) hangs.  So unfortunately the `ng lint` execution will just sit there without any sort of error ever occurring.

bbarry:
fixes #68

Attempting to match TS with a regex is a folly, but this should perform better.

### DIFF
--- a/packages/eslint-plugin-template/src/processors.ts
+++ b/packages/eslint-plugin-template/src/processors.ts
@@ -1,7 +1,7 @@
 import ts from 'typescript';
 
 function quickExtractComponentDecorator(text: string) {
-  const matches = text.match(/@Component\({(\s.*\s)*}\)/);
+  const matches = text.match(/@Component\s*\(\s*{(?:(?!}\s*\))(?:\s|.))*}\s*\)/);
   if (!matches || !matches.length) {
     return null;
   }


JamesHenry:
Thank you for contributing, please see https://github.com/angular-eslint/angular-eslint/issues/60#issuecomment-612591044

I have a PR in the works to update all of this original quick POC work

PR #72, opened 2020-04-20
feat(eslint-plugin): no-attribute-decorator rule

GavinWu1991:
Implement no-attribute-decorator rule behavior of codelyzer.


packages/eslint-plugin/tests/rules/no-attribute-decorator.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    `
+      class Test {
+        constructor(@Optional() foo: any, @Optional() bar: any) {}
+      }
+    `,
+  ],
+  invalid: [
+    {
+      // should fail if constructor has one parameter with @Attribute decorator
+      code: `
```
JamesHenry, line 55: Please use the `convertAnnotatedSourceToFailureCase()` utility to use the visual error annotations as has been done on the other rules

GavinWu1991: Got it, I made the updates for the tests, please help review again. 
I suppose the `convertAnnotatedSourceToFailureCase` method only supports the scenario just have one error place, is that right?

paulotokimatu: Maybe you mean something like this?
https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/tests/rules/cyclomatic-complexity.test.ts#L120

GavinWu1991: @paulotokimatu Cool, thanks for your information.

GavinWu1991: @JamesHenry I have duplicated the method implementation of `convertAnnotatedSourceToFailureCase` from `eslint-plugin-template/tests` to `eslint-plugin/tests` to ensure the tests in the plugin folder also can take advantage of enhanced `convertAnnotatedSourceToFailureCase` method, but I'm wondering if we can extract this method to a shared folder so that no need to worry about the future refactoring of this method and duplicate it again.

JamesHenry: Good call out, thanks! I consolidated things into a utils package in #76, you can pull in that update and use the shared utils now

GavinWu1991: @JamesHenry Got it, I pulled your commit and merge to my branch. I appreciate it if you can continue to review the PR again. 👍

packages/eslint-plugin/tests/rules/no-attribute-decorator.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    }),
+
+    {
+      // should fail if constructor has multiple parameters and all with @Attribute decorator
+      code: `
+      class Test {
+        constructor(@Attribute() foo: any, @Attribute() bar: any) {}
+      }
+    `,
+      errors: [
```
JamesHenry, line 111: Please convert to use the annotations

packages/eslint-plugin/tests/rules/no-attribute-decorator.test.ts
JamesHenry: Hmm they still haven't been converted to use the multiple annotations though...

packages/eslint-plugin/tests/rules/no-attribute-decorator.test.ts
GavinWu1991: Sorry, I pushed the changes just a moment.

PR #116, opened 2020-08-14
feat(eslint-plugin): no-empty-lifecycle-method

### LINKED ISSUE #104: [rule-request]: Warn on empty lifecycle hooks
Empty lifecycle hooks can have some performance implications without bringing any benefits. Since the runtime will process them we'll get:

- Slower processing time
- Larger memory consumption

```ts
@Component({... })
export class AppComponent implements OnInit {
  // This should warn
  ngOnInit() {}
}
```

andrefilimono:
Closes #104 

The rule checks only classes having Directive and Component decorators, as according to Angular's documentation only those classes can have lifecycle methods.

Looking forward to your feedback, as it's my first try in writing custom rules and maybe I forgot some cases or could do something better.


JamesHenry (approved):
Thank you!

JamesHenry:
This is available in `v0.1.0-beta.1`

PR #128, opened 2020-09-08
feat(eslint-plugin-template): accessibility-tabindex-no-positive

andrefilimono:
Rewritten version of the `template-accessibility-tabindex-no-positive` rule from Codelyzer.

Looking forward to hearing your feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ eslint --ext .ts,.component.html
 | [`template-accessibility-alt-text`]             |                    |
 | [`template-accessibility-elements-content`]     |                    |
 | [`template-accessibility-label-for`]            |                    |
-| [`template-accessibility-tabindex-no-positive`] |                    |
+| [`template-accessibility-tabindex-no-positive`] | :white_check_mark: |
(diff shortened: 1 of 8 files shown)


JamesHenry (changes requested):
Thanks so much!

Given we have this chance for fresh perspective, I think we could rename this to just: “no-positive-tabindex”

- having the no- first is more in keeping with eslint conventions
- the fact this rule exists due to accessibility concerns can be expressed through documentation, I don’t think it needs to be in the rule name

WDYT?

andrefilimono:
> Thanks so much!
> 
> Given we have this chance for fresh perspective, I think we could rename this to just: “no-positive-tabindex”
> 
> * having the no- first is more in keeping with eslint conventions
> * the fact this rule exists due to accessibility concerns can be expressed through documentation, I don’t think it needs to be in the rule name
> 
> WDYT?

Yeah, that's definitely a better name for the rule. Thanks!

JamesHenry (approved):

JamesHenry:
Thanks again for this!

Issue #138, opened 2020-09-21
@angular-eslint/schematics creates an invalid .eslintrc.json


kuzivany:
The `.eslintrc.json` created by schematics is an invalid JSON file. It contains a trailing semicolon as seen [in the template](https://github.com/angular-eslint/angular-eslint/blob/0d7746a9a17edb8a6623b23b111443725db5e5bb/packages/schematics/src/add-config-to-project/files/root-files/.eslintrc.json__tmpl__#L105)

JamesHenry:
Thanks! Would you like to submit a quick fix PR?

kuzivany:
> Thanks! Would you like to submit a quick fix PR?

Cool, let me get one done

PR #139, opened 2020-09-21
fix(schematics): create valid .eslintrc.json

kuzivany:
Create valid JSON .eslintrc.json file for #138


JamesHenry (approved):
Thanks a lot!

Issue #78, opened 2020-04-29
Enhancement: Add ability to hide warnings from the output (a quiet option)


mikg2003:
Eslint provides a `--quiet` option that results in only errors being output.  I'm sure some will say that you should just fix your warnings, but when conditions like "having ignored files" result in warnings, it'd be awfully nice to be able to hide those when looking for the error that's preventing your build from succeeding.

JamesHenry:
I would happily accept a PR for this!

PR #140, opened 2020-09-26
feat(builder): quiet option support

AlexanderFSP:
Resolve https://github.com/angular-eslint/angular-eslint/issues/78


packages/builder/tests/index.test.ts
```diff
@@ -326,6 +343,70 @@ describe('Linter Builder', () => {
...
+        {
+          errorCount: 0,
+          warningCount: 4,
+          results: [],
+          messages: [
+            {
+              line: 0,
+              column: 0,
+              message: 'Mock warning message',
+              severity: 2,
```
JamesHenry, line 358: Something seems wrong here, severity `2` is an error, not a warning

JamesHenry (changes requested):

packages/builder/tests/index.test.ts
AlexanderFSP: Yeah, the issue was that `mockFormatter.format()` returned string[] (`[ 'Mock warning message' ]`) but in the test, it compared with the string literal (`'Mock warning message'`). It's fixed now.

JamesHenry (approved):

JamesHenry:
Thanks!

PR #145, opened 2020-10-01
fix(eslint-plugin): update no-host-metadata-property STYLE_GUIDE_URL

### LINKED ISSUE #143: no-host-metadata-property has incorrect STYLE_GUIDE_URL
Currently links to https://angular.io/styleguide#style-05-12 (Decorate input and output properties)

Should link to https://angular.io/styleguide#style-06-03 (HostListener/HostBinding decorators versus host metadata)

MikeMatusz:
fixes #143


JamesHenry:
Thanks @MikeMatusz, you need to update the integration tests accordingly

JamesHenry (changes requested):
Good that this came up because it has shone a spotlight on the fact that the error is also wrong.

Please update it to match: https://github.com/mgechev/codelyzer/blob/master/src/noHostMetadataPropertyRule.ts

It should not be referencing `@Output` like it does currently

MikeMatusz:
Updated now.  On a side note, would you be open to a PR making a couple modifications making the repo friendlier to develop on Windows?  I can open a separate issue to discuss it further, but it would just be using something like [shx](https://www.npmjs.com/package/shx) to handle commands like `rm` and `cp`, and a minor change to `integration.ts` to support backslashes in paths.

JamesHenry:
@MikeMatusz yeah if you have thoughts on cross-platform support now is the time to discuss it because https://github.com/angular-eslint/angular-eslint/pull/146 has been opened

Please add comments on that PR if you have alternative proposals in mind

JamesHenry:
@MikeMatusz You said `Updated now`, but there has not been an update?

JamesHenry (approved):

JamesHenry:
Thanks!

PR #152, opened 2020-10-05
fix(contextual-lifecycle): fix uninterpolated decorator name in message

mijelin:
Fix message for the `contextual-lifecycle` rule so that the decorator name is properly interpolated.

For example, for the following code:
```ts
@Injectable()
class Test {
  ngOnInit() {}
}
```

Before:
```
The method ngOnInit is not allowed for class Test because it is decorated with {{decoratorName}}
```

After:
```
The method ngOnInit is not allowed for class Test because it is decorated with Injectable
```


JamesHenry (approved):
Thanks a lot!

Issue #151, opened 2020-10-05
An unhandled exception occurred: Cannot read property 'range' of undefined


AlexanderFSP:
**Prerequisites:**
I have generated a new angular 10.1 app and perform migration to ESLint using `@angular-eslint/schematics:add-config-to-project` schematic. And also I change a bit `.eslintrc.json` to use  `plugin:@angular-eslint/template/all` config instead of recommended.

This is my pretty simple AppComponent:
```
import { Component } from '@angular/core';

@Component({
  selector: 'app-root',
  template: `<input type="text" tabindex="20" />`,
  styleUrls: ['./app.component.css'],
})
export class AppComponent {}
```
**Problem:**
And when I try to lint my code I get an error:
1. In the console, when I execute `lint` npm run script:
`An unhandled exception occurred: Cannot read property 'range' of undefined`
2. Also in VSCode output console (using `dbaeumer.vscode-eslint` extension):
```
[Error - 17:26:22] ESLint stack trace:
[Error - 17:26:22] TypeError: Cannot read property 'range' of undefined
    at C:\GitHub\ng10\node_modules\@angular-eslint\eslint-plugin-template\dist\index.js:1:1953
    at Array.map (<anonymous>)
    at postprocess (C:\GitHub\ng10\node_modules\@angular-eslint\eslint-plugin-template\dist\index.js:1:1762)
    at Linter._verifyWithProcessor (C:\GitHub\ng10\node_modules\eslint\lib\linter\linter.js:1327:16)
    at Linter._verifyWithConfigArray (C:\GitHub\ng10\node_modules\eslint\lib\linter\linter.js:1264:25)
    at Linter.verify (C:\GitHub\ng10\node_modules\eslint\lib\linter\linter.js:1226:25)
    at Linter.verifyAndFix (C:\GitHub\ng10\node_modules\eslint\lib\linter\linter.js:1416:29)
    at verifyText (C:\GitHub\ng10\node_modules\eslint\lib\cli-engine\cli-engine.js:224:48)
    at CLIEngine.executeOnText (C:\GitHub\ng10\node_modules\eslint\lib\cli-engine\cli-engine.js:882:26)
    at c:\Users\root\.vscode\extensions\dbaeumer.vscode-eslint-2.1.8\server\out\eslintServer.js:1:40795
```

This issue happens right here due to `no-positive-tabindex` rule isn't an auto-fixable => the passed message hasn't `fix` property. So, additional check needed here.
[...]

PR #155, opened 2020-10-07
fix(eslint-plugin-template): handle optional fix prop in postprocessor, fixed preprocessor return type

AlexanderFSP:
Resolve https://github.com/angular-eslint/angular-eslint/issues/151

### DIFF
--- a/packages/eslint-plugin-template/src/processors.ts
+++ b/packages/eslint-plugin-template/src/processors.ts
@@ -149,7 +149,7 @@ export function preprocessComponentFile(text: string, filename: string) {
 export function postprocessComponentFile(
   multiDimensionalMessages: any[][],
   filename: string,
-) {
+): string[] {
   const messagesFromComponentSource = multiDimensionalMessages[0];
   const messagesFromInlineTemplateHTML = multiDimensionalMessages[1];
   /**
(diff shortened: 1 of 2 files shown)


JamesHenry (changes requested):
Please revert all the unnecessary renaming of things in the tests, it has made things less easy to scan and know what is coming from the current test case vs what is a fixed variable, and just in general it's not great for reviewing PRs to change more than is in scope

JamesHenry:
Sorry for not being clear - you made some functional changes around using a consistent interface for what we return for the processor, that was fine, and did require a small change to the tests, but you can had renamed and refactored a load of other parts of the test code which were not required to get them to pass. It was the latter point I was requesting to revert, not the entire thing.

If you want to bring that change back (without the refactoring) that's absolutely fine, just let me know either way how you wish to proceed with this PR

AlexanderFSP:
> Sorry for not being clear - you made some functional changes around using a consistent interface for what we return for the processor, that was fine, and did require a small change to the tests, but you can had renamed and refactored a load of other parts of the test code which were not required to get them to pass. It was the latter point I was requesting to revert, not the entire thing.
> 
> If you want to bring that change back (without the refactoring) that's absolutely fine, just let me know either way how you wish to proceed with this PR

So, I've rolled back changes for `preprocessor`, carefully fixed tests & added basic tests to cover at least first if statement in `postprocessor`. To cover the rest functionality - I think refactoring also needed because there is no possibility to mock `rangeData` Map in an easy way...

JamesHenry (approved):

JamesHenry:
Great, thanks!

PR #168, opened 2020-10-22
fix(builder): check for quiet flag when logging results

### LINKED ISSUE #164: Blank lines printed when using "quiet" option
Hello,

If there are warnings from eslint and the quiet option is used, there will be blank lines printed instead of the warnings.

miked1090:
Fixes #164


JamesHenry:
Thanks @miked1090!

PR #172, opened 2020-11-07
chore(integration-tests): refactor test execution and fixture naming convention

JamesHenry:
- Uses a full local registry to publish packages to (verdaccio)
- This allows us to run an actual `ng add` as part of integration tests
- Applies naming convention to fixtures to make it clear the Angular CLI version and intended scope/methodology

These changes pave the way for more powerful fixtures and the upcoming project conversion schematic


JamesHenry:
I don't really understand why only the yarn verdaccio uplink is working, and not the npm one, but YOLO 🤷‍♂️

PR #183, opened 2020-11-17
fix(eslint-plugin): incorrect message for no-empty-lifecycle-method

### LINKED ISSUE #181: Wrong error message for no-empty-lifecycle-method
Came across this today, and I believe the error message it's displaying is wrong. Pretty sure the only place you can define a lifecycle method is inside a class. Perhaps `Declaring an empty ngOnInit lifecycle method is not recommended.` or `Lifecycle methods should not be empty.` as it's a best practice so saying `should not` translates that a bit better.

image

https://github.com/angular-eslint/angular-eslint/blob/50b244751e940fc25c98bbcbec8cafca4e7a8b3d/packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts#L26

### LINKED ISSUE #188: Unclear warning message for "no-empty-lifecycle-method"
The error for the `no-empty-lifecycle-method` rule is currently "Declaring {{methodName}} lifecycle method in a class is not recommended", but this is confusing, as I thought the rule being flagged was actually `no-conflicting-lifecycle`. The warning message should be something like "{{methodName}} lifecycle method should not be empty".

napei:
Closes #181 
Closes #188 

Changes the message for the `no-empty-lifecycle-method` rule from
`Declaring {{methodName}} lifecycle method in a class is not recommended`
to
`Lifecycle method {{methodName}} should not be empty`

### DIFF
--- a/packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts
+++ b/packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts
@@ -23,7 +23,7 @@ export default createESLintRule<Options, MessageIds>({
     },
     schema: [],
     messages: {
-      noEmptyLifecycleMethod: `Declaring {{methodName}} lifecycle method in a class is not recommended`,
+      noEmptyLifecycleMethod: `Lifecycle method {{methodName}} should not be empty`,
     },
   },
   defaultOptions: [],


JamesHenry:
Thanks, @napei!

JamesHenry:
Released in `0.8.0-beta.2`

PR #205, opened 2020-11-29
feat(eslint-plugin-template): no-any

rafaelss95:


packages/eslint-plugin-template/tests/rules/no-any.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description: 'it should fail with call expression in property binding',
+      annotatedSource: `
+      <a [href]="$any(getHref())">Click here</a>
+                 ~~~~~~~~~~~~~~~
+      `,
+      messageId,
+    }),
+    convertAnnotatedSourceToFailureCase({
```
rafaelss95, line 66: Hey @JamesHenry I was unable to make this test pass. I got this message:

> Message:
    Error line should be 4

 Do you have any clue?

JamesHenry: Have you tried using `parserServices.convertNodeSourceSpanToLoc()` like all the other template rules do?

rafaelss95: Hmm, not really. I followed the same idea of `no-call-expression` (which doesn't use this), but now you said I tested `parserServices.convertNodeSourceSpanToLoc` and I got the following result:

```ts
{
  start: { line: NaN, column: undefined },
  end: { line: NaN, column: undefined }
}

Comment (JamesHenry) [reply packages/eslint-plugin-template/tests/rules/no-any.test.ts:66]: Hmm I don't time to dig into this loads right now but I've noticed that if you wrap your middle assertion in a div, then it works.

Worth looking into the AST that the angular compiler actually produces for that source.

I added the div and fixed the alignment of the last issue by one char, that made the tests pass for me:

Comment (JamesHenry) [reply packages/eslint-plugin-template/tests/rules/no-any.test.ts:66]: Oh damn it wouldn't let me push to your branch...

File: packages/eslint-plugin-template/tests/rules/no-any.test.ts [hunk 2]
```diff
@@ -0,0 +1,92 @@
...
+      description: 'it should fail for multiple cases',
+      annotatedSource: `
+        {{ $any(framework).name }}
+            ~~~~~~~~~~~~~~~
+        {{ this.$any(framework).name }}
+            ^^^^^^^^^^^^^^^^^^^^
+        <a [href]="$any(getHref())">Click here</a>'
+                    ###############
```
JamesHenry, line 74:
```suggestion
            ~~~~~~~~~~~~~~~
       <div>
        {{ this.$any(framework).name }}
            ^^^^^^^^^^^^^^^^^^^^
        </div>
        <a [href]="$any(getHref())">Click here</a>'
                   ###############
```

packages/eslint-plugin-template/tests/rules/no-any.test.ts
JamesHenry: Added it as a suggestion below.

rafaelss95: Hey @JamesHenry sry for the delay, I've been busy these days. I think I'll be able to take a look at this tomorrow. Anyway it's really strange you couldn't push to my branch because I checked to "Allow edits and access [...]".

JamesHenry: Thanks @rafaelss95, for some reason looks like it worked via the github UI... 🤷‍♂️

JamesHenry: All green now. Ready for this to go in @rafaelss95?

I do think we need to figure out this "location when no parent element" issue, but I don't think it's specific to this rule so I'm happy to press ahead with this for now

rafaelss95: @JamesHenry sure, let's move on.

JamesHenry:
Thanks a lot!

Issue #185, opened 2020-11-17
multi-line inline-templates with CRLF line endings


jr01:
Linting fails on component files that have CRLF (eg. Windows) line endings.

Same issue as https://github.com/angular-eslint/angular-eslint/issues/162, but now with steps to reproduce (on Linux):
```
git clone https://github.com/angular-eslint/angular-eslint
cd angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config
npm install
ng add @angular-eslint/schematics
unix2dos src/app/example.component.ts
ng lint v1014-multi-project-manual-config
```

Shows error
```
Linting "v1014-multi-project-manual-config"...
Error: Debug Failure. False expression: position cannot precede the beginning of the file
    at computeLineOfPosition (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/typescript/lib/typescript.js:8934:22)
    at computeLineAndCharacterOfPosition (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/typescript/lib/typescript.js:8912:26)
    at Object.getLineAndCharacterOfPosition (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/typescript/lib/typescript.js:8953:16)
    at SourceFileObject.getLineAndCharacterOfPosition (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/typescript/lib/typescript.js:143364:23)
    at preprocess (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/@angular-eslint/eslint-plugin-template/dist/index.js:1:1430)
    at Linter._verifyWithProcessor (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/eslint/lib/linter/linter.js:1292:30)
    at Linter._verifyWithConfigArray (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/eslint/lib/linter/linter.js:1264:25)
    at Linter.verify (/home/me/repos/angular-eslint/packages/integration-tests/fixtures/v1014-multi-project-manual-config/node_modules/eslint/lib/linter/linter.js:1226:25)
[...]

PR #206, opened 2020-11-30
fix: components with inline templates and CRLF

jr01:
Fixes https://github.com/angular-eslint/angular-eslint/issues/185

> > I haven't been able to compile the full source yet
> 
> I'm not sure what you mean by this

I couldn't get `npm run build` in `/packages/eslint-plugin-template` working the other week for some reason... Today it worked and I manually verified that the compiled/build package with the fix works in my other project.


JamesHenry:
Thanks a lot @jr01!

PR #210, opened 2020-12-03
fix(eslint-plugin): directive-selector does not ignore indentation differences nor work with multiple selectors

### LINKED ISSUE #166: directive-selector does not ignore indentation differences nor work with multiple selectors
Using `{ type: 'attribute', prefix: ['app'], style: 'kebab-case' }`, this works:

```
@Directive({
  selector: `[app-foo-bar]`
})
```

but this does not, while it used to in the former Codelyzer rule:

```
@Directive({
  selector: `
    [app-foo-bar]
  `
})
```

This also used to work but no longer does, and is commonly used in "advanced" directives:

```
@Directive({
  selector: `
    [app-foo-bar],
    [app-foo-baz]
  `
})
```

Looking at the code in https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/utils/property-selector.ts#L165, the selector list is parsed with `CssSelector.parse()` but the result is ignored when checking for prefixes and the rule simply work at the raw string level, instead of working at the CSS AST one:

```
listSelectors = CssSelector.parse(node.quasis[0].value.raw);
hasExpectedPrefix = arrayify<string>(
  node.quasis[0].value.raw as string,
).some((selector) =>
  prefixOption.some((prefix) =>
    SelectorValidator.prefix(prefix, styleOption)(selector),
  ),
);
```

GavinWu1991:
fix #166

### DIFF
--- a/packages/eslint-plugin/src/utils/property-selector.ts
+++ b/packages/eslint-plugin/src/utils/property-selector.ts
@@ -162,6 +162,7 @@ export const checkSelector = (
   } else if (node && isTemplateLiteral(node) && node.quasis[0]) {
     listSelectors = CssSelector.parse(node.quasis[0].value.raw);
     hasExpectedPrefix = arrayify<string>(
+      // FIXME: @Gavin, use listSelector here
       node.quasis[0].value.raw as string,
     ).some((selector) =>
       prefixOption.some((prefix) =>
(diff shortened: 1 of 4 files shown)


JamesHenry (approved):
Thanks a lot!

PR #219, opened 2020-12-09
feat(eslint-plugin): import-destructuring-spacing

arturovt:


rafaelss95:
Hey @arturovt @JamesHenry @mgechev @wKoza I'm not sure if it worths the cost of maintaining this rule in this repo specifically because IMHO a formatter like Prettier, for example, can do it really better than a linter, but I'd want to hear you opinion on this also.

In any case, if it'd be implemented in a linter IMHO it should be added to https://github.com/typescript-eslint/typescript-eslint instead, as it's not direct related to Angular.

In fact, there's already an [ESLint rule for this](https://eslint.org/docs/rules/object-curly-spacing) and a [closed PR](https://github.com/typescript-eslint/typescript-eslint/pull/1884) in the repo mentioned above for reference.

JamesHenry:
I completely agree, @rafaelss95 

Thank you for taking the time to do this @arturovt, but I do agree that things like this are best left to a dedicated code formatter, and even if implemented as a lint rule, should not fall under the scope of angular-eslint.

Thanks again!

PR #221, opened 2020-12-09
feat(eslint-plugin-template): accessibility-valid-aria

arturovt:


packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -0,0 +1,60 @@
+const { aria } = require('aria-query');
```
JamesHenry, line 1: Please use an import (you'll need to install `@types/aria-query` as a devDep)

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -0,0 +1,60 @@
...
+      accessibilityValidAria:
+        '{{attribute}}: This attribute is an invalid ARIA attribute.',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const parserServices = getTemplateParserServices(context);
+
+    return {
+      // Note that we're using the `Element` visitor except of `TextAttribute`
```
JamesHenry, line 34: You need to update the visitor keys in the template-parser `index.ts`, `attributes` should be added to the existing `Element` definition, and any additional key types should be added there as and when required

JamesHenry (changes requested):

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
arturovt: Updated.

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
arturovt: Thanks, updated.

packages/template-parser/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { parseTemplate, ParseSourceSpan } from '@angular/compiler';
-import EventEmitter from 'events';
```
JamesHenry, line 2: Hmm why did this change?

arturovt: Umm this was an intentional change because `yarn build` didn’t succeed. I will paste an error here when I get back to PC.

arturovt: @JamesHenry 

image

This happens when the `EventEmitter` is imported via the default import.

arturovt: I have reverted the change to `import EventEmitter`, this seems like `yarn` accidently updated `@types/node` in my workspace to the latest version, because the latest `@types/node` version doesn't allow to import `EventEmitter` via default import.

JamesHenry (approved):

JamesHenry:
Great, thanks @arturovt!

PR #222, opened 2020-12-10
feat(eslint-plugin-template): accessibility-table-scope

arturovt:


packages/eslint-plugin-template/src/rules/accessibility-table-scope.ts
```diff
@@ -0,0 +1,51 @@
...
+    messages: {
+      accessibilityTableScope: 'Scope attribute can only be on <th> element.',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const parserServices = getTemplateParserServices(context);
+
+    return {
+      Element(node: any) {
```
JamesHenry, line 29: Same feedback as previous PR regarding visitor keys

JamesHenry (changes requested):

packages/eslint-plugin-template/src/rules/accessibility-table-scope.ts
arturovt: Ok, will update it after the previous PR gets merged and rebasing.

JamesHenry: Great, thanks!

packages/eslint-plugin-template/tests/rules/accessibility-table-scope.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      annotatedSource: `
+        <div scope></div>
+             ~~~~~
+      `,
+    }),
+    convertAnnotatedSourceToFailureCase({
+      messageId,
+      description: 'should fail when element other than th has scope input',
+      annotatedSource: `
+        <div [attr.scope]="scope"></div>
```
arturovt, line 35: @JamesHenry the `BoundAttribute` visitor is not called for `attr.` bindings. Am I missing anything in the code?

JamesHenry: You’ll probably just need to update the visitor keys in the parser again right?

JamesHenry:
@arturovt Please try and avoid force pushing on an already open and under review PR, it makes it very difficult to track the history and how discussions and reviews relate to it.

The history is all cleaned up and squashed on merge anyway

arturovt:
> @arturovt Please try and avoid force pushing on an already open and under review PR, it makes it very difficult to track the history and how discussions and reviews relate to it.
> 
> The history is all cleaned up and squashed on merge anyway

This was a force push after rebase since I had to fix conflict.

JamesHenry:
You can fix conflicts as part of merge commits and not lose the history, rebasing is not required

arturovt:
> You can fix conflicts as part of merge commits and not lose the history, rebasing is not required

Okay, got it.

Regarding the `BoundAttribute`... Don't wanna bother you with this stuff but I've checked visitor keys and it seems like everything is there, btw the `BoundAttribute` visitor is not called only for `attr` and I can't seem to understand why :shrug:

JamesHenry:
Have you tried adding some logging around the AST that is produced for the specific example? Perhaps the AST is actually different to what we think it is/has changed since the visitor keys were last updated.

I don't have any specific advice I'm afraid, that's what I would be doing to try and figure it out myself

arturovt:
@JamesHenry I've found an issue, see my changes in the commit. Basically it worked in tests for property bindings e.g. `[binding]="myProperty"`, because the property binding type is `0` and `!c.type` returned `true`:
```ts
const enum BindingType {
  // A regular binding to a property (e.g. `[property]="expression"`).
  Property,
  // A binding to an element attribute (e.g. `[attr.name]="expression"`).
  Attribute,
}
```
And it didn't work for attribute and other binding types since `!1` returned false.

JamesHenry (approved):
Ok cool, thanks for catching that!

PR #225, opened 2020-12-10
feat(eslint-plugin-template): i18n missing feature ignore bound text

kmcs:
Hi,
once again a small feature for i18n checks, so that bound text could ignored if there is no text.
Sorry for the git history.. maybe someone have a link how do it the right way?
best,
Timo


JamesHenry:
I have to admit I'm not super familiar with the native i18n support in Angular but this seems fine 👍

PR #227, opened 2020-12-10
feat(eslint-plugin-template): accessibility-alt-text

arturovt:


packages/eslint-plugin-template/tests/rules/accessibility-alt-text.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    '<input type="image" alt="This is descriptive!">',
+    '<input type="image" aria-label="foo">',
+    '<input type="image" aria-labelledby="id1">',
+  ],
+  invalid: [
+    convertAnnotatedSourceToFailureCase({
+      messageId,
+      description: 'should fail image does not have alt text',
+      annotatedSource: `
+        <img src="foo">
```
arturovt, line 43: @JamesHenry 

Hey James. I've got a question. The end source span is **not** set for void elements. What is the proper way to get the _location_ of this node?

JamesHenry: I’d be curious if it was there if you added the / in front of the > denoting it is the closing tag of a void element...

In any case though questions about the AST the angular compiler produces for us kind of have to be posted on their repo. Most things I know about it have just been inferred by fiddling with and logging the output of things, and fortunately that’s been enough most of the time.

arturovt: I've added a feature to get the location of void elements. Could you have a look, please.

JamesHenry: Nice work, this seems reasonable to me!

PR #233, opened 2020-12-19
feat(eslint-plugin): contextual-decorator

arturovt:


packages/eslint-plugin/src/rules/contextual-decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+    schema: [],
+    messages: {
+      contextualDecorator:
+        'Decorator out of context for "@{{classDecoratorName}}()"',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      'MethodDefinition[kind="set"], MethodDefinition[kind="get"], MethodDefinition[kind="method"], ClassProperty, TSParameterProperty'(
```
rafaelss95, line 36:
```suggestion
      'MethodDefinition[kind=/^(get|set|method)$/], ClassProperty, TSParameterProperty'(
```

packages/eslint-plugin/src/rules/contextual-decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+});
+
+function validateNode(
+  context: TSESLint.RuleContext<MessageIds, []>,
+  node:
+    | TSESTree.MethodDefinition
+    | TSESTree.ClassProperty
+    | TSESTree.TSParameterProperty,
+): void {
+  if (!node.decorators || node.decorators.length === 0) {
```
rafaelss95, line 55:
```suggestion
  if (!node.decorators?.length) {
```

packages/eslint-plugin/src/rules/contextual-decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+    .filter(isNotNullOrUndefined)
+    .find(isAngularClassDecorator);
+}
+
+function validateDecorator(
+  context: TSESLint.RuleContext<MessageIds, []>,
+  decorator: TSESTree.Decorator,
+  classDecoratorName: AngularClassDecoratorKeys,
+): void {
+  const decoratorName: string | undefined = getDecoratorName(decorator);
```
rafaelss95, line 104:
```suggestion
  const decoratorName = getDecoratorName(decorator);
```

packages/eslint-plugin/src/rules/contextual-decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+    node = node.parent;
+  }
+
+  return null;
+}
+
+function getClassDecoratorName(
+  classDeclaration: TSESTree.ClassDeclaration,
+): AngularClassDecoratorKeys | undefined {
+  return (classDeclaration.decorators || [])
```
rafaelss95, line 93:
```suggestion
  return (classDeclaration.decorators ?? [])
```

rafaelss95 (approved):
LGTM with minor nits.

packages/eslint-plugin/src/rules/contextual-decorator.ts
arturovt: Thanks!

packages/eslint-plugin/src/rules/contextual-decorator.ts
arturovt: In that particular case `||` will do the same thing as we don't care about empty string and `0` number :smile: (according to ECMA scec).

arturovt:
@rafaelss95 I've updated the code per your review. Btw are these queries, that're used as visitor keys, coming from this package https://github.com/phenomnomnominal/tsquery?

packages/eslint-plugin/src/rules/contextual-decorator.ts
rafaelss95: Yeah, I prefer to be explicit but that's just a matter of preference and doesn't impact in anyway :)

rafaelss95 (dismissed):

rafaelss95:
> @rafaelss95 I've updated the code per your review. Btw are these queries, that're used as visitor keys, coming from this package https://github.com/phenomnomnominal/tsquery? (I mean `MethodDefinition[kind=/REGEX/], ...`, etc.)

@JamesHenry can have a better answer here, but I believe that ESLint consumes selector implementations from https://github.com/estools/esquery.

JamesHenry:
@arturovt As @rafaelss95 says, https://github.com/phenomnomnominal/tsquery is based on the library esquery which has been used within ESLint for years.

esquery works based on an ESTree-like AST (which is what ESLint uses), tsquery is designed to work on the TypeScript AST.

Even though we are dealing with TypeScript source code, we are using it in an ESLint context, so our selectors are based on an ESTree-like AST, rather than a TypeScript AST. This bridge between those two AST worlds is provided by `typescript-eslint`.

packages/eslint-plugin/src/rules/contextual-decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Ensures that classes use contextual decorators in its body.',
+      category: 'Best Practices',
+      recommended: 'error',
```
JamesHenry, line 25: This has been marked as recommended but not added to the recommended config, so it is currently failing a new check in CI.

If this was recommended in codelyzer I'm happy for it to be recommended in angular-eslint (and in general I don't want to be a dictator about recommended configs, I spend a lot of my time outside of an Angular context at the moment so am not always the best judge anyway)

JamesHenry (changes requested):

packages/eslint-plugin/src/rules/contextual-decorator.ts
arturovt: Not sure if it was recommended in codelyzer :smile: I've marked it as "false" for now.

arturovt:
> @arturovt As @rafaelss95 says, https://github.com/phenomnomnominal/tsquery is based on the library esquery which has been used within ESLint for years.
> 
> esquery works based on an ESTree-like AST (which is what ESLint uses), tsquery is designed to work on the TypeScript AST.
> 
> Even though we are dealing with TypeScript source code, we are using it in an ESLint context, so our selectors are based on an ESTree-like AST, rather than a TypeScript AST. This bridge between those two AST worlds is provided by `typescript-eslint`.

Thanks! I was trying to find some documentation on how to write those selectors "correctly" but wasn't successful :(

JamesHenry (approved):

PR #235, opened 2020-12-19
refactor(eslint-plugin-template): switch to query

arturovt:


packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -30,36 +29,31 @@ export default createESLintRule<Options, MessageIds>({
...
 
-    return parserServices.defineTemplateBodyVisitor({
-      BoundAttribute(attribute: any) {
-        validateAttribute(context, parserServices, attribute);
-      },
-      TextAttribute(attribute: any) {
-        validateAttribute(context, parserServices, attribute);
+    return {
+      'TextAttribute[name=/aria-*/], BoundAttribute[name=/aria-*/]'(
+        attribute: TmplAstTextAttribute | TmplAstBoundAttribute,
```
rafaelss95, line 34:
```suggestion
        { name, sourceSpan }: TmplAstTextAttribute | TmplAstBoundAttribute,
```

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -30,36 +29,31 @@ export default createESLintRule<Options, MessageIds>({
...
-  context.report({
-    loc,
-    messageId: 'accessibilityValidAria',
-    data: {
-      attribute: attribute.name,
-    },
-  });
+let ariaAttributes: Set<string> | null = null;
+function getAriaAttributes(): Set<string> {
+  return ariaAttributes || (ariaAttributes = new Set<string>(aria.keys()));
```
rafaelss95, line 58: Could you explain why this function is necessary? Can't we just store it in a variable as before?

arturovt: This is a "lazy initialization", the set should not be initialized until this rule is applied. Otherwise, we will have an initialized set that we do not use.

rafaelss95: That makes sense!

rafaelss95 (approved):
LGTM!

PR #244, opened 2020-12-20
docs: update README with links and status

rafaelss95:
This updates the README with links of ESLint equivalent rules as well as put some status referent to stylistic rules.

Btw, I borrow it from https://github.com/typescript-eslint/typescript-eslint/blob/c8a4dadeab8a64fb4768deda8f65475435dd2cad/packages/eslint-plugin/ROADMAP.md :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,84 +393,90 @@ If you are still having problems after you have done some digging into these, fe
 
 <br>
 
-### Rules List
+# Codelyzer Migration Guide
 
-<!-- begin rule list -->
+This document serves as a guide to help you migrate from Codelyzer.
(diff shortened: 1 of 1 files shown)


rafaelss95:
Btw while I was doing this I noticed that Codelyzer site is using a expired an expired SSL certificate. cc @mgechev

JamesHenry:
I think you must have missed that the rules lists are maintained by tooling, and so this PR cannot proceed.

I'm sorry that you have put in this work manually and that it cannot be used. It might be safer to start a discussion first next time.

If you want to propose an alternative update to the tooling, please feel free

rafaelss95:
Oh, I hadn't realized that a tool was generating it. I will definitely take a look at this, thanks!

PR #245, opened 2020-12-20
refactor: remove use-pipe-decorator rule

### LINKED ISSUE #241: use-pipe-decorator seems to be unnecessary
If you try to declare a `@Pipe` in a `@NgModule` without the decorator, you'll get a compile-time error:

> The class 'XXXPipe' is listed in the declarations of the NgModule 'XXXModule', but is not a directive, a component, or a pipe. Either remove it from the NgModule's declarations, or add an appropriate Angular decorator.

Reference: https://github.com/mgechev/codelyzer/issues/592

rafaelss95:
BREAKING CHANGE

Closes #241.


JamesHenry:
I wonder if we should have a way of marking it as deprecated initially? I know there are some folks are still not on Ivy, and this rule would still be valuable to them right?

rafaelss95:
> I wonder if we should have a way of marking it as deprecated initially? 

I thought of marking it as deprecated, however, as I don't know what policy has been adopted in beta versions here, I opened it as a draft for discussion 😆.

> I know there are some folks are still not on Ivy, and this rule would still be valuable to them right?

Not really, I tested it and even on Angular 8 (without Ivy, because until then it hadn't even been officially released) the results are the same: an error is thrown if you declare a class in `@ NgModule` without `Component`/`Directive`/`Pipe` decorators.

Btw, you can check demos here:

Angular 8 (Ivy doesn't released it): https://stackblitz.com/edit/angular-8-getting-started-zbxkh7
Angular 9 (without Ivy): https://stackblitz.com/edit/angular-9-starter-jc5u5j

JamesHenry:
Ah nice, well yeah then I think you have proved this lint rule doesn't need to exist at all :)

The way I have dealt with breaking changes during this pre-stable phase is to increment the number after the zero, so we can merge this and release 0.9

JamesHenry (changes requested):
Couple of things I'd like to do before merging...

- Rather than removing from the docs, extend the docs generator to include a custom "no longer required" type message
- Figure out with @JoshuaKGoldberg how to update tslint-to-eslint-config appropriately

Josh, what we have here is an unusual case where the rule simply does not need to exist, so the conversion for this: https://github.com/typescript-eslint/tslint-to-eslint-config/blob/main/src/converters/lintConfigs/rules/ruleConverters/codelyzer/use-pipe-decorator.ts

...should really be a noop...

What do you think is the best way to handle that? Is there precedent for an explicitly handled noop in the library? 

------

If we don't do the two things above we'll be setting ourselves up for a ton of issues with people wondering where the rule implementation is and/or getting rule cannot be found errors after auto-converting

JoshuaKGoldberg:
Sure! I'm not able to do it right this moment, but you could file an issue & send a PR over to [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config) with:

```diff
import { RuleConverter } from "../../ruleConverter";

export const convertUsePipeDecorator: RuleConverter = () => {
    return {
        rules: [
-            {
-                ruleName: "@angular-eslint/use-pipe-decorator",
-            },
        ],
-        plugins: ["@angular-eslint/eslint-plugin"],
    };
};
```

...and that would be sufficient. 👍 

(if nobody does in a few days, I will)

rafaelss95:
I'm afraid now that it wasn't merged before v1, it wouldn't be merged anytime soon, right?

JamesHenry:
@rafaelss95 We can release v2 relatively quickly, I don't any reason to artificially delay that when we only have low impact breaking changes like this one on the table right now, but naturally the tasks I mentioned above still need to be done before that can be thought about

JamesHenry:
I update the description to include `BREAKING CHANGE` that is important for clarity as well as lerna's version computation

rafaelss95:
PR opened: https://github.com/typescript-eslint/tslint-to-eslint-config/pull/924.

JoshuaKGoldberg:
☝️ released in [tslint-to-eslint-config@2.1.0](https://www.npmjs.com/package/tslint-to-eslint-config/v/2.1.0) ([release notes](https://github.com/typescript-eslint/tslint-to-eslint-config/releases/tag/v2.1.0)). ✅

PR #251, opened 2020-12-22
refactor(eslint-plugin-template): no-any switch to selectors

rafaelss95:


packages/eslint-plugin-template/src/rules/no-any.ts
```diff
@@ -13,42 +11,33 @@ export default createESLintRule<Options, MessageIds>({
...
-      MethodCall({ name, receiver, sourceSpan }: any) {
-        const isAnyTypeCastFunction = name === ANY_TYPE_CAST_FUNCTION_NAME;
-        const isAngularAnyTypeCastFunction =
-          !receiver.expression && !receiver.name;
-
-        if (!isAnyTypeCastFunction || !isAngularAnyTypeCastFunction) return;
-
-        const start = sourceCode.getLocFromIndex(sourceSpan.start);
-        const end = sourceCode.getLocFromIndex(sourceSpan.end);
+    return {
```
JamesHenry, line 27: Why have you removed the use of the `getTemplateParserServices`? 

The use of parserServices ensures that the rule is being in the right context. As things stand you have changed to rule so that it will attempt to operate on any AST that has a `MethodCall` node type, so I believe now if it were configured by users to be used with the wrong parser it would just silently do nothing.

This is arguably the worst thing a linting rule can do, because users will think they have no violations when they might have many.

`getTemplateParserServices` will throw if you are using the wrong parser so it's a valuable protection against user error.

JamesHenry (changes requested):

packages/eslint-plugin-template/src/rules/no-any.ts
JamesHenry: Just to add more context to this: `defineTemplateBodyVisitor` was added a lonnnggg time ago when trying to figure out how template linting could possible work, and there was some prior art in the Vue ecosystem.

I think you have correctly identified that actually in our Angular context, we only need to be dealing with one AST at a time. It's either a TypeScript source file, or an HTML template file (inline templates are extracted by a processor and so get treated like HTML files).

It is therefore enough for us to just leverage the custom visitorKeys to instruct ESLint on how to traverse the Angular template AST, and we do not require the additional helper method.

I think therefore we just need a new way of getting that same benefit around checks for the right parser being in place...

JamesHenry:
It wouldn't let me push the change from my local to your remote branch unfortunately...

I added an `ensureTemplateParser(context);` utility in #264 and applied the change on top of your commit, so the attribution is still there to you as a co-author. Thanks again!

I'll apply the utility to the remaining PRs as a suggestion where needed

PR #253, opened 2020-12-22
refactor(eslint-plugin-template): no-call-expression switch to selectors

rafaelss95:


packages/eslint-plugin-template/src/rules/no-call-expression.ts
```diff
@@ -1,52 +1,43 @@
...
       recommended: false,
     },
     schema: [],
     messages: {
-      noCallExpression: 'Avoid calling expressions in templates.',
+      noCallExpression: 'Avoid calling expressions in templates',
     },
   },
   defaultOptions: [],
   create(context) {
```
JamesHenry, line 24:
```suggestion
  create(context) {
    ensureTemplateParser(context);
```

packages/eslint-plugin-template/src/rules/no-call-expression.ts
```diff
@@ -1,52 +1,43 @@
-import {
-  createESLintRule,
-  getTemplateParserServices,
-} from '../utils/create-eslint-rule';
+import { MethodCall } from '@angular/compiler';
+import { createESLintRule } from '../utils/create-eslint-rule';
```
JamesHenry, line 2:
```suggestion
import {
  createESLintRule,
  ensureTemplateParser,
} from '../utils/create-eslint-rule';
```

JamesHenry (changes requested):

JamesHenry:
I've attempted to apply the ensureTemplateParser utility myself via suggestions, the only risk is I may not have got the formatting exactly right but 🤞

JamesHenry:
It worked!

PR #254, opened 2020-12-22
refactor(eslint-plugin-template): no-negated-async switch to selectors

rafaelss95:


packages/eslint-plugin-template/src/rules/no-negated-async.ts
```diff
@@ -27,54 +24,48 @@ export default createESLintRule<Options, MessageIds>({
   },
   defaultOptions: [],
   create(context) {
```
JamesHenry, line 26:
```suggestion
  create(context) {
    ensureTemplateParser(context);
```

packages/eslint-plugin-template/src/rules/no-negated-async.ts
```diff
@@ -1,7 +1,4 @@
-import {
-  createESLintRule,
-  getTemplateParserServices,
-} from '../utils/create-eslint-rule';
+import { createESLintRule } from '../utils/create-eslint-rule';
```
JamesHenry, line 1:
```suggestion
import {
  createESLintRule,
  ensureTemplateParser,
} from '../utils/create-eslint-rule';
```

JamesHenry (changes requested):

JamesHenry (approved):

packages/eslint-plugin-template/src/rules/no-negated-async.ts
```diff
@@ -27,54 +27,49 @@ export default createESLintRule<Options, MessageIds>({
...
+    ensureTemplateParser(context);
     const sourceCode = context.getSourceCode();
 
-    return parserServices.defineTemplateBodyVisitor({
-      ['BindingPipe[name=async]'](node: any) {
-        if (node.parent.type === 'PrefixNot') {
-          const additionalOffset =
-            node.parent.parent.type === 'Interpolation' ? -1 : 0;
+    return {
+      'PrefixNot > BindingPipe[name=async]'({
```
rafaelss95, line 34: Just noticed that I missed quotes in `name=async`, but I believe that isn't it a problem, right @JamesHenry?

PR #262, opened 2020-12-22
chore: update verdaccio version and config

JamesHenry:
Trying to combat integration test install flakiness by implementing the suggestions from: https://github.com/verdaccio/verdaccio/discussions/2038

I needed to regenerate the README because there are open rule PRs

### DIFF
--- a/packages/integration-tests/local-registry/config.yml
+++ b/packages/integration-tests/local-registry/config.yml
@@ -51,4 +51,5 @@ packages:
 
 # log settings
 logs:
-  - { type: stdout, format: pretty, level: http }
+  # Change level to http to get more verbose output when working locally if needed
+  - { type: stdout, format: pretty, level: warn }
--- a/packages/integration-tests/local-registry/config.yml
+++ b/packages/integration-tests/local-registry/config.yml
@@ -13,6 +13,10 @@ uplinks:
     timeout: 60s
     cache: false
     maxage: 30m
+    agent_options:
+      keepAlive: true
+      maxSockets: 40
+      maxFreeSockets: 10
   yarn:
     url: https://registry.yarnpkg.com
     max_fails: 30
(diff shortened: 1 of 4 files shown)


packages/integration-tests/local-registry/config.yml
```diff
@@ -8,12 +8,18 @@ auth:
...
+    cache: false
+    maxage: 30m
   yarn:
     url: https://registry.yarnpkg.com
-    max_fails: 20
+    max_fails: 30
     fail_timeout: 10m
+    timeout: 60s
+    cache: false
+    maxage: 30m
```
juanpicado, line 22: You missed the `keep-alive` is the one which actually boots performance.

juanpicado: https://github.com/verdaccio/verdaccio/blob/master/conf/default.yaml#L39

juanpicado (changes requested):
Once you apply `keep-alive` all performance issues should gone.

packages/integration-tests/local-registry/config.yml
juanpicado: Also this
```
logs:
  - { type: stdout, format: pretty, level: http }
  ```
slow down a lot, don't use `logs` on CI, change `http` to `warn`.

JamesHenry: Sorry maybe I misunderstood how the config gets applied then, I had assumed default config meant I didn't have to enable it? Are you saying that by specifying any kind of uplink in my own config I am no longer getting any of the default values?

juanpicado: Yes, currently the keep alive is not built in, you must defined it. In the next minor would be built in then you don't need that configuration anymore (only keep alive)

JamesHenry: Ah ok thanks

JamesHenry: And thanks I've updated the logs config

juanpicado (approved):
Looks like pretty green now :)

JamesHenry:
Nice, thank you @juanpicado! I'll let you know if I observe any more issues

PR #263, opened 2020-12-22
perf(eslint-plugin): initialize the set lazily only when rule is applied

arturovt:


arturovt:
@rafaelss95

JamesHenry:
@arturovt I didn't raise it when this same pattern was part of other PRs but this feels like a bit of a micro-optimization where the upside is questionable.

Fundamentally we aren't measuring anything here, so at best this "feels like" a performance improvement.

Even intuitively though, do we really feel like a Set of a a few dozen strings is causing a performance problem? I'm really not sure

arturovt:
> Set of a a few dozen strings is causing a performance problem

Of course not, it only consumes memory. E.g. we're using t2.micro AWS instances which have 1 GB of RAM for CI flows and currently it's enough, sometimes we have to deal with npm packages that do a lot of unnecessary initializations like sets, arrays, maps, objects, etc.

before
---
after

I will leave a final decision to you if you think it's unnecessary.

JamesHenry:
Thanks for sharing the measurement! We can be relatively certain this will not do any harm so happy to merge

PR #268, opened 2020-12-23
feat(eslint-plugin-template): accessibility-label-for

### LINKED ISSUE #277: FR: Codelyzer (Angular) rule: template-accessibility-label-for
🚀 Feature Request

Package Source
Codelyzr: http://codelyzer.com/rules/template-accessibility-label-for

PS: if there is no intention in adding this rule, is there any alternative for it?

rafaelss95:
As soon as I started this implementation, I noticed that [the current version in Codelyzer](https://github.com/mgechev/codelyzer/blob/master/src/templateAccessibilityLabelForRule.ts) has some problems:

1. It doesn't report labels without association, e.g.:

```html
<label for="nonExistentId">Label</label>
<!-- there's no element with `nonExistentId`, but the rule passes --> 
```

2. One could pass control component selectors through the `controlComponents` option, like this:

```json
"template-accessibility-label-for": [true, {
  "controlComponents": ["p-inputMask"],  
}]
```

```html
<app-label labelFor="username">Label<p-inputMask></p-inputMask></app-label>
```

...however one could have a component with an attribute that acts like an `id`, like [the `inputId` here, by example](https://www.primefaces.org/primeng/showcase/#/inputmask) and thus doesn't need to be wrapped by a label/custom label component. This case isn't covered by the rule, e.g.:

```html
<app-label labelFor="username">Label</app-label>
<p-inputMask formControlName="username" inputId="username" mask="(999) 999-9999? x99999"></p-inputMask>
```

3. The `labelAttributes` and `labelComponents` are stored in separated structures and it could be problematic, because a label attribute can cause conflicts depending on the label component you're using, e.g.:

```json
"template-accessibility-label-for": [true, {
  "labelAttributes": ["assoc", "elementId"],
  "labelComponents": ["app-label", "ngx-label"],  
}]
```

```html
<app-label assoc="id"></app-label> <!-- ok -->
<ngx-label assoc="id"></ngx-label > <!-- not ok, but the rule doesn't report -->
<app-label elementId="id"></app-label> <!-- not ok, but the rule doesn't report -->
```

<hr>

After thinking a little bit on how we could store the structures to meet these cases, I came up with this:

```js
"template-accessibility-label-for": [true, {
  "controlComponents": [
    { "inputs": ["id"], "selector": "input" }, // one of the defaults
    { "inputs": ["inputId"], "selector": "p-inputMask" },
    { "selector": "bs4-input" } // when it doesn't have `inputs`, it must be wrapped by a label/label component.
  ],
  "labelComponents": [
    { "inputs": ["for", "htmlFor"], "selector": "label" }, // default
    { "inputs": ["assoc"], "selector": "app-label" },
    { "inputs": ["elementId", "legacyId"], "selector": "ngx-label" },
    { "selector": "ng2-label" }, // when it doesn't have `inputs`, it must wrap a control component.
  ],
}]
```

That being said, would it be a problem for the migration tool to a migrate a rule that changes options like this?
To maintain the behavior in a migration the changes would be like this (I'm using TSLint as an example):

```diff
"template-accessibility-label-for": [true, {
- "controlComponents": ["p-inputMask", "bs4-input"],
- "labelAttributes": ["assoc", "elementId"],
- "labelComponents": ["app-label", "ngx-label"],
+ "controlComponents": [
+   { "selector": "p-inputMask" }, { "selector": "bs4-input" }
+ ],
+ "labelComponents": [
+   { "inputs": ["assoc", "elementId"], "selector": "app-label" },
+   { "inputs": ["assoc", "elementId"], "selector": "ngx-label" }
+ ],
}]
```

I would appreciate a feedback on this, thanks.

Closes #277.

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,7 +463,7 @@ If you are still having problems after you have done some digging into these, fe
 | [`prefer-on-push-component-change-detection`]   | :white_check_mark: |
 | [`template-accessibility-alt-text`]             | :white_check_mark: |
 | [`template-accessibility-elements-content`]     | :white_check_mark: |
-| [`template-accessibility-label-for`]            |                    |
+| [`template-accessibility-label-for`]            | :white_check_mark: |
 | [`template-accessibility-tabindex-no-positive`] | :white_check_mark: |
 | [`template-accessibility-table-scope`]          | :white_check_mark: |
 | [`template-accessibility-valid-aria`]           | :white_check_mark: |
--- a/packages/eslint-plugin-template/src/configs/all.json
+++ b/packages/eslint-plugin-template/src/configs/all.json
@@ -3,6 +3,7 @@
   "rules": {
     "@angular-eslint/template/accessibility-alt-text": "error",
     "@angular-eslint/template/accessibility-elements-content": "error",
+    "@angular-eslint/template/accessibility-label-for": "error",
     "@angular-eslint/template/accessibility-table-scope": "error",
     "@angular-eslint/template/accessibility-valid-aria": "error",
     "@angular-eslint/template/banana-in-box": "error",
--- a/packages/eslint-plugin-template/src/index.ts
+++ b/packages/eslint-plugin-template/src/index.ts
@@ -9,6 +9,9 @@ import accessibilityAltText, {
 import accessibilityElementsContent, {
   RULE_NAME as accessibilityElementsContentRuleName,
 } from './rules/accessibility-elements-content';
+import accessibilityLabelFor, {
+  RULE_NAME as accessibilityLabelForRuleName,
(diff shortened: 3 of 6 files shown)


rafaelss95:
What's your opinion on this, @JamesHenry?

JamesHenry (changes requested):
I have never personally used this rule or anything similar to it, so will need to take some time to understand it and what you've written.

My gut reaction is that we should prioritise merging in the converted version as it exists today in v1, and then look to fix up issues in follow ups.

If that requires breaking changes we can release the updates as part of v2, and provide an ng update schematic to migrate user configs

JamesHenry (approved):
Thanks a lot!

PR #269, opened 2020-12-25
fix(eslint-plugin-template): no-call-expression not reporting failures for safe method c…

rafaelss95:
…alls

As reported in https://github.com/mgechev/codelyzer/issues/757, `no-call-expression` doesn't report failures for safe method calls, e.g.:

```html
{{ a?.b() }}
```

About https://github.com/mgechev/codelyzer/issues/757#issuecomment-620679346, I don't know if we would be able to perform this check without turn the `requiresTypeChecking` on...

### DIFF
--- a/packages/eslint-plugin-template/tests/rules/no-call-expression.test.ts
+++ b/packages/eslint-plugin-template/tests/rules/no-call-expression.test.ts
@@ -21,17 +21,15 @@ ruleTester.run(RULE_NAME, rule, {
   valid: [
     `
       {{ info }}
-    `,
-    `
       <button type="button" (click)="handleClick()">Click Here</button>
-    `,
-    `
       {{ $any(info) }}
+      <input (change)="obj?.changeHandler()">
     `,
   ],
   invalid: [
     convertAnnotatedSourceToFailureCase({
-      description: 'it should fail with call expression in expression binding',
+      description:
+        'it should fail for call expression in an expression binding',
       annotatedSource: `
         <div>{{ getInfo() }}</div>
                 ~~~~~~~~~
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-template/tests/rules/no-call-expression.test.ts
```diff
@@ -31,45 +31,46 @@ ruleTester.run(RULE_NAME, rule, {
...
       messageId,
     }),
-    convertAnnotatedSourceToFailureCase({
-      description: 'it should fail with call expression in expression binding',
-      annotatedSource: `
-        <a href="http://example.com">{{ getInfo() }}</a>
-                                        ~~~~~~~~~
-      `,
-      messageId,
-    }),
```
rafaelss95, line 48: Removed as duplicated test.

packages/eslint-plugin-template/tests/rules/no-call-expression.test.ts
```diff
@@ -31,45 +31,46 @@ ruleTester.run(RULE_NAME, rule, {
...
       messageId,
     }),
     convertAnnotatedSourceToFailureCase({
-      description: 'it should fail with a property access call expression',
+      description: 'it should fail for safe/unsafe method calls',
       annotatedSource: `
-        <a [href]="helper.getUrl()">info</a>
-                   ~~~~~~~~~~~~~~~
+        {{ obj?.nested1() }} {{ obj!.nested1() }}
+            ~~~~~~~~~~~~~~       ^^^^^^^^^^^^^^
```
rafaelss95, line 62: @JamesHenry, 2 points to notice here:

- While I was doing this, I remembered that we discussed this at https://github.com/angular-eslint/angular-eslint/pull/205#discussion_r532132323, so I decided to debug deeper this time and finally figured out why it happens. The "problem" is that Angular condenses consequent interpolations in a single `BoundText`, and so in order to check the error by passing a `char`, we should inline all the interpolations in a single line.

- Despite having solved this, I noticed another one where the error column is off by one (the same happens in the `template-no-any`)... maybe is it an bug with `getLocFromIndex`? I also noticed that it was reported in https://github.com/mgechev/codelyzer/issues/859.

PR #272, opened 2020-12-30
fix(eslint-plugin-template): accessibility-valid-aria reporting non a…

### LINKED ISSUE #271: options for valid aria attributes rule
It would be nice if I could provide an exceptions list/ignored properties list for `@angular-eslint/template/accessibility-valid-aria` rule.

I enabled this rule and ended up with this error:
image

In this particular case `variant` is an `@input` of a button. Ideally of course this should not be an issue, but I doubt eslint has enough information to infer this.

If someone agrees, I could open a PR :)

rafaelss95:
…ria attributes

Currently the rule `accessibility-valid-aria` reports any word that **contain** the word "aria", like "variant", "arial", "ariaAnother", etc., while it should only report words that **start** with "aria-".

Fixes #271.

### DIFF
--- a/packages/eslint-plugin-template/tests/rules/accessibility-valid-aria.test.ts
+++ b/packages/eslint-plugin-template/tests/rules/accessibility-valid-aria.test.ts
@@ -14,31 +14,31 @@ import rule, {
 const ruleTester = new RuleTester({
   parser: '@angular-eslint/template-parser',
 });
-
 const messageId: MessageIds = 'accessibilityValidAria';
 
 ruleTester.run(RULE_NAME, rule, {
   valid: [
     '<input aria-labelledby="Text">',
-    '<input [attr.aria-labelledby]="text">',
+    '<div ariaselected="0"></div>',
+    '<textarea [attr.aria-readonly]="readonly"></textarea>',
+    '<button [variant]="variant">Text</button>',
   ],
   invalid: [
     convertAnnotatedSourceToFailureCase({
-      messageId,
-      description:
-        'should fail when aria attributes are misspelled or if they does not exist',
+      description: 'should fail if the attribute is an invalid ARIA attribute',
       annotatedSource: `
-        <input aria-labelby="text">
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-template/tests/rules/accessibility-valid-aria.test.ts
```diff
@@ -20,25 +20,26 @@ const messageId: MessageIds = 'accessibilityValidAria';
...
-        <input [aria-labelby]="text">
-               ~~~~~~~~~~~~~~~~~~~~~
+        <div aria-roledescriptio="text">Text</div>
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~
+        <input [aria-labelby]="label">
+               ^^^^^^^^^^^^^^^^^^^^^^               
+        <input [attr.aria-requiredIf]="required">
+               #################################
       `,
+      messages: [
```
rafaelss95, line 38: @JamesHenry is there anyway to pass `data` when having multiple assertions?

rafaelss95:
While I was fixing this I noticed that we could also expand this rule to report incorrect values, ex:

`<div aria-orientation="test"></div>`

This should be reported since the allowed values are `undefined`, `'horizontal'` and `'vertical'` as we could see in https://github.com/A11yance/aria-query/blob/e4b6d28df777c5e6f2a14fa3cdfba74bc3007e19/src/ariaPropsMap.js#L143-L-147.

WDYT, @JamesHenry? If so, I could send another PR for this.

packages/eslint-plugin-template/tests/rules/accessibility-valid-aria.test.ts
JamesHenry: I can't remember I'm afraid :)

JamesHenry:
@rafaelss95 yeah that sounds like a good enhancement which still fits under the scope of the rule name!

Issue #197, opened 2020-11-24
Docs for individual rules


gSarciotto:
Hello everyone,

is there a plan for adding docs to individual rules? Something similar to what [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin/docs/rules) has. 

If yes, how can I help?

JamesHenry:
This would be a very welcome contribution! Feel free to take inspiration and code from typescript-eslint as the starting point

gSarciotto:
@JamesHenry  Fantastic! I will try to do something going this week.

Where should I put the docs? I thought something like {{plugin}}/docs/rules

And how should we proceed, I create an issue for each rule or just the PR?

JamesHenry:
Sounds good on the location, no need for extra issues, you can just start with a draft PR.

Maybe do 1 PR per plugin for now

sam-s4s:
Am just moving from tslint to eslint for angular, and am finding it very difficult to do with no documentation on the rules lol. I guess for now you just have to go and look at the rules' source code to figure out how it works.

Looks like there's "only" 32 rules.

Perhaps if we all work together we can get the docs done?

jfgreffier:
It seems that the strategy is to auto-generate docs from rules, although there were no progress since May https://github.com/angular-eslint/angular-eslint/pull/507#issuecomment-848048082

JamesHenry:
Thanks for your patience on this everyone - I have just merged the first version of the automated docs generation logic. We will continue to improve on it from here, but for now it addresses this issue.

Thanks again!

jfgreffier:
That's great ! Thank you @JamesHenry

PR #276, opened 2021-01-01
docs: Created docs for rule 'component-class-suffix'

gSarciotto:
Docs for rule [component-class-sufix](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/rules/component-class-suffix.ts), in accordance to the effort for documentation #197

### DIFF
--- a/packages/eslint-plugin/docs/rules/component-class-suffix.md
+++ b/packages/eslint-plugin/docs/rules/component-class-suffix.md
@@ -1,0 +1,74 @@
+# Enforces components to have a specified suffix (`component-class-suffix`)
+
+This rule enforces that classes decorated with @Component must have suffix "Component" (or custom) in their name.
+
+## Rule Details
+
+This rule follows the recommendation from the [Angular styleguide](https://angular.io/guide/styleguide#style-02-03).
+
+Examples of **incorrect** code for this rule (with default configuration):
+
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/docs/rules/component-class-suffix.md
```diff
@@ -0,0 +1,80 @@
...
+
+This rule enforces that classes decorated with @Component must have suffix "Component" (or custom) in their name.
+
+## Rule Details
+
+This rule follows the recommendation from the [Angular styleguide](https://angular.io/guide/styleguide#style-02-03).
+
+Examples of **incorrect** code for this rule (with default configuration):
+
+```javascript
```
JamesHenry, line 11: These should all be ` ```ts `, rather than ` ```javascript`

JamesHenry (changes requested):
Thanks a lot for kicking off this effort! See the one comment (which applies multiple times)

packages/eslint-plugin/docs/rules/component-class-suffix.md
gSarciotto: Oh, I completely missed it haha, thanks!

JamesHenry (approved):
Thanks!

PR #278, opened 2021-01-03
fix(eslint-plugin-template): accessibility-valid-aria not reporting i…

rafaelss95:
…nvalid 'aria-' values

Relates to https://github.com/angular-eslint/angular-eslint/pull/272#issuecomment-752797643.

PS: Not sure it should be considered a feat or a fix.


packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -1,13 +1,29 @@
...
 import {
   createESLintRule,
   getTemplateParserServices,
 } from '../utils/create-eslint-rule';
 
+// As in the time it being write the `ARIAPropertyDefinition` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/64ae44025dbd41afb3b865ca37deddef7de1c5fb/types/aria-query/index.d.ts#L302-L306)
+// is incorrect (`value` should be `values`), we created this hackish type to avoid `any`.
+type ARIAPropertyDefinitionHack = Omit<ARIAPropertyDefinition, 'value'> & {
+  readonly values: ARIAPropertyDefinition['value'];
+};
```
rafaelss95, line 21: I hope the comment is clear enough to express the intent.

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -1,13 +1,29 @@
...
+  TmplAstBoundAttribute,
+  TmplAstTextAttribute,
+} from '@angular/compiler';
 
 import {
   createESLintRule,
   getTemplateParserServices,
 } from '../utils/create-eslint-rule';
 
+// As in the time it being write the `ARIAPropertyDefinition` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/64ae44025dbd41afb3b865ca37deddef7de1c5fb/types/aria-query/index.d.ts#L302-L306)
```
JamesHenry, line 17:
```suggestion
// At the time of writing, the `ARIAPropertyDefinition` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/64ae44025dbd41afb3b865ca37deddef7de1c5fb/types/aria-query/index.d.ts#L302-L306)
```

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
```diff
@@ -1,13 +1,29 @@
...
+// is incorrect (`value` should be `values`), we created this hackish type to avoid `any`.
```
JamesHenry, line 18:
```suggestion
// is incorrect (`value` should be `values`), we created this hacky workaround type to avoid `any`.
```

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
JamesHenry: Thanks yes, made sense, just added some minor English language suggestions

JamesHenry (changes requested):
Just the minor suggestions to apply, thanks a lot!

packages/eslint-plugin-template/src/rules/accessibility-valid-aria.ts
rafaelss95: I've opened a PR in the DefinitelyTyped repo: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50528. What do you think about delaying this merge a little bit until the PR is merged? Of course, if it takes too long to merge, we can move on as is.

rafaelss95: I removed this as https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50528 has been merged.

JamesHenry (approved):
Thanks again!

PR #279, opened 2021-01-03
fix(eslint-plugin-template): conditional-complexity not reporting all cases

rafaelss95:
This PR aims to fix the cases:

- report when using any other bound attribute besides `ngIf`, which is the only one currently reported;
- *report when using a `Pipe` as the last statement in a condition;
- report when using `Conditional`;
- report when using `Interpolation`;

This also removes some duplicate unit tests and include `data` to ensure the complexity is computed correctly.

*: This case was reported in #280 (**case 3**) and is partially fixed here, because it was fixed only when it appears as the last statement in a condition. To completely fix it, further investigation will be required and will take some more time and probably better to do in another PR.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/conditional-complexity.ts
+++ b/packages/eslint-plugin-template/src/rules/conditional-complexity.ts
@@ -1,6 +1,10 @@
 import {
+  AST,
   ASTWithSource,
   Binary,
+  BindingPipe,
+  Conditional,
+  Interpolation,
   Lexer,
   Parser,
   TmplAstBoundAttribute,
@@ -8,11 +12,11 @@ import {
 
 import {
   createESLintRule,
+  ensureTemplateParser,
   getTemplateParserServices,
 } from '../utils/create-eslint-rule';
 
 type Options = [{ maxComplexity: number }];
-
 export type MessageIds = 'conditionalСomplexity';
 export const RULE_NAME = 'conditional-complexity';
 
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-template/src/rules/conditional-complexity.ts
```diff
@@ -21,37 +25,41 @@ export default createESLintRule<Options, MessageIds>({
...
     ],
     messages: {
       conditionalСomplexity:
-        'The condition complexity "{{totalComplexity}}" exceeds the defined limit "{{maxComplexity}}"',
+        'The conditional complexity "{{totalComplexity}}" exceeds the defined limit "{{maxComplexity}}"',
     },
   },
   defaultOptions: [{ maxComplexity: 5 }],
   create(context, [{ maxComplexity }]) {
+    ensureTemplateParser(context);
```
rafaelss95, line 47: I had to change the signature of `ensureTemplateParser` in order to compile:

Captura de Tela 2021-01-04 às 02 20 24

packages/eslint-plugin-template/src/utils/create-eslint-rule.ts
```diff
@@ -25,12 +25,11 @@ export function getTemplateParserServices(context: any): ParserServices {
  * If @angular-eslint/template-parser is not the configured parser when the function is invoked it will throw
  */
 export function ensureTemplateParser(
-  context: TSESLint.RuleContext<string, []>,
+  context: TSESLint.RuleContext<string, readonly unknown[]>,
```
rafaelss95, line 28: I had to change the signature of `ensureTemplateParser` in order to compile:

Captura de Tela 2021-01-04 às 02 20 24

packages/eslint-plugin-template/src/rules/conditional-complexity.ts
rafaelss95, line 47: Is this still necessary since I'm using `parserServices` (for the `BoundAttribute` case)?

JamesHenry: No, you're correct, it is only necessary if not using `parserServices`, just helps protect users from leveraging template rules with the wrong parser in their configs

packages/eslint-plugin-template/src/utils/create-eslint-rule.ts
JamesHenry: Cool

JamesHenry (commented):
It's not going to do any harm to leave it in, but given you've called it out I'll wait until you remove the `ensureTemplateParser` call

JamesHenry:
Thanks a lot!

Issue #293, opened 2021-01-14
Rule proposal: no duplicate attributes


mattlewis92:
When investigating switching from ViewEngine to ivy we stumbled on a problem where we had components with the same attribute being passed in twice and the behaviour is different between view engine and ivy. e.g. consider the component:

```
@Component({
  selector: 'foo',
  template: '{{ bar }}',
})
export class FooComponent {
  @Input() bar: string
}
```

Is used like this:
```
<foo [bar]="'bam'" [bar]="'baz'"></foo>
```

ViewEngine: uses the first input and prints `bam`
Ivy: uses the second input and prints `baz`

So I think this would make sense as a lint rule to detect these duplicate attributes so that they can be removed (usually they are added by accident by merge conflicts). I did a quick check with [AST explorer](https://astexplorer.net/#/gist/90f46d6f946b91408ccc02f5f99efd35/f99663d6a169f0cef0d51f885c46327f2a1763c0) and it looks like they are represented in the 
AST so it should be possible to write a lint rule to detect them.

I would be happy to work on this myself, I think it should be straightforward to implement. If anyone has any suggestions on a descriptive name for the rule that would be very helpful 🚲

rafaelss95:
Hey @mattlewis92, thanks for opening this proposal here ([it looks like the same I proposed in Codelyzer](https://github.com/mgechev/codelyzer/issues/635)) and I would love to see this rule here also :). If it gets approved, it would be nice to handle not only `BoundAttribute`, but also `TextAttribute` and `BoundEvent`, so all these cases should have error/warn:

```html
<input (change)="changeHandler($event)" (change)="changeHandler($event)"><!-- BoundEvent + BoundEvent -->
<input [(ngModel)]="model" [(ngModel)]="model"><!-- BoundEvent + BoundEvent -->
<foo [bar]="'bam'" bar="baz"></foo><!-- BoundAttribute + TextAttribute -->
```

...and if that's the case, in order to describe the intent better, the name would need to be more descriptive.

<hr>

Hmm, another case came to my mind:

```html
<input [(ngModel)]="model" (ngModelChange)="model = $event">
```

In this case, the assignment would happen **two** times, since `[(ngModel)]` is just a short-cut to `[ngModel]="model" (ngModelChange)="model = $event"`, as detailed in https://github.com/angular/angular/issues/11234#issuecomment-244067124, **but** if it's considered much restrictive for this rule, could we have an option for this then?

[...]

PR #302, opened 2021-01-17
feat(eslint-plugin-template): add no duplicate attributes rule

mattlewis92:
This PR implements the rule suggested in https://github.com/angular-eslint/angular-eslint/issues/293. This was my first time writing an eslint rule, so apologies if I made any silly mistakes 😄 It was a _really_ nice codebase to contribute to with lots of examples so hopefully there is nothing too bad in there. I tried to highlight the things I wasn't sure about, please let me know if you'd like me to make any changes! 😄

### DIFF
--- a/packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
+++ b/packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
@@ -1,0 +1,158 @@
+import {
+  BindingType,
+  TmplAstBoundAttribute,
+  TmplAstBoundEvent,
+  TmplAstElement,
+  TmplAstTextAttribute,
+  ParsedEventType,
+} from '@angular/compiler';
+import {
+  createESLintRule,
+  getTemplateParserServices,
+} from '../utils/create-eslint-rule';
+
+type Options = [
+  {
+    allowTwoWayDataBinding?: boolean;
+  },
+];
+export type MessageIds = 'noDuplicateAttributes';
+export const RULE_NAME = 'no-duplicate-attributes';
+
+const defaultOptions = {
+  allowTwoWayDataBinding: true,
+};
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Ensures that there are no duplicate input properties or output event listeners',
+      category: 'Possible Errors',
+      recommended: false,
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allowTwoWayDataBinding: {
+            type: 'boolean',
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      noDuplicateAttributes: 'Duplicate attribute "{{attributeName}}"',
+    },
+  },
+  defaultOptions: [defaultOptions],
+  create(context, [{ allowTwoWayDataBinding }]) {
+    const parserServices = getTemplateParserServices(context);
+
+    return {
+      Element({ inputs, outputs, attributes }: TmplAstElement) {
+        const duplicateInputsAndAttributes = findDuplicates([
+          ...inputs,
+          ...attributes,
+        ]);
+
+        const filteredOutputs = allowTwoWayDataBinding
+          ? outputs.filter((output) => {
+              return !inputs.some(
+                (input) =>
+                  input.sourceSpan.start === output.sourceSpan.start &&
+                  input.sourceSpan.end === output.sourceSpan.end,
+              );
+            })
+          : outputs;
+
+        const duplicateOutputs = findDuplicates(filteredOutputs);
+
+        const allDuplicates = [
+          ...duplicateInputsAndAttributes,
+          ...duplicateOutputs,
+        ];
+
+        allDuplicates.forEach((duplicate) => {
+          const loc = parserServices.convertNodeSourceSpanToLoc(
+            duplicate.sourceSpan,
+          );
+
+          context.report({
+            messageId: 'noDuplicateAttributes',
+            loc,
+            data: {
+              attributeName: getAttributeName(duplicate),
+            },
+          });
+        });
+      },
+    };
+  },
+});
+
+interface BoundAttribute extends Omit<TmplAstBoundAttribute, 'type'> {
+  type: 'BoundAttribute';
+  __originalType: BindingType;
+}
+
+interface BoundEvent extends Omit<TmplAstBoundEvent, 'type'> {
+  type: 'BoundEvent';
+  __originalType: ParsedEventType;
+}
+
+function getAttributeName(
+  attribute:
+    | BoundAttribute
+    | TmplAstTextAttribute
+    | BoundEvent
+    | { name: string },
+): string {
+  if ('type' in attribute) {
+    if (attribute.type === 'BoundAttribute') {
+      switch (attribute.__originalType) {
+        case BindingType.Class:
+          return `class.${attribute.name}`;
+        case BindingType.Style:
+          return `style.${attribute.name}${
+            attribute.unit ? '.' + attribute.unit : ''
+          }`;
+        case BindingType.Animation:
+          return `@${attribute.name}`;
+      }
+    } else if (attribute.type === 'BoundEvent') {
+      if (attribute.__originalType === ParsedEventType.Animation) {
+        return `@${attribute.name}${
+          attribute.phase ? '.' + attribute.phase : ''
+        }`;
+      }
+
+      if (attribute.target) {
+        return `${attribute.target}:${attribute.name}`;
+      }
+    }
+  }
+
+  return attribute.name;
+}
+
+function findDuplicates(
+  elements: Array<TmplAstBoundEvent>,
+): Array<TmplAstBoundEvent>;
+function findDuplicates(
+  elements: Array<TmplAstBoundAttribute | TmplAstTextAttribute>,
+): Array<TmplAstBoundAttribute | TmplAstTextAttribute>;
+function findDuplicates(
+  elements: Array<{ name: string }>,
+): Array<{ name: string }> {
+  return elements.filter((element) => {
+    return elements.some(
+      (otherElement) =>
+        otherElement !== element &&
+        getAttributeName(otherElement) === getAttributeName(element),
+    );
+  });
+}
--- a/packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
+++ b/packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
@@ -1,0 +1,163 @@
+import {
+  convertAnnotatedSourceToFailureCase,
+  RuleTester,
+} from '@angular-eslint/utils';
+import rule, {
+  MessageIds,
+  RULE_NAME,
+} from '../../src/rules/no-duplicate-attributes';
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parser: '@angular-eslint/template-parser',
+});
+
+const messageId: MessageIds = 'noDuplicateAttributes';
+
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
+    '<input name="foo">',
+    '<input [name]="foo">',
+    '<input (change)="bar()">',
+    '<input [(ngModel)]="foo">',
+    '<input [(ngModel)]="model" (ngModelChange)="modelChanged()">',
+    '<div (@fade.start)="animationStarted($event)" (@fade.done)="animationDone($event)"></div>',
+    '<div (window:keydown)="windowKeydown($event)" (document:keydown)="documentKeydown($event)" (document:keyup)="documentKeyup($event)" (keyup)="keyup($event)" (keydown)="keydown($event)"></div>',
+    '<div [style.width.px]="col.width" [width]="col.width"></div>',
+    '<button [class.disabled]="!enabled" [disabled]="!enabled"></button>',
+    '<button [@.disabled]="!enabled" [.disabled]="!enabled"></button>',
+    '<div [style.width]="col.width + \'px\'" [width]="col.width"></div>',
+  ],
+  invalid: [
+    convertAnnotatedSourceToFailureCase({
+      description: 'should fail with 2 inputs with the same name',
+      annotatedSource: `
+        <input [name]="foo" [name]="bar">
+               ~~~~~~~~~~~~ ^^^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description:
+        'should fail with an input and a text attribute with the same name',
+      annotatedSource: `
+        <input [name]="foo" name="bar">
+               ~~~~~~~~~~~~ ^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description: 'should fail with 2 text attributes with the same name',
+      annotatedSource: `
+        <input name="foo" name="bar">
+               ~~~~~~~~~~ ^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description: 'should fail with 2 outputs with the same name',
+      annotatedSource: `
+        <input (change)="foo($event)" (change)="bar($event)">
+               ~~~~~~~~~~~~~~~~~~~~~~ ^^^^^^^^^^^^^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description: 'should fail with 2 banana in a box with the same name',
+      annotatedSource: `
+        <input [(ngModel)]="model" [(ngModel)]="otherModel">
+               ~~~~~~~~~~~~~~~~~~~ ^^^^^^^^^^^^^^^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description:
+        'should fail with duplicate attributes but allow non duplicates',
+      annotatedSource: `
+        <input [name]="foo" [other]="bam" [name]="bar">
+               ~~~~~~~~~~~~               ^^^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description: 'should fail with 3 duplications',
+      annotatedSource: `
+        <input [name]="foo" [name]="bar" [name]="bam">
+               ~~~~~~~~~~~~ ^^^^^^^^^^^^ ############
+      `,
+      messages: [
+        { char: '~', messageId },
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
...
+} from '../utils/create-eslint-rule';
+import {
+  BoundAttribute,
+  BoundEvent,
+  TextAttribute,
+} from '@angular/compiler/src/render3/r3_ast';
+
+type Options = [];
+export type MessageIds = 'noDuplicateAttributes';
+export const RULE_NAME = 'no-duplicate-attributes';
```
mattlewis92, line 14: The name could probably use some bike shedding here, I couldn't really think of much else though. An input and output on a component are technically html attributes so I don't think it's an incorrect name to use but maybe someone else can think of a better name?

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
...
+    docs: {
+      description:
+        'Ensures that no components have duplicate input properties or output event listeners',
+      category: 'Best Practices',
+      recommended: false,
+    },
+    schema: [],
+    messages: {
+      noDuplicateAttributes:
+        'The attribute "{{attributeName}}" was declared multiple times',
```
mattlewis92, line 29: Should we have separate error messages for when there are duplicate inputs and duplicate outputs (and maybe text attributes as well?). I just kept it simple for the time being but I think it could make sense to separate them out.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
...
+    const parserServices = getTemplateParserServices(context);
+
+    return {
+      Element({ inputs, outputs, attributes }: TmplAstElement) {
+        const duplicateInputsAndAttributes = findDuplicates([
+          ...inputs,
+          ...attributes,
+        ]);
+
+        const nonTwoWayBindingOutputs = outputs.filter((output) => {
```
mattlewis92, line 43: I did not implement the option for treating `<input [(ngModel)]="model" (ngModelChange)="model = $event">` as an error as suggested in https://github.com/angular-eslint/angular-eslint/issues/293#issuecomment-760316991 but it would probably be quite easy to add. As long as we decide not to make it the default it could be added in a followup PR and not be breaking

packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    '<input name="foo">',
+    '<input [name]="foo">',
+    '<input (change)="bar()">',
+    '<input [(ngModel)]="foo">',
+    '<input [(ngModel)]="model" (ngModelChange)="modelChanged()">',
+  ],
+  invalid: [
+    {
+      code: '<input [name]="foo" [name]="bar">',
+      errors: [
```
mattlewis92, line 28: Sadly the `convertAnnotatedSourceToFailureCase` did not work here as it only supports one failing error, it might be nice in a future PR to add support for outputting multiple errors from that util.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
...
+          ...duplicateInputsAndAttributes,
+          ...duplicateOutputs,
+        ];
+
+        allDuplicates.forEach((duplicate) => {
+          const loc = parserServices.convertNodeSourceSpanToLoc(
+            duplicate.sourceSpan,
+          );
+
+          context.report({
```
mattlewis92, line 63: This will output 2 errors for each duplicate attribute e.g. `<input [name]="foo" [name]="bar">` would output 2 errors, one for `[name]="foo"` and one for `[name]="bar"`. I think this would be the expected behaviour but it would be good to get confirmation that this is correct.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
...
+function findDuplicates(
+  elements: Array<{ name: string }>,
+): Array<{ name: string }> {
+  return elements.filter((element) => {
+    return elements.some(
+      (otherElement) =>
+        otherElement !== element && otherElement.name === element.name,
+    );
+  });
+}
```
rafaelss95, line 89: If we agree that we should only report the duplications (excluding the first occurrence), I have this suggestion that could be faster than two implicit loops (but that's just a suggestion):

```suggestion
type Attribute = TmplAstBoundAttribute | TmplAstBoundEvent | TmplAstTextAttribute;

function findDuplicates(elements: readonly Attribute[]): readonly Attribute[] {
  return elements.reduce<readonly Attribute[]>(
    (previousValue, currentValue, index) =>
      previousValue.concat(
        currentValue.name === elements[index + 1]?.name
          ? elements[index + 1]
          : [],
      ),
    [],
  );
}
```

rafaelss95 (commented):
Thanks for the work, @mattlewis92 :) I left my answers to the questions and a feedback with suggestions to further improve the implementation.

mattlewis92:
> Thanks for the work, @mattlewis92 :) I left my answers to the questions and a feedback with suggestions to further improve the implementation.

Awesome, thank you so much! I will get to work on those comments in the next few days.

Some more test cases I found when testing this on our app which need handling


```
(@flyInOut.done)="animationDone($event)"
(@flyInOut.start)="animationStarted($event)"
```

```
class="batch-merge-tasks__dropdown-merge-button"
[class.disabled]="!selectedMergeTask"
[disabled]="!selectedMergeTask"
```

```
[style.width.px]="col.width"
[width]="col.width"
```

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
JamesHenry: If this only deals with Angular's Inputs and Outputs, rather than all HTML attributes, would `no-duplicate-bindings` be more appropriate?

mattlewis92: It does also deal with all html attributes e.g. if you did `<input type="text" type="number">` it would get caught by this rule. Would `no-duplicate-bindings-or-attributes` be too verbose as a rule name?

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
+import { TmplAstElement } from '@angular/compiler';
+import {
+  createESLintRule,
+  getTemplateParserServices,
+} from '../utils/create-eslint-rule';
+import {
+  BoundAttribute,
+  BoundEvent,
+  TextAttribute,
+} from '@angular/compiler/src/render3/r3_ast';
```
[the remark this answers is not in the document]
mattlewis92: oops, I'm not sure why webstorm did that, will fix!

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
mattlewis92: I was just running low on time, I will add it today 😄

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        const nonTwoWayBindingOutputs = outputs.filter((output) => {
+          return !inputs.find(
+            (input) =>
+              input.sourceSpan.start === output.sourceSpan.start &&
+              input.sourceSpan.end === output.sourceSpan.end,
+          );
+        });
+
+        const duplicateOutputs = findDuplicates(nonTwoWayBindingOutputs);
```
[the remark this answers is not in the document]
mattlewis92: By having 2 calls, my thinking was that it would be faster, as you can assume there won't be any inputs or outputs with the same name, if you merge the 2 arrays together you will have to iterate more elements when checking for duplicates. i.e. when checking for a duplicate input I would have to iterate over all of the outputs to check there isn't a duplicate which I know won't be the case.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
mattlewis92: By reporting the second attribute as an error though I think it might make developers introduce bugs when fixing the issue. e.g. if this was reported to me as an error:
```
<input [name]="foo" [name]="bar">
                    ~~~~~~~~~~~~
```

My instinct would be just to follow the error and change the code to:
```
<input [name]="foo">
```

However this would now introduce a bug as the value of `[name]` would have changed from `bar` to `foo`. Really I should have deleted the first attribute.

and just to make the situation more complex, if you're not using ivy then it's the second attribute that should have been removed 😅

packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
mattlewis92: Ah I missed that, thank you!

packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+    {
+      code: '<input [(ngModel)]="model" [(ngModel)]="model">',
+      errors: [
+        { column: 8, endColumn: 27, line: 1, messageId },
+        { column: 28, endColumn: 47, line: 1, messageId },
+      ],
+    },
+  ],
+});
```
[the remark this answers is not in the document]
mattlewis92: Added!

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
mattlewis92: @rafaelss95 can you think of a good name for the option?

packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  valid: [
+    '<input name="foo">',
+    '<input [name]="foo">',
+    '<input (change)="bar()">',
+    '<input [(ngModel)]="foo">',
+    '<input [(ngModel)]="model" (ngModelChange)="modelChanged()">',
+    '<div (@fade.start)="animationStarted($event)" (@fade.done)="animationDone($event)"></div>',
+    '<div (window:keydown)="windowKeydown($event)" (document:keydown)="documentKeydown($event)" (document:keyup)="documentKeyup($event)" (keyup)="keyup($event)" (keydown)="keydown($event)"></div>',
+    '<div [style.width.px]="col.width" [width]="col.width"></div>',
+    '<button [class.disabled]="!enabled" [disabled]="!enabled"></button>',
```
mattlewis92, line 30: @rafaelss95 it doesn't look like `[style.foo]` and `[class.foo]` are represented in [the AST](https://astexplorer.net/#/gist/790639325c3eb4acf9860199d260fa02/9254a75724151f64163fd49104dfea4f2606ed1f) as the `style.` and `class.` are dropped. Any ideas on how we can work around this limitation?

mattlewis92 (commented):
I've addressed most of the review comments + fixed some bugs that I found when testing this rule on our app. The outstanding things are:

1/ decide on the best name for the rule https://github.com/angular-eslint/angular-eslint/pull/302/files#r559227915
2/ add option for treating `<input [(ngModel)]="model" (ngModelChange)="model = $event">` as an error (I can't think of a good name for the option) https://github.com/angular-eslint/angular-eslint/pull/302/files#r559228372
3/ should we report duplicate errors for each attribute? https://github.com/angular-eslint/angular-eslint/pull/302/files#r559228851
4/ `[style.foo]` and `[class.foo]` don't appear to be properly represented in the AST so we need a way to workaround that

packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
JamesHenry: If you take a look at them on astexplorer.net, e.g. here: https://astexplorer.net/#/gist/077162bbaa73aaf055fefbf24dc54cf3/b3c105a1403d3bb5280ee48a7f3229fa8c507e59

You'll notice that the `BoundAttribute`s have a `type` property on them, and they have different values in each case.

Therefore we will need to dig into what these `type`s map to within `@angular/compiler` and use them to differentiate our logic within the rule

JamesHenry: (We apply that original `type` from the `@angular/compiler` onto a property called `__originalType` on our final AST so that it is preserved. I don't think we've actually needed it before now!)

mattlewis92: Ahhh I completely missed that, nice spot! Clicking the definition gives me:
```
export declare const enum BindingType {
    Property = 0,
    Attribute = 1,
    Class = 2,
    Style = 3,
    Animation = 4
}
```

So I can definitely work with that! 😄

rafaelss95: TIL, really nice to know that. Thanks for sharing this info, @JamesHenry :)

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
rafaelss95: Would `allowTwoWayDataBinding` (that is `true` by _default_) be too weird? 🤔

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
rafaelss95: My main concern is the visual noise and possibly unnecessary reports 😞 , but now I understood why you went with this option.

rafaelss95:
Hey, while I was testing some scenarios, I discovered that the compiler doesn't throw error for duplicate variables.

Case 1 (duplicate template reference variables on the same element):
```ts
<input #inputRef type="number" inputmode="numeric" #inputRef>
```

Case 2 (duplicate template reference variables on the same template):
```ts
<div #divRef></div>
<div #divRef></div>
```

Case 3 (duplicate scoped variables on the same template):
```ts
<div *ngFor="let item of [1, 2]; index as index; odd as odd; index as index"></div>
```

Another case is a possible conflict with a component property:

```ts
@Component({
  template: `
    <input #test>
    {{ test }}
  `
})
class TestComponent {
  test = 2;
}
```

 I've reported it in https://github.com/angular/angular/issues/40536, however the [available fix](https://github.com/angular/angular/pull/40538) seems to fix the first case only. That being said, I'm not sure it's still in the scope of this rule, since we're dealing with duplicates or maybe it's a new rule proposal. Anyway, I thought it was important to leave it here.

mattlewis92:
> Hey, while I was testing some scenarios, I discovered that the compiler doesn't throw error for duplicate variables.
> 
> Case 1 (duplicate template reference variables on the same element):
> 
> ```
> <input #inputRef type="number" inputmode="numeric" #inputRef>
> ```
> 
> Case 2 (duplicate template reference variables on the same template):
> 
> ```
> <div #divRef></div>
> <div #divRef></div>
> ```
> 
> Case 3 (duplicate scoped variables on the same template):
> 
> ```
> <div *ngFor="let item of [1, 2]; index as index; odd as odd; index as index"></div>
> ```
> 
> Another case is a possible conflict with a component property:
> 
> ```ts
> @Component({
>   template: `
>     <input #test>
>     {{ test }}
>   `
> })
> class TestComponent {
>   test = 2;
> }
> ```
> 
> I've reported it in [angular/angular#40536](https://github.com/angular/angular/issues/40536), however the [available fix](https://github.com/angular/angular/pull/40538) seems to fix the first case only. That being said, I'm not sure it's still in the scope of this rule, since we're dealing with duplicates or maybe it's a new rule proposal. Anyway, I thought it was important to leave it here.

Oh that is a great point, cases 2 + 4 I've seen a lot of problems in the wild caused by those issues. I think it would make sense as a separate rule though, or even for the angular compiler to just handle all of those cases.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
mattlewis92: Works for me, added to the PR!

mattlewis92:
I've fixed the failing test cases and added the option to disallow two way binding + an output with the same name. We just need to decide on the best name for the rule https://github.com/angular-eslint/angular-eslint/pull/302/files#r559227915 and whether to report errors for each attribute: https://github.com/angular-eslint/angular-eslint/pull/302/files#r559228851 and then this should be ready to go 😁

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -0,0 +1,158 @@
...
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      noDuplicateAttributes: 'Duplicate attribute "{{attributeName}}"',
+    },
+  },
+  defaultOptions: [defaultOptions],
```
rafaelss95, line 51: nit: we generally prefer to inline these options:

```suggestion
  defaultOptions: [{ allowTwoWayDataBinding: true }],
```

packages/eslint-plugin-template/tests/rules/no-duplicate-attributes.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        <input (window:resize)="windowResized($event)" (resize)="resize()" (window:resize)="windowResized2($event)">
+               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+      `,
+      messages: [
+        { char: '~', messageId },
+        { char: '^', messageId },
+      ],
+    }),
+    convertAnnotatedSourceToFailureCase({
+      description: 'not allow two way binding and output with the change event',
```
rafaelss95, line 151: nit: what about changing the description to something like this to maintain the standard that was adopted in other cases in this file:

```suggestion
      description: 'should fail with two way binding and its corresponding output',
```

rafaelss95:
I left some nits, but everything in general looks great to me, especially the tests that I really liked the granularity :stuck_out_tongue:. Really good work, @mattlewis92, thanks!

> I've fixed the failing test cases and added the option to disallow two way binding + an output with the same name. We just need to decide on the best name for the rule https://github.com/angular-eslint/angular-eslint/pull/302/files#r559227915 and whether to report errors for each attribute: https://github.com/angular-eslint/angular-eslint/pull/302/files#r559228851 and then this should be ready to go 😁

Let's wait James' opinion on these.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
JamesHenry: Ah ok cool. Yeah it's a tricky one - correctness vs terseness... The Angular syntax binds to properties, not attributes, so it's technically missing out that distinction, but how much does that matter in practice? Probably not a lot.

Let's proceed as is for now with the more terse name, I don't think there is one term that technically works for attributes, bindings and properties.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
JamesHenry: Yeah tricky one - let's stick with flagging both for now and let the user make the decision.

JamesHenry (approved):

JamesHenry:
Thank you @mattlewis92!

JamesHenry:
@rafaelss95 I think your nits were fair enough, feel free to adjust them in future if you want to

PR #313, opened 2021-01-23
fix(eslint-plugin): component-max-inline-declarations animations not being checked properly 

### LINKED ISSUE #304: @angular-eslint/component-max-inline-declarations" rule not works properly with animations
**Description**
@angular-eslint/component-max-inline-declarations" rule not works properly with animations section. It works fine when animations are declared as string (like in [tests](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts))
or
 
```
@Component({
animations: [
        `trigger('openClose', [
 
          state('open', style({
            height: '200px',
            opacity: 1,
            backgroundColor: 'yellow'
          })),
          state('closed', style({
            height: '100px',
            opacity: 0.5,
            backgroundColor: 'green'
          })),
          transition('open => closed', [
            animate('1s')
          ]),
          transition('closed => open', [
            animate('0.5s')
          ])
        ])`
      ]
})
```

But it doesn't lint inline animations when animation declared explicitly (like in [angular docs](https://angular.io/guide/transition-and-triggers#predefined-states-and-wildcard-matching)) , without string

```
@Component({
animations: [
        trigger('openClose', [
 
          state('open', style({
            height: '200px',
            opacity: 1,
            backgroundColor: 'yellow'
          })),
          state('closed', style({
            height: '100px',
            opacity: 0.5,
            backgroundColor: 'green'
          })),
          transition('open => closed', [
            animate('1s')
          ]),
          transition('closed => open', [
            animate('0.5s')
          ])
        ])
      ]
})
```

**Expectations**
Lint explicitly declared  animations and add such case for test suite.

rafaelss95:
While working on a fix for #304, I noticed that we could clean up this rule/tests, and so I did. Here's the context of each commit:

**First commit**: Currently we perform some checks to verify that the received limit is > `-1`. Now, we rely on the schema that can handle it automatically.

**Second commit**: Currently we use the function `getDecoratorPropertyValue` to get the desired property. Now, we use selectors instead.

**Third commit**: The tests are a bit verbose with a lot of annotations and also we don't check if the lines are calculated properly. With this commit, we have cleaner tests with properties checking through `data`.

**Fourth commit**: Fixes #304, before this commit we were calculating animation lines as `string`, however it isn't. Now, it may be checked properly.

### DIFF
--- a/packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
+++ b/packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
@@ -3,7 +3,7 @@ import {
   RuleTester,
 } from '@angular-eslint/utils';
 import rule, {
-  messageId,
+  MessageIds,
   RULE_NAME,
 } from '../../src/rules/component-max-inline-declarations';
 
@@ -14,10 +14,11 @@ import rule, {
 const ruleTester = new RuleTester({
   parser: '@typescript-eslint/parser',
 });
+const messageId: MessageIds = 'componentMaxInlineDeclarations';
 
 ruleTester.run(RULE_NAME, rule, {
   valid: [
-    // should succeed if the number of lines does not exceeds the default lines limit (template)
+    // should succeed if the number of the template lines does not exceeds the default lines limit
     `
     @Component({
       template: '<div>just one line template</div>'
@@ -25,20 +26,16 @@ ruleTester.run(RULE_NAME, rule, {
     class Test {}
     `,
     {
-      // should succeed if a negative limit is used and the number of lines does not exceeds the default lines limit (template)
+      // should succeed if a negative limit is used and the number of the template lines does not exceeds the default lines limit
       code: `
       @Component({
         template: '<div>first line</div>'
       })
       class Test {}
       `,
-      options: [
-        {
-          template: -5,
-        },
-      ],
+      options: [{ template: -5 }],
     },
-    // should succeed if the number of lines does not exceeds the default lines limit (styles)
+    // should succeed if the number of the styles lines does not exceeds the default lines limit
     `
     @Component({
       styles: ['div { display: none; }']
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
```diff
@@ -14,103 +14,68 @@ import rule, {
...
-      // should succeed if a negative limit is used and the number of lines does not exceeds the default lines limit (styles)
-      code: `
-      @Component({
-        styles: ['div { display: none; }']
-      })
-      class Test {}
-      `,
-      options: [
-        {
-          styles: -5,
```
JamesHenry, line 58: There no longer seems to be any coverage for the case where the user provides an invalid negative number

packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
```diff
@@ -14,103 +14,68 @@ import rule, {
...
+      templateUrl: './foobar.html',
+    })
+    class Test {}
+    `,
   ],
   invalid: [
     convertAnnotatedSourceToFailureCase({
       description:
-        'should fail if the number of lines exceeds the default lines limit (template)',
+        'should fail if the number of the template lines exceeds the default lines limit',
```
JamesHenry, line 63: Why this change to not highlight the whole template/styles etc? Seems like it would be way easier to miss in an editor this way?

If the concern is visual noise, then surely:

```ts
  template: `...`
  ~~~~~~~~
```

...would be a better way to go?

JamesHenry (changes requested):
Thanks for the fixes! Got some notes inline

packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
rafaelss95: Yup, as I've said in the PR description, I moved the manual check we had to schema and thus we don't need to have tests for this anymore, as the configuration is now invalid. That said, I noticed that despite not being a correct configuration or thinking that someone has this configuration with a negative number, it can be considered a breaking change, right? In any case, I'm okay in dropping the first commit and open a follow-up PR to be possibly merged in v2 if that's the case. Let me know your thoughts so I can see what I can make here, thanks.

packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
rafaelss95: The reported node remains the _property_ value as before. What I did was just a clean up to reduce the visual noise in tests only:

```diff
-        styles: [
-                ~
-          \`
-~~~~~~~~~~~
-            div {
-~~~~~~~~~~~~~~~~~
-              display: block;
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-              height: 40px;
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
-            }
-~~~~~~~~~~~~~
-          \`
-~~~~~~~~~~~
-        ]
-~~~~~~~~~

+        styles: [
+               ~
+          \`
+            div {
+              display: block;
+              height: 40px;
+            }
+          \`
+        ]
+        ~
```

Correct me I'm wrong but In tests we don't need to put the character on every line/character to ensure that, we can simply put the character in the begin and other in the end.

packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
JamesHenry: Yeah I'll mark it as a breaking change, but we can do v2 soon so not a big deal.

I'm still not 100% sure on why you wouldn't want to have test coverage for invalid configuration? Surely that's a good thing to capture in tests so that it doesn't inadvertently get changed over time?

packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
JamesHenry: Ha well if that's the case then that's good, I must have forgotten sorry 😅

packages/eslint-plugin/tests/rules/component-max-inline-declarations.test.ts
rafaelss95: Unless I'm missing something, I don't think we can test it, because as soon as we put it there, we receive the following error:

Captura de Tela 2021-02-02 às 09 22 21

That being said, can we move this part to a follow-up PR and then merge this? I want avoid blocking #304 fix with this.

JamesHenry: So you can't move it to the invalid cases (looks like you still have it in valid) and capture it there? That might be the case I just assumed invalid would also mean invalid configuration for the RuleTester...

Are you saying you will remove the breaking change? That sounds fine if so

rafaelss95: > So you can't move it to the invalid cases (looks like you still have it in valid) and capture it there? That might be the case I just assumed invalid would also mean invalid configuration for the RuleTester...

Let's move this discussion to a follow-up PR. I'll open it as soon as it gets merged.

> Are you saying you will remove the breaking change? That sounds fine if so

Yes, I already did it. Take a look please :)

rafaelss95: @JamesHenry Sorry to ping you, but have you ever have a chance to take another look here? 😄

JamesHenry (approved):

PR #243, opened 2020-12-20
fix(eslint-plugin): handle DoBootstrap correctly in lifecycle rules

rafaelss95:
This:

- Handles `DoBootstrap`/`ngDoBootstrap` for all 4 lifecycle rules correctly now;
- `no-empty-lifecycle-method` handles all decorators. Before it was caring only about `Component` and `Directive`;
- `no-empty-lifecycle -method` now ignores non-angular classes;
- `use-lifecycle-interface` now ignores non-angular classes.

PR #314, opened 2021-01-23
fix(eslint-plugin): no-lifecycle-call invalid super calls not being reported

### LINKED ISSUE #246: [no-lifecycle-call]: should only allow super calls in the corresponding child method
Currently the rule `no-lifecycle-call` allows any super calls, even if I try to call `super.ngOnInit()` inside `ngOnDestroy` method, for example. We should only allow it in the corresponding child method.

rafaelss95:
**First commit**: Adds a method to find the nearest node based on the received node.

**Second commit**: This clean up the huge amount of tests I noticed in https://github.com/angular-eslint/angular-eslint/pull/243#issuecomment-749695110 (most of them copied from `contextual-lifecycle`).

**Third commit**: Fixes #246, before this commit this rule didn't report any `super <lifecycle method>()`, regardless of the context and also reports any class, even if the class isn't decorated by Angular decorators. Now it may be fixed.


JamesHenry (approved):
LGTM, thanks!

PR #328, opened 2021-02-06
fix(builder): pass all results to a single call to format

JamesHenry:
BREAKING CHANGE


JamesHenry:
Out of an abundance of caution, and because we are doing v2 very soon anyway, I am classing this as a breaking change in case anyone is depending on the current behaviour with some funky custom tooling. Weirder things have happened!

Issue #337, opened 2021-02-11
Expressions with nested pipes do not output all BindingPipes as AST nodes 


detljh:
In expressions with nested pipes, currently only the last pipe appears as an AST node. The other pipes will appear as an `exp` property of the AST node meaning when creating custom ESLint rules, using the `BindingPipe` node as the selector will not catch the inner pipes. 

```
convertAnnotatedSourceToFailureCase({
  description: 'should fail when async pipe is used with multiple pipes' ,
  annotatedSource: `
    {{ foo | async | number }}
        ~~~~~~~~~~~~~~~
  `,
  messageId: noAsyncPipe
}),
```

```
create(context) {
  return {
    BindingPipe(node: any) {
      console.log(node)
    }
  }
}
```

Outputs only one `BindingPipe` node, ie. the last `number` pipe.

```
BindingPipe {
  span: ParseSpan { start: 4, end: 24 },
  sourceSpan: AbsoluteSourceSpan { start: 11, end: 31 },
  nameSpan: AbsoluteSourceSpan { start: 25, end: 31 },
  exp: BindingPipe {
    span: ParseSpan { start: 4, end: 15 },
    sourceSpan: AbsoluteSourceSpan { start: 11, end: 22 },
    nameSpan: AbsoluteSourceSpan { start: 17, end: 22 },
    exp: PropertyRead {
      span: [ParseSpan],
      sourceSpan: [AbsoluteSourceSpan],
      nameSpan: [AbsoluteSourceSpan],
      receiver: [ImplicitReceiver],
      name: 'foo'
    },
    name: 'async',
    args: []
  },
  name: 'number',
  args: [],
  type: 'BindingPipe',
  parent: <ref *2> Interpolation {
    span: ParseSpan { start: 0, end: 28 },
    sourceSpan: AbsoluteSourceSpan { start: 7, end: 35 },
    strings: [ ' ', ' ' ],
    expressions: [ [Circular *1] ],
    type: 'Interpolation',
    parent: ASTWithSource {
      span: [ParseSpan],
      sourceSpan: [AbsoluteSourceSpan],
      ast: [Circular *2],
      source: ' {{ foo | async | number }} ',
      location: '<input>@1:6',
      errors: [],
      type: 'ASTWithSource',
      parent: [BoundText]
    }
  }
}

```

JamesHenry:
[...]

PR #338, opened 2021-02-12
fix(template-parser): add BindingPipe exp to VisitorKeys (#337)

detljh:
Added `BindingPipe: ['exp']` to `template-parser` for (#337)

### DIFF
--- a/packages/template-parser/src/index.ts
+++ b/packages/template-parser/src/index.ts
@@ -44,6 +44,7 @@ const KEYS: VisitorKeys = {
   Program: ['templateNodes'],
   PropertyRead: ['receiver'],
   Template: ['templateAttrs', 'children', 'inputs'],
+  BindingPipe: ['exp'],
 };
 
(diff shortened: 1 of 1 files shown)


JamesHenry (changes requested):
Thanks! Please also add some relevant test coverage in the template plugin which captures the source you initially discovered this issue on

detljh:
Sorry I'm not sure where I would add relevant tests

JamesHenry:
@detljh You reference tests from this repo in your original bug report - I'm simply referring to have an example of what was missing before and how you originally found this

detljh:
Oh in my original bug report those tests weren't from this repo. I found the bug while creating tests for a rule I wrote

JamesHenry:
@detljh Ah ok, well yeah I think we currently don't have anything in this repo which is affected by this, but I agree it makes sense to have, so I'm going to merge it

PR #351, opened 2021-02-28
feat(builder): add maxWarnings option

Lehoczky:
Implements [ESLint's max-warnings CLI option](https://eslint.org/docs/user-guide/command-line-interface#:~:text=quiet%20file.js-,%2D%2Dmax%2Dwarnings,with%20a%20success%20exit%20status.).

When migrating from TSLint to ESLint recently, we wanted to show linting errors only as warnings during development in vscode (as the TSLint extension does), but at the same time we also use the ng lint command in our CI and want to fail on these warnings.
At first, I thought about a `--strict` flag, but after reading [this ESLint thread](https://github.com/eslint/eslint/issues/2309), it looked best to just implement the max-warnings flag.

### DIFF
--- a/packages/builder/src/index.ts
+++ b/packages/builder/src/index.ts
@@ -98,7 +98,7 @@ async function run(
   const tooManyWarnings = maxWarnings >= 0 && totalWarnings > maxWarnings;
   if (tooManyWarnings && printInfo) {
     context.logger.error(
-      `Found too many warnings (maximum: ${options.maxWarnings}).`,
+      `Found ${totalWarnings} warnings, which exceeds your configured limit (${options.maxWarnings}). Either fix some of them or increase the value of the maxWarnings option.`,
     );
   }
 
--- a/packages/builder/src/index.ts
+++ b/packages/builder/src/index.ts
@@ -18,6 +18,7 @@ async function run(
   const projectName = context.target?.project || '<???>';
   const printInfo = options.format && !options.silent;
   const reportOnlyErrors = options.quiet;
+  const maxWarnings = options.maxWarnings;
 
   context.reportStatus(`Linting ${JSON.stringify(projectName)}...`);
   if (printInfo) {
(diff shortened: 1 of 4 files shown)


packages/builder/src/index.ts
```diff
@@ -94,8 +95,15 @@ async function run(
     context.logger.info('All files pass linting.\n');
   }
 
+  const tooManyWarnings = maxWarnings >= 0 && totalWarnings > maxWarnings;
+  if (tooManyWarnings && printInfo) {
+    context.logger.error(
+      `Found too many warnings (maximum: ${options.maxWarnings}).`,
```
JamesHenry, line 101: It would be good to include the number that was found too, and give the user a hint on what to do next e.g.:

"Your configured maxWarnings is 100, but 123 warnings were found. Either increase your maxWarnings or fix some of the lint warnings."

JamesHenry (changes requested):

packages/builder/src/index.ts
Lehoczky: Sure, sounds good to me. My original intention was to mimic the output of the ESLint CLI, but this can be an improvement over it.

packages/builder/src/index.ts
```diff
@@ -94,8 +95,15 @@ async function run(
     context.logger.info('All files pass linting.\n');
   }
 
+  const tooManyWarnings = maxWarnings >= 0 && totalWarnings > maxWarnings;
+  if (tooManyWarnings && printInfo) {
+    context.logger.error(
+      `Found ${totalWarnings} warnings, which exceeds your configured limit (${options.maxWarnings}). Either fix some of them or increase the value of the maxWarnings option.`,
```
JamesHenry, line 101: Hard to explain, but to me as a native English speaker the following inversion of the second sentence makes it come across less bluntly than your current text:

```suggestion
      `Found ${totalWarnings} warnings, which exceeds your configured limit (${options.maxWarnings}). Either increase your maxWarnings limit or fix some of the lint warnings.`,
```

JamesHenry: (You may well be a native English speaker too, I'm referencing that fact about me to justify why it's more of a feeling than anything else :) )

JamesHenry (approved):

PR #364, opened 2021-03-10
update eslint-plugin-prettier config to use correct parser

snebjorn:


JamesHenry (changes requested):
Thanks, just missed out the paragraph below referencing HTML as well

snebjorn:
> Thanks, just missed out the paragraph below referencing HTML as well

Oops, sorry. It has been corrected :)

JamesHenry (approved):

PR #379, opened 2021-03-17
Suggest removal of codelyzer in README

json-derulo:
Codelyzer is added by default to every Angular project. Some people who migrate their Angular project to ESLint might not know that `codelyzer` can be removed after removing TSLint. So I thought it would be better to suggest removing it in the README.


JamesHenry:
Thanks @json-derulo!

Issue #374, opened 2021-03-15
Probably wrong error reporting for "no-call-expression"


svzi:
Hi there!

First of all thank you very much for creating and maintaining this awesome plugin, I really appreciate it! 👏 

We used TSLint in our Angular project before and recently made the switch to ESLint. We nearly managed to get everything going like before, but this rule lets us scratching our heads:
`@angular-eslint/template/no-call-expression` (TSLint: `template-no-call-expression`)

Afaik this rule is supposed to prevent coding like this in the template:
```html
<div>{{ getTitle() }}</dv>
```
That's a bad practice because the function will be called on every change detection cycle. The TSLint implementation worked like a charm for us. Sadly the ESLint implementation doesn't. Because it will also report this piece of code as faulty:
```html
<div (click)="doSomething()">...
```
I don't think the linter should mark this code as an error, because this method won't be executed in every change detection cycle. It will only get called when the `click` event occurs.

Because I didn't find a way to configure this rule I was forced to deactivate it for our whole project. But we would love to  reactivate it, because it helped us a lot in the past.

Just in case this rule should be a replacement for the already mentioned TSLint rule, then I think here is something going wrong. But maybe I'm just misunderstanding the rule. Any feedback would be really much appreciated! 🙏 

Best,
Sven

JamesHenry:
Hi @svzi, we have an explicit unit test for the rule which asserts that there should be no error for event bindings: https://github.com/angular-eslint/angular-eslint/blob/d7c6aa49132100de23bd101f36ffccf6a8b6414e/packages/eslint-plugin-template/tests/rules/no-call-expression.test.ts#L24

This would suggest that there is something more complex going on. Please could you therefore create a full reproduction repo with the issue in for us to investigate.

Many thanks!

JamesHenry:
Please feel free to reopen this whenever you have chance to reproduce @svzi but as mentioned our current understanding is that this case is covered, so closing for now

hisomu:
I am also seeing a similar issues, however it's around forms

`<form [formGroup]="form" (ngSubmit)="form.valid && save()">`

rafaelss95:
> I am also seeing a similar issues, however it's around forms
> 
> `<form [formGroup]="form" (ngSubmit)="form.valid && save()">`

I was able to reproduce not only this case, but also with `||` and ternary. I submitted #390 that aims to fix this behavior.

PR #390, opened 2021-04-03
fix(eslint-plugin): no-call-expression incorrect reports for conditio…

rafaelss95:
…nals

Currently, the `no-call-expression` rule incorrectly reports the use of `methods` within `BoundEvent`, when preceded by conditionals, as related in https://github.com/angular-eslint/angular-eslint/issues/374#issuecomment-809386614.

### DIFF
--- a/packages/eslint-plugin-template/src/utils/get-nearest-node-from.ts
+++ b/packages/eslint-plugin-template/src/utils/get-nearest-node-from.ts
@@ -1,5 +1,5 @@
-import { AST } from '@angular/compiler';
-import { Node } from '@angular/compiler/src/render3/r3_ast';
+import type { AST } from '@angular/compiler';
+import type { Node } from '@angular/compiler/src/render3/r3_ast';
 
 type ASTOrNodeWithParent = (AST | Node) & { parent?: ASTOrNodeWithParent };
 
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-template/src/utils/get-nearest-node-from.ts
```diff
@@ -0,0 +1,23 @@
+import { AST } from '@angular/compiler';
```
JamesHenry, line 1: Please convert to type only import

packages/eslint-plugin-template/src/utils/get-nearest-node-from.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Node } from '@angular/compiler/src/render3/r3_ast';
```
JamesHenry, line 2: Please convert to type only import

packages/eslint-plugin-template/src/utils/get-nearest-node-from.ts
JamesHenry, line 1:
```suggestion
import type { AST } from '@angular/compiler';
```

packages/eslint-plugin-template/src/utils/get-nearest-node-from.ts
JamesHenry, line 2:
```suggestion
import type { Node } from '@angular/compiler/src/render3/r3_ast';
```

PR #391, opened 2021-04-03
refactor(eslint-plugin): component-max-inline-declarations no longer …

rafaelss95:
…accept negative values

BREAKING CHANGE.

As discussed in https://github.com/angular-eslint/angular-eslint/pull/313#discussion_r565800014, `component-max-inline-declarations` currently accepts negative values as input, but we just ignore this and use a default value instead.

With these changes, this rule no longer accept negative values.


JamesHenry:
@rafaelss95 So what happens after this change if someone has set a negative value today?

rafaelss95:
> So what happens after this change if someone has set a negative value today? 

The linter won't run properly, because the rule no longer accepts a negative value in its schema.

> Is there a recommended change/automated migration we can do for them?

We could perhaps create a migration to put the default values instead. That's the current behavior: if you pass a negative value, the rule ignores it and assume the default one.
A question: do we have any migration like this in repo already?

JamesHenry (changes requested):
I've changed the base branch to be the v3 branch, please rebase against its latest changes - that is where the current crop of breaking changes are being collected together.

Yes, on that branch, and previously for v2 I provided migration schematics, please add to the v3 one with the logic to update people's eslint configs if they are explicitly using negative numbers

packages/schematics/src/migrations/update-3-0-0/update-3-0-0.ts
```diff
@@ -118,6 +118,72 @@ function applyRecommendedExtraExtends(host: Tree) {
...
+  rule: Linter.RuleEntry | undefined,
+) {
+  if (!Array.isArray(rule) || rule.length !== 2) return;
+  const [, currentSchema] = rule;
+  rule[1] = Object.entries(currentSchema).reduce(
+    (accumulator, [key, value]) => {
+      return Number(value) < 0 ? accumulator : { ...accumulator, [key]: value };
+    },
+    {},
+  );
```
rafaelss95, line 131:
```json
**Before**:
"@angular-eslint/component-max-inline-declarations": [
  "error",
  {
    "animations": -1,
    "styles": 3,
    "template": 10
  }
],
```

**After**:
```json
"@angular-eslint/component-max-inline-declarations": [
  "error",
  {
    "styles": 3,
    "template": 10
  }
],
```

That's what we do currently: ignore the negative value and assume the default one.

packages/schematics/src/migrations/update-3-0-0/update-3-0-0.ts
```diff
@@ -118,6 +118,72 @@ function applyRecommendedExtraExtends(host: Tree) {
...
+}
+
+function updateComponentMaxInlineDeclarationsSchema({
+  overrides,
+  rules,
+}: Linter.Config) {
+  removeNegativeValuesFromComponentMaxInlineDeclarations(
+    rules?.['component-max-inline-declarations'],
```
rafaelss95, line 139: Do we need to specify it as `"@angular-eslint/component-max-inline-declarations"` or it just work as is? I just saw that in v2 migration, you used `use-pipe-decorator`...

rafaelss95:
While seeing the v2 migration, I found you used `use-pipe-decorator` there (https://github.com/angular-eslint/angular-eslint/blob/c80df75267f7f66db35035b12b7e40b253494118/packages/schematics/src/migrations/update-2-0-0/update-2-0-0.ts#L88), but shouldn't it be `@angular-eslint/use-pipe-decorator`?

JamesHenry:
@rafaelss95 Great catch, yes it should

JamesHenry (approved):
Thanks! Nice work on the migration

PR #392, opened 2021-04-04
feat(eslint-plugin-template): add rule accessibility-label-has-associ…

### LINKED ISSUE #288: About template/accessibility-label-for
I noticed some problems with `template/accessibility-label-for` and here are the suggestions (previously commented in #268):

1. It doesn't report labels without association, e.g.:

```html
<label for="nonExistentId">Label</label>
<!-- there's no element with `nonExistentId`, but the rule passes --> 
```

2. One could pass control component selectors through the `controlComponents` option, like this:

```json
"template-accessibility-label-for": [true, {
  "controlComponents": ["p-inputMask"],  
}]
```

```html
<app-label labelFor="username">Label<p-inputMask></p-inputMask></app-label>
```

...however one could have a component with an attribute that acts like an `id`, like [the `inputId` here, by example](https://www.primefaces.org/primeng/showcase/#/inputmask) and thus doesn't need to be wrapped by a label/custom label component. This case isn't covered by the rule, e.g.:

```html
<app-label labelFor="username">Label</app-label>
<p-inputMask formControlName="username" inputId="username" mask="(999) 999-9999? x99999"></p-inputMask>
```

3. The `labelAttributes` and `labelComponents` are stored in separated structures and it could be problematic, because a label attribute can cause conflicts depending on the label component you're using, e.g.:

```json
"template-accessibility-label-for": [true, {
  "labelAttributes": ["assoc", "elementId"],
  "labelComponents": ["app-label", "ngx-label"],  
}]
```

```html
<app-label assoc="id"></app-label> <!-- ok -->
<ngx-label assoc="id"></ngx-label > <!-- not ok, but the rule doesn't report -->
<app-label elementId="id"></app-label> <!-- not ok, but the rule doesn't report -->
```

<hr>

After thinking a little bit on how we could store the structures to meet these cases, I came up with this:

```js
"template-accessibility-label-for": [true, {
  "controlComponents": [
    { "inputs": ["id"], "selector": "input" }, // one of the defaults
    { "inputs": ["inputId"], "selector": "p-inputMask" },
    { "selector": "bs4-input" } // when it doesn't have `inputs`, it must be wrapped by a label/label component.
  ],
  "labelComponents": [
    { "inputs": ["for", "htmlFor"], "selector": "label" }, // default
    { "inputs": ["assoc"], "selector": "app-label" },
    { "inputs": ["elementId", "legacyId"], "selector": "ngx-label" },
    { "selector": "ng2-label" }, // when it doesn't have `inputs`, it must wrap a control component.
  ],
}]
```

To maintain the behavior in a schematic migration, the changes could be something like this:

```diff
"template-accessibility-label-for": ["error", {
- "controlComponents": ["p-inputMask", "bs4-input"],
- "labelAttributes": ["assoc", "elementId"],
- "labelComponents": ["app-label", "ngx-label"],
+ "controlComponents": [
+   { "selector": "p-inputMask" }, { "selector": "bs4-input" }
+ ],
+ "labelComponents": [
+   { "inputs": ["assoc", "elementId"], "selector": "app-label" },
+   { "inputs": ["assoc", "elementId"], "selector": "ngx-label" }
+ ],
}]
```

rafaelss95:
…ated-control

Hey @JamesHenry, I decided to deprecate `accessibility-label-for` and create a new rule because IMHO implement this new behavior under a rule named `label-for` could be weird as a label can be valid without `for` if it matches the configuration or even if wraps the input.

I also looked at the [`eslint-plugin-jsx-a11y#label-has-for`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md) and just noticed that it was replaced by [`eslint-plugin-jsx-a11y#label-has-associated-control`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md).

----

Another thing from #288:

> 1. It doesn't report labels without association, e.g.:

```html
<label for="nonExistentId">Label</label>
<!-- there's no element with `nonExistentId`, but the rule passes --> 
```

I decided not to implement this as part of this rule, as this rule can be implemented separately as for example [`eslint-plugin-jsx-a11y#control-has-associated-label`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/control-has-associated-label.md).

---

In any case, let me know your opinion on these statements.

Closes #288.


JamesHenry (approved):
Thanks @rafaelss95 I think this makes a lot of sense!

JamesHenry (dismissed):

JamesHenry (approved):

PR #394, opened 2021-04-04
fix(eslint-plugin): directive-class-suffix reporting selectorless directives

### LINKED ISSUE #353: directive-class-suffix rule not compatibly with Ivy requirements
Since Angular 9 and Ivy, undecorated abstract base classes which use Angular features have to be decorated with an empty `@Directive` (or `@Component` without template, but this seems to cause runtime issues at least in Angular 9). In many cases, those abstract classes are not really a Directive or Component and may therefore not have this Suffix. 

The examples provided by Angular would cause this rule to trigger: https://angular.io/guide/ivy-compatibility-examples#undecorated-classes

Shouldn't abstract classes be excluded from this rule?

rafaelss95:
1st. commit is the fix for #353.
2nd. commit is a refactor that:

- makes report message more-friendly;
- removes unnecessary tests and uses `data` property to ensure we pass the correct parameters to the user.

In any case, if you prefer, I could split it into 2 PRs.

Fixes #353.

### DIFF
--- a/packages/eslint-plugin/src/rules/directive-class-suffix.ts
+++ b/packages/eslint-plugin/src/rules/directive-class-suffix.ts
@@ -5,7 +5,7 @@ import {
   getClassName,
   getDeclaredInterfaceNames,
   getDecoratorPropertyValue,
-  getReadablePrefixes,
+  toHumanReadableText,
 } from '../utils/utils';
 
 type Options = [{ readonly suffixes: readonly string[] }];
@@ -72,7 +72,7 @@ export default createESLintRule<Options, MessageIds>({
             messageId: 'directiveClassSuffix',
             data: {
               className,
-              suffixes: getReadablePrefixes(allSuffixes),
+              suffixes: toHumanReadableText(allSuffixes),
             },
           });
         }
--- a/packages/eslint-plugin/src/rules/directive-class-suffix.ts
+++ b/packages/eslint-plugin/src/rules/directive-class-suffix.ts
@@ -5,19 +5,15 @@ import {
   getClassName,
   getDeclaredInterfaceNames,
   getDecoratorPropertyValue,
+  getReadablePrefixes,
 } from '../utils/utils';
 
-type Options = [
-  {
-    suffixes: string[];
-  },
-];
+type Options = [{ readonly suffixes: readonly string[] }];
 export type MessageIds = 'directiveClassSuffix';
 export const RULE_NAME = 'directive-class-suffix';
-
 const STYLE_GUIDE_LINK = 'https://angular.io/styleguide#style-02-03';
-
-const ValidatorSuffix = 'Validator';
+const DEFAULT_SUFFIXES = ['Directive'] as const;
+const VALIDATOR_SUFFIX = 'Validator';
 
 export default createESLintRule<Options, MessageIds>({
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/directive-class-suffix.ts
```diff
@@ -39,41 +39,40 @@ export default createESLintRule<Options, MessageIds>({
...
+          !allSuffixes.some((suffix) => className.endsWith(suffix))
         ) {
           context.report({
-            node: classParent.id ? classParent.id : classParent,
+            node: classParent.id ?? classParent,
             messageId: 'directiveClassSuffix',
             data: {
               className,
-              suffixes,
+              suffixes: getReadablePrefixes(allSuffixes),
```
JamesHenry, line 75: This is confusing now, please could we rename the `getReadablePrefixes` utility?

It is just a generic helper for listing some possible alternative values, so maybe we could call it:

`listPossibleValues()`

packages/eslint-plugin/tests/rules/directive-class-suffix.test.ts
```diff
@@ -24,48 +24,50 @@ ruleTester.run(RULE_NAME, rule, {
...
       selector: 'sgBarFoo'
     })
     class TestValidator implements AsyncValidator {}
-`,
+    `,
     `
-    @Directive
-    class TestDirective {}
-`,
+    @Directive()
```
JamesHenry, line 41: You seem to have removed the `@Directive` case, please add it back I think it's worth having just so we know how the linter behaves.

JamesHenry: Please also add a comment next to the new `@Directive()` case and reference back to the issue discussion

JamesHenry (changes requested):
See comments

packages/eslint-plugin/src/rules/directive-class-suffix.ts
rafaelss95: The rename makes perfect sense! What do you say about `toHumanReadableText`?

packages/eslint-plugin/tests/rules/directive-class-suffix.test.ts
rafaelss95: > You seem to have removed the @Directive case, please add it back I think it's worth having just so we know how the linter behaves.

Hmm, yes I've removed because it doesn't even compile as you can see here: 

Captura de Tela 2021-04-06 às 19 54 21

[**Repro link**](https://stackblitz.com/edit/angular-ivy-oipuw1)

But, let me know if you want me to add it back anyway.

> Please also add a comment next to the new `@Directive()` case and reference back to the issue discussion

Added the link, let me know if we need an additional description as well.

JamesHenry: The Angular tooling doesn't like it but it is valid TypeScript, so it's still good to understand how our tooling behaves when it exists, right?

rafaelss95: No problem. Added it back.

JamesHenry (approved):

PR #396, opened 2021-04-06
build: enforce type imports

rafaelss95:
Given some comments like https://github.com/angular-eslint/angular-eslint/pull/390#discussion_r607592984 and https://github.com/angular-eslint/angular-eslint/pull/386#discussion_r607589136, I'm submitting this PR that enables the rule [`@typescript-eslint/consistent-type-imports`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) to report and fix these imports automatically for us.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,5 +7,8 @@
     "plugin:@typescript-eslint/recommended",
     "prettier",
     "prettier/@typescript-eslint"
-  ]
+  ],
+  "rules": {
+    "@typescript-eslint/consistent-type-imports": "error"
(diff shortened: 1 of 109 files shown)


JamesHenry (approved):
Good thinking, thanks!

JamesHenry (changes requested):
Sorry the new rule that prompted this wasn't actually compliant it turns out, hopefully it's a quick update

JamesHenry (approved):

PR #399, opened 2021-04-07
refactor(eslint-plugin-template): no-negated-async no longer perform equality checks

### LINKED ISSUE #315: [no-negated-async]: false positives with equality report
When I have a template like this:

```HTML
<div *ngIf="(item$ | async) == 1">{{ item }}</div>
```

...with this configuration:

```
{
  "extends": [
    "plugin:@angular-eslint/template/recommended"
  ],
  "files": ["*.html"],
  "rules": {}
}
```

...I receive the following error message:

> Async pipes must use strict equality `===` when comparing with `false`

...that comes from here:

 https://github.com/angular-eslint/angular-eslint/blob/c8dd14c95f881593d7b6ca9f7ba80b37b7eaf40f/packages/eslint-plugin-template/src/rules/no-negated-async.ts#L25

...which is obviously incorrect, but what intrigues me even more is why we report the loose equality as IMO it's outside the scope of this rule. In any case, you may have more context on why we report this.

The possible fixes that I can see are:
- Report only `false` as the message indicates (but why only `false` and not `== null` or `== undefined` per example?);
- Report all falsy values (`false`, `0`, `-0`, `''`, `null`, `undefined`) and update message accordingly;
- Drop the equality report;

My choice would be the drop of the equality report, however as this rule is included in the _recommended_ config and also @mgechev mentioned sometime ago that they've introduced it in Google codebase I want to hear your opinion about this before touching in any code.

If the drop is a consesus, we could implement something like [**`eqeqeq`**](https://eslint.org/docs/rules/eqeqeq)/[**`triple-equals`**](https://palantir.github.io/tslint/rules/triple-equals) for templates ([that I've suggested in Codelyzer](https://github.com/mgechev/codelyzer/issues/643)) to handle the equality in a more general way.

cc @JamesHenry @mgechev

rafaelss95:
BREAKING CHANGE.

Currently it's a proposal. If it gets approved, we could implement a rule to perform equality checks.

Fixes #315.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/no-negated-async.ts
+++ b/packages/eslint-plugin-template/src/rules/no-negated-async.ts
@@ -1,11 +1,11 @@
-import type { Binary, PrefixNot } from '@angular/compiler';
+import type { BindingPipe } from '@angular/compiler';
 import {
   createESLintRule,
   ensureTemplateParser,
 } from '../utils/create-eslint-rule';
 
 type Options = [];
-export type MessageIds = 'noNegatedAsync' | 'noLooseEquality';
+export type MessageIds = 'noNegatedAsync';
 export const RULE_NAME = 'no-negated-async';
 
 export default createESLintRule<Options, MessageIds>({
@@ -13,17 +13,14 @@ export default createESLintRule<Options, MessageIds>({
   meta: {
     type: 'suggestion',
     docs: {
-      description:
-        'Ensures that strict equality is used when evaluating negations on async pipe output',
+      description: 'Ensures that async pipe results are not negated',
       category: 'Best Practices',
       recommended: 'error',
     },
     schema: [],
     messages: {
       noNegatedAsync:
-        'Async pipes should not be negated. Use (observable | async) === (false || null || undefined) to check its value instead',
-      noLooseEquality:
-        'Async pipes must use strict equality `===` when comparing with `false`',
+        'Async pipe results should not be negated. Use (observable | async) === (false || null || undefined) to check its value instead',
     },
   },
   defaultOptions: [],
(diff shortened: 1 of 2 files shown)


JamesHenry (changes requested):
Yeah seems like a good idea, and we'll definitely want to ensure we don't lose functionality overall. So:

- [x] add back the equality as a new rule (#444)
- [ ] change the base branch of this PR to ~v3~ `next-major-release/v12` and rebase
- [x] add a migration in for user configs
- [ ] prepare a PR for tslint-to-eslint-config so that future conversions apply both rules when converting the old codelyzer rule
- [ ] once the PR is merged, apply the new version of tslint-to-eslint-config to this branch
- [ ] Merge this PR into `next-major-release/v12`

JamesHenry:
I'd like to get v3 out as soon as we can so going to mark this one for v12 (the next major version to align us with the next major angular version)

JamesHenry:
@rafaelss95 No worries if not, but do you think you'll be able to work on this this week? Angular 12 is getting release today and I'd like to follow it up with @angular-eslint 12 stable as soon as we can.

This will be our last breaking change until Angular 13 so that we can stay aligned with Angular's own major version to make it really straightforward for users

rafaelss95:
> @rafaelss95 No worries if not, but do you think you'll be able to work on this this week? Angular 12 is getting released today and I'd like to follow it up with @angular-eslint 12 stable as soon as we can.
> 
> This will be our last breaking change until Angular 13 so that we can stay aligned with Angular's own major version to make it really straightforward for users

Oh, I'll try to do my best. PR for the new rule is open: #444. I'll try to work on a migration now.

JamesHenry:
@rafaelss95 Yes exactly, whatever it takes to achieve no change in behaviour when running `ng lint` on v4 vs v12

JamesHenry:
I have changed the base branch to be the v12 one, which already has the start of the migration in for v12 so you just need to add to that

JamesHenry:
I've turned my previous comment into a TODO list

JamesHenry:
@rafaelss95 Are you clear on the next steps above?

JamesHenry:
Thanks @rafaelss95 I've updated the list with the progess, I will open the PR to `tslint-to-eslint-config`

JamesHenry:
We can apply the updated tslint-to-eslint-config outside this branch, so please go ahead and get this PR ready for final review

JamesHenry (approved):

PR #427, opened 2021-04-29
fix(eslint-plugin-template): no-negated-async message tweak

Quraian:
A single vertical line means it's a pipe in Angular.


JamesHenry:
Thanks @Quraian!

Issue #202, opened 2020-11-27
[feat] angular templates: lint expressions in html


frederikdekegel:
Hi,

I added the default eslint eqeqeq rule to the *.component.html override as stated below

    {
      files: ['*.component.html'],
      parser: '@angular-eslint/template-parser',
      plugins: ['@angular-eslint/template'],
      rules: {
        eqeqeq: ['error', 'smart'],
        '@angular-eslint/template/banana-in-box': 'error',
        '@angular-eslint/template/no-negated-async': 'error',
      },
    }, 

Doing so, I thought, I could lint things like **<div class="col-4" *ngIf="isNewEmployee == 0">** => **<div class="col-4" *ngIf="isNewEmployee === 0">**

Unfortunately, this doesn't work. I guess that expressions in angular templates cannot be eslinted.

Is this a feature that could be added? For TS files everything is a completely setupped in eslint, so it is unfortunate that TS / JS code in the angular templates can pass through all quality checks we have set.

Thank you in advance.

JamesHenry:
Is this possible today in TSLint/Codelyzer?

frederikdekegel:
Hi James,

From the top of my head, I would say "No, it is not included in TSlint" though my experience with TSLint is limited. Is existance in TSLint /codelyzer a prerequisite for asking about a feature?

Kind regards
Frederik Dekegel

JamesHenry:
No, but naturally parity with the existing solution is the top priority so it’s useful context

mustafawm:
I'm not sure if this is related but...
I'm trying to get `eslint-plugin-jsx-a11y` to work but no luck yet (even after configuring eslint in VSCode to validate `.html` files.

Do you think what I'm doing is related to this issue?

```json
{
      "files": ["*.html"],
      "plugins": ["jsx-a11y"],
      "extends": ["plugin:@angular-eslint/template/recommended", "plugin:jsx-a11y/recommended"]
}
```

JamesHenry:
@mustafawm JSX and HTML are really very different things and so rules intended to run on JSX will not work on HTML.

mustafawm:
😅 first time i work on ng and totally missed that...though it's in the name of the package 

Thanks.

JamesHenry:
Hey folks, I'm going to close this one as it is simply not feasible to implement.

[...]

PR #444, opened 2021-05-13
feat(eslint-plugin-template): add rule eqeqeq

rafaelss95:
Relates to #202, #315 and #399.

~Waiting for #445 to mark it as ready for review.~

### DIFF
--- a/packages/eslint-plugin-template/src/rules/eqeqeq.ts
+++ b/packages/eslint-plugin-template/src/rules/eqeqeq.ts
@@ -1,0 +1,98 @@
+import type { AST, Binary } from '@angular/compiler';
+import { ASTWithSource, LiteralPrimitive } from '@angular/compiler';
+import {
+  createESLintRule,
+  ensureTemplateParser,
+} from '../utils/create-eslint-rule';
+import { getNearestNodeFrom } from '../utils/get-nearest-node-from';
+
+type Options = [{ readonly allowNilCheck?: boolean }];
+export type MessageIds = 'eqeqeq';
+export const RULE_NAME = 'eqeqeq';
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-template/src/rules/eqeqeq.ts
```diff
@@ -0,0 +1,98 @@
...
+      description: 'Requires `===` and `!==` in place of `==` and `!=`',
+      category: 'Best Practices',
+      recommended: false,
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allowNilCheck: {
```
JamesHenry, line 28: It's confusing to use the term "nil" in the context of a language which doesn't have it IMO. Please rename the option

JamesHenry (changes requested):

packages/eslint-plugin-template/src/rules/eqeqeq.ts
rafaelss95: Hmmm, what about: 

- `allowNullish` (ref: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/sort-type-union-intersection-members.md#options);
- `allowNullOrUndefined`;
- enter a suggestion.

JamesHenry: Either of those is good I think, "nullish" is a term in the language now thanks to the new operator, and null or undefined is descriptive :)

JamesHenry: Sorry missed your final point, I'm not sure to be honest, I don't have a great sense of which is more "idiomatic" for equivalent options in equivalent rules

PR #445, opened 2021-05-13
fix(template-parser): add missing `Conditional` and its keys to `VisitorKeys`

rafaelss95:
When working on #444, I realized that this key was missing and it was a little difficult to make the rules work correctly without a lot of effort.


JamesHenry (approved):
Great, thanks!

PR #459, opened 2021-05-15
feat(eslint-plugin): [prefer-output-readonly] add suggestion

rafaelss95:
To be reviewed after #458's merge.

Note that I went with a `suggest` instead of `fix`, because if we fix it directly we can break the code since there is no guarantee that `@Output` property is not being mutated somewhere.

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-output-readonly.ts
+++ b/packages/eslint-plugin/src/rules/prefer-output-readonly.ts
@@ -2,7 +2,7 @@ import type { TSESTree } from '@typescript-eslint/experimental-utils';
 import { createESLintRule } from '../utils/create-eslint-rule';
 
 type Options = [];
-export type MessageIds = 'preferOutputReadonly';
+export type MessageIds = 'preferOutputReadonly' | 'suggestFix';
 export const RULE_NAME = 'prefer-output-readonly';
 
 export default createESLintRule<Options, MessageIds>({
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/prefer-output-readonly.ts
```diff
@@ -19,17 +19,24 @@ export default createESLintRule<Options, MessageIds>({
     messages: {
       preferOutputReadonly:
         'Prefer to declare `@Output` as readonly since they are not supposed to be reassigned',
+      suggestFix: 'Add readonly modifier',
```
JamesHenry, line 22: I don't see any reason not to make the message key descriptive, it is what the eslint docs show in all their examples: https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

...and allows for multiple suggestions to more easily be provided if applicable.

```suggestion
      addReadonlyModifier: 'Add readonly modifier',
```

JamesHenry (changes requested):

packages/eslint-plugin/src/rules/prefer-output-readonly.ts
JamesHenry: (Note the suggestion was just an easy way to add the feedback, it would need to be applied in all locations across multiple files)

rafaelss95: Good idea, I went with `suggestAddReadonlyModifier` after seeing many cases with the `suggest` prefix in `@typescript-eslint` repo as https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/eslint-plugin/src/rules/no-non-null-assertion.ts#L22.

JamesHenry (approved):
Thanks a lot!

PR #475, opened 2021-05-18
feat(eslint-plugin): [no-input-rename] add option to allow some inputs

### LINKED ISSUE #423: [no-input-rename] Add option for whitelist of allowed input renames
I'd like to enable the `no-input-rename` rule in our component library codebase, but we have a lot of components that alias the input 'class' to the property `className` in order to be ES3 compatible. It essentially looks like:

```ts
@Input('class') className: string;
```

Can there be a whitelist of allowed input renames that could be specified in the rule's options? Perhaps something like:

```ts
'@angular-eslint/no-input-rename': ['error', 'class']

// or, potentially even better:

'@angular-eslint/no-input-rename': ['error', { allowedRenames: ['class'] }]
```

GavinWu1991:
Fix #423

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-input-rename.md
+++ b/packages/eslint-plugin/docs/rules/no-input-rename.md
@@ -1,0 +1,97 @@
+# Enforces directive input alias name (`no-input-rename`)
+
+This rule enforces the name for directive input alias with the property name, disallows renaming directive inputs by providing a string to the decorator.
+
+## Rule Details
+
+For details in the Angular styleguide see: [input](https://angular.io/styleguide#style-05-13).
+
+This rule has **NO** default options.
+
+## Options
+
+The schema for the options of this rule is:
+
+```ts
+interface Options {
+  // A list with allowed input names
+  readonly allowedNames?: readonly string[];
+}
+```
+
+## Examples
+
+In the examples below, we will use the following configuration:
+
+```json
+"rules": {
+  "@angular-eslint/no-input-rename": [
+    "warn",
+    {
+      "allowedNames": ["name"]
+    }
+  ]
+}
+```
+
+Examples of **incorrect** code:
+
+```ts
+import { Input } from '@angular/core';
+@Component({
+  selector: 'foo',
+})
+class TestComponent {
+  @Input('bar') label: string;
+}
+
+import { Input } from '@angular/core';
+@Component({
+  selector: 'foo',
+})
+class TestComponent {
+  @Input('bar') set label(label: string) {}
+}
+
+import { Input } from '@angular/core';
+@Component({
+  selector: 'foo',
+})
+class TestComponent {
+  @Input('foo') label: string;
+}
+```
+
+Examples of **correct** code:
+
+```ts
+import { Input } from '@angular/core';
+@Component
+class TestComponent {
+  @Input() label: string;
+}
+
+import { Input } from '@angular/core';
+@Directive({
+  selector: '[foo]',
+})
+class TestDirective {
+  @Input('foo') bar = new EventEmitter<void>();
+}
+
+import { Input } from '@angular/core';
+@Directive({
+  selector: '[foo], label2',
+})
+class TestDirective {
+  @Input() foo: string;
+}
+
+import { Input } from '@angular/core';
+@Directive({
+  selector: 'foo',
+})
+class TestDirective {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/no-input-rename.ts
```diff
@@ -10,7 +10,11 @@ import {
   isImportedFrom,
 } from '../utils/utils';
 
-type Options = [];
+type Options = [
+  {
+    allowedRenames: string[];
```
rafaelss95, line 15: I just checked some examples in `@typescript-eslint` repo ([**`no-this-alias`**](https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/eslint-plugin/src/rules/no-this-alias.ts#L10) and [**`explicit-module-boundary-types`**](https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/eslint-plugin/src/rules/explicit-module-boundary-types.ts#L19)) and saw that they usually call options like this as "allowedNames".

```suggestion
    readonly allowedNames?: readonly string[];
```

What do you think about this change?

packages/eslint-plugin/src/rules/no-input-rename.ts
```diff
@@ -64,13 +68,31 @@ export default createESLintRule<Options, MessageIds>({
...
     },
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allowedRenames: {
+            type: 'array',
+            items: {
+              type: 'string',
```
rafaelss95, line 78:
```suggestion
              type: 'string',
              uniqueItems: true,
```

packages/eslint-plugin/src/rules/no-input-rename.ts
```diff
@@ -64,13 +68,31 @@ export default createESLintRule<Options, MessageIds>({
...
+            },
+            description: 'A whitelist of allowed input renames',
```
rafaelss95, line 80: We generally don't fill this schema option, but if we do I would rephrase it to:

```suggestion
            description: 'A list with allowed input names',
```

packages/eslint-plugin/src/rules/no-input-rename.ts
```diff
@@ -93,7 +115,8 @@ export default createESLintRule<Options, MessageIds>({
           propertyAlias: string,
           propertyName: string,
         ): boolean => {
-          return !!(
+          return (
+            allowedRenames.indexOf(propertyAlias) !== -1 ||
```
rafaelss95, line 119:
```suggestion
            allowedRenames.includes(propertyAlias) ||
```

rafaelss95 (commented):
Thanks @GavinWu1991 for addressing this! I've left some minor comments, lmk what you think.

packages/eslint-plugin/src/rules/no-input-rename.ts
GavinWu1991: Thank you for your review, I found that the description of our rules is directly quoted from codelyzer's page, but there are many rules we have enhanced, so that why I just added this part in the schema. 
I think I can try to add an extra readme document in the `docs` directory for this rule since it already supports additional configuration options. I guess this would make it more intuitive for the user. 😀

packages/eslint-plugin/src/rules/no-input-rename.ts
GavinWu1991: Good suggestion, I like it.

GavinWu1991:
All the review findings have been remediated, and the description document created for `no-input-rename` rule.

packages/eslint-plugin/docs/rules/no-input-rename.md
```diff
@@ -0,0 +1,97 @@
...
+      "allowedNames": ["name"]
+    }
+  ]
+}
+```
+
+Examples of **incorrect** code:
+
+```ts
+import { Input } from '@angular/core';
```
rafaelss95, line 40: I think we can remove these imports, no?

rafaelss95 (approved):
Just one nit, but LGTM! 🚀

rafaelss95 (approved):

JamesHenry:
Thanks @GavinWu1991!

PR #478, opened 2021-05-18
feat(eslint-plugin): [no-host-metadata-property] add option to allow static values

### LINKED ISSUE #161: Rule Option Proposal: no-host-metadata-property allow-static
The rationale behind Angular Style [06-03](https://angular.io/guide/styleguide#style-06-03) doesn't really apply to static cases, where you want to set the property of a host to a constant value.  I'd like to propose optionally allowing an exception to the rule to only forbid dynamic host properties, indicated by brackets or parentheses, e.g. `[property]` or `(property)`.

Example configuration:
`'@angular-eslint/no-host-metadata-property': ['error', 'allow-static']`

Example allowed usage:
```ts
@Component({
    selector: 'my-component',
    templateUrl: 'my-component.html',
    host: { class: 'my-component' }
})
export class MyComponent {
```

As the alternative HostBinding in this case would set up unnecessary change detection, even though the property could be `readonly`.  Additionally, since it's something that any consumer of the component can rely on as being constant, it seems like metadata is an appropriate place to specify it.

rafaelss95:
Closes #161.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-host-metadata-property.ts
+++ b/packages/eslint-plugin/src/rules/no-host-metadata-property.ts
@@ -3,13 +3,14 @@ import { createESLintRule } from '../utils/create-eslint-rule';
 import { COMPONENT_OR_DIRECTIVE_CLASS_DECORATOR } from '../utils/selectors';
 import {
   AngularInnerClassDecorators,
-  getDecoratorPropertyValue,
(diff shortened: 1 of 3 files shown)


JamesHenry (approved):

MikeMatusz:
Thanks @rafaelss95!

PR #480, opened 2021-05-18
fix(template-parser): generate correct index.d.ts when building

### LINKED ISSUE #479: published `types` is incorrect

https://unpkg.com/browse/@angular-eslint/template-parser@12.0.0/dist/index.d.ts

**Versions**

| `@typescript-eslint       `        | `X.Y.Z` |
| `TypeScript`                       | `X.Y.Z` |
| `ESLint`                           | `X.Y.Z` |
| `node`                             | `X.Y.Z` |

```sh


```

JamesHenry:
Fixes #479

### DIFF
--- a/packages/template-parser/src/index.ts
+++ b/packages/template-parser/src/index.ts
@@ -246,7 +246,7 @@ function parseForESLint(code: string, options: { filePath: string }) {
   };
 }
 
-module.exports = {
+export default {
   parseForESLint,
   parse: function parse(code: string, options: { filePath: string }) {
     return parseForESLint(code, options).ast;
--- a/README.md
+++ b/README.md
@@ -815,4 +815,6 @@ If you see a rule below that has **no status** against it, then please feel free
 
 <!-- PR Links -->
 
+[`pr475`]: https://api.github.com/repos/angular-eslint/angular-eslint/pulls/475
+
 <!-- end rule list -->
--- a/packages/template-parser/tests/index.test.ts
+++ b/packages/template-parser/tests/index.test.ts
@@ -1,10 +1,13 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const { parseForESLint } = require('../src/index');
+/**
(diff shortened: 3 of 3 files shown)


packages/template-parser/src/index.ts
```diff
@@ -246,7 +246,7 @@ function parseForESLint(code: string, options: { filePath: string }) {
   };
 }
 
-module.exports = {
+export default {
```
JounQin, line 249: @JamesHenry 

Why not use `export { parseForESLint, parse }`?

So that it will also work with cjs + `@ts-check`?

image

JamesHenry: It's written in the comment on this PR:

```
/**
 * This usage seems a little wonky, but don't want to run any risk of breaking changes
 * to the compiled output of the template-parser node module, so leaving until v13.
 */
```

JamesHenry: This is not blocking anything AFAICT, you can achieve everything you need to via the default export and it will work with ts-check

JounQin: It doesn't, `require('@angular-eslint/template-parser')`

```js
// index.js
const _ = require('@angular-eslint/template-parser')

console.log(_.default)
console.log(_.parseForESLint)
```

```log
$ tsc -p jsconfig.json
index.js:11:15 - error TS2339: Property 'parseForESLint' does not exist on type 'typeof import("/workspace/angular-eslint-template-parser/node_modules/@angular-eslint/template-parser/dist/index")'.

11 console.log(_.parseForESLint)
```

```log
// run `node index.js`
undefined
[Function: u]
```

JounQin: > It's written in the comment on this PR:

The current solution is actually a breaking change for typings.

Why `export { parseForESLint }` would break too? I think it will be just fine.

JounQin: @JamesHenry Hi, any possible to fix?

PR #484, opened 2021-05-19
fix(schematics): skip e2e tsconfig for add to non e2e project

### LINKED ISSUE #453: Not existing `e2e/tsconfig.json` referenced for new Angular 12 app
When adding `angular-eslint` to a newly created Angular 12 app the `parserOptions` contains a reference to `e2e/tsconfig.json`, which does not exist since the Angular team removed a default e2e setup.

```
"parserOptions": {
        "project": [
          "tsconfig.json",
          "e2e/tsconfig.json"
        ],
```

Expected: Don't add the file if no e2e setup exists?

GavinWu1991:
fix #453 
The pr contains the change to determine if the project contains e2e content by checking for the presence of e2e architect in the `angular.json`, then skip adding e2e's tsconfig to`parserOptions.projects` if the target project is a non-e2e project.

### DIFF
--- a/packages/schematics/src/utils.ts
+++ b/packages/schematics/src/utils.ts
@@ -427,6 +427,5 @@ export function determineTargetProjectHasE2E(
   angularJSON: any,
   projectName: string,
 ): boolean {
-  const { architect } = angularJSON.projects[projectName];
-  return Object.prototype.hasOwnProperty.call(architect, 'e2e');
+  return !!getTargetsConfigFromProject(angularJSON.projects[projectName])?.e2e;
 }
--- a/packages/schematics/src/utils.ts
+++ b/packages/schematics/src/utils.ts
@@ -218,14 +218,18 @@ type ProjectType = 'application' | 'library';
 export function setESLintProjectBasedOnProjectType(
   projectRoot: string,
   projectType: ProjectType,
+  hasE2e?: boolean,
 ) {
   let project;
   if (projectType === 'application') {
     project = [
       `${projectRoot}/tsconfig.app.json`,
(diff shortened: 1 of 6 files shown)


packages/schematics/src/utils.ts
```diff
@@ -396,3 +418,15 @@ export function determineTargetProjectName(
...
+
+/**
+ * Checking if the target project has e2e setup
+ * Method will check if angular project architect has e2e configuration to determine if e2e setup
+ */
+export function determineTargetProjectHasE2E(
+  angularJSON: any,
+  projectName: string,
+): boolean {
+  const { architect } = angularJSON.projects[projectName];
```
JamesHenry, line 430: `architect` can also be called `targets`, we already have a utility function to normalize this within this file called `getTargetsConfigFromProject()` so you just need to use that here

packages/schematics/src/utils.ts
```diff
@@ -396,3 +418,15 @@ export function determineTargetProjectName(
...
+  return Object.prototype.hasOwnProperty.call(architect, 'e2e');
```
JamesHenry, line 431: Nit: we're dealing with a very simple object here - with its origins in `.json` - and no prototypal chain to consider, so a simple `!!architect.e2e` or `Boolean(architect.e2e)` would have also been fine

JamesHenry (changes requested):
Thanks! Just one point to change and one nit

packages/schematics/src/utils.ts
GavinWu1991: Got it, I simplified the determination logic. Thx

JamesHenry:
Thanks @GavinWu1991!

PR #486, opened 2021-05-19
feat(eslint-plugin-template): [no-any] add suggestion

rafaelss95:


packages/eslint-plugin-template/tests/rules/no-any.test.ts
```diff
@@ -43,43 +43,52 @@ ruleTester.run(RULE_NAME, rule, {
...
+        {
+          messageId: suggestRemoveAny,
+          output: `
+        <a [href]="getHref()">Click here</a>
+                   
       `,
-      messages: [
-        { char: '~', messageId },
-        { char: '^', messageId },
-        { char: '#', messageId },
```
rafaelss95, line 82: I removed this because in addition to looking redundant, since we already have these tests, I was unable to make multiple `suggestions` work at the same time.

JamesHenry: In the past I definitely remember this particular test being useful for catching location information regressions, so I am hesitant to change it...

Have you tried seeing if you can make it work with multiple suggestions using the raw RuleTester, i.e. removing the `convertAnnotatedSourceToFailureCase` util initially?

rafaelss95: Oh, I forgot to leave a message here saying that I found the cause for this... the fix is in #491.

packages/eslint-plugin-template/src/rules/no-any.ts
```diff
@@ -14,13 +14,15 @@ export default createESLintRule<Options, MessageIds>({
   meta: {
     type: 'suggestion',
     docs: {
-      description: `The use of "${ANY_TYPE_CAST_FUNCTION_NAME}" nullifies the compile-time benefits of the Angular's type system.`,
+      description: `The use of "${ANY_TYPE_CAST_FUNCTION_NAME}" nullifies the compile-time benefits of the Angular's type system`,
```
JamesHenry, line 17: I know you didn't make this change here, but makes sense to fix it while we are touching the message

```suggestion
      description: `The use of "${ANY_TYPE_CAST_FUNCTION_NAME}" nullifies the compile-time benefits of Angular's type system`,
```

JamesHenry (changes requested):

packages/eslint-plugin-template/src/rules/no-any.ts
JamesHenry: @rafaelss95 You marked this as resolved but you didn't make the change? 😅

rafaelss95: Ups, sorry, completely forgot this 😄

PR #489, opened 2021-05-19
feat(eslint-plugin-template): [accessibility-valid-aria] add suggestion

rafaelss95:


packages/eslint-plugin-template/tests/rules/accessibility-valid-aria.test.ts
```diff
@@ -47,7 +47,7 @@ ruleTester.run(RULE_NAME, rule, {
         <div aria-roledescriptio="text">Text</div>
```
JamesHenry, line 47: Hmm conceptually is the "fix" here really to remove what the user has written here?

Let's imagine this unit test is a real scenario, they clearly _tried_ to add a real attribute and just made a typo. Then they run linting with `--fix` and instead of having chance to review their mistake and make their markup more accessible for users, they just have an unannotated div...

Maybe we have to go for suggestion only on this rule for that reason?

JamesHenry (changes requested):

packages/eslint-plugin-template/tests/rules/accessibility-valid-aria.test.ts
rafaelss95: I think you have a point. In this case, let's wait for #491 that offers the support for multiple `suggestions` in `convertAnnotatedSourceToFailureCase `.

rafaelss95: @JamesHenry I just noticed that in [**`codelyzer#templateAccessibilityValidAriaRule.ts#L35`**](https://github.com/mgechev/codelyzer/blob/b2ec45be494f223e9615a07bfbb9fb7c68864dfc/src/templateAccessibilityValidAriaRule.ts#L35) we suggest the closest aria attribute possible, instead of removing it, based on [**`damerau-levenshtein`**](https://github.com/tad-lispy/node-damerau-levenshtein).

So my question is: should we add this lib to mimic this behavior or is it fine as is?

PR #501, opened 2021-05-23
feat(eslint-plugin): [use-component-view-encapsulation] add suggestion

rafaelss95:


packages/eslint-plugin/src/utils/utils.ts
```diff
@@ -362,6 +363,18 @@ export function getImplementsRemoveFix(
   return [];
 }
 
+export function getNodeToCommaRemoveFix(
```
rafaelss95, line 366: I couldn't think of a better name, please suggest, if you want to.

packages/eslint-plugin/src/rules/use-component-view-encapsulation.ts
```diff
@@ -2,52 +2,66 @@ import type { TSESTree } from '@typescript-eslint/experimental-utils';
...
+              messageId: 'suggestRemoveViewEncapsulationNone',
+              fix: (fixer) => {
+                const importDeclarations = getImportDeclarations(
+                  node,
+                  '@angular/core',
+                );
+
+                return [
+                  getNodeToCommaRemoveFix(sourceCode, node, fixer),
+                ].concat(
```
JamesHenry, line 54: Why is concat being used here? Why not a simple array including the return values of both functions?

JamesHenry (changes requested):

packages/eslint-plugin/src/rules/use-component-view-encapsulation.ts
rafaelss95: I used `.concat` because the return type of the`getImportRemoveFix` function can be either a single item or an `array`. See #518.

JamesHenry (changes requested):
Pending #518

JamesHenry:
#518 has been merged

JamesHenry (approved):

JamesHenry:
Thanks @rafaelss95!

PR #506, opened 2021-05-25
docs: add more package versions in issues templates

rafaelss95:
While doing this, I also noticed that we could improve this by adding repro section, that includes configuration and use-case(s). WDYT?

### DIFF
--- a/.github/ISSUE_TEMPLATE/builder.md
+++ b/.github/ISSUE_TEMPLATE/builder.md
@@ -32,7 +32,7 @@ Please consider creating an isolated reproduction repo to make it easy for the v
 
 | package                            | version |
 | ---------------------------------- | ------- |
-| `@typescript-eslint       `        | `X.Y.Z` |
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/builder.md
```diff
@@ -32,7 +32,7 @@ Please consider creating an isolated reproduction repo to make it easy for the v
 
 | package                            | version |
 | ---------------------------------- | ------- |
-| `@typescript-eslint       `        | `X.Y.Z` |
+| `@angular-eslint/builder`          | `X.Y.Z` |
 | `TypeScript`                       | `X.Y.Z` |
```
rafaelss95, line 36: `Typescript` is already informed when you run `ng version` (something that is done just below), so I don't know if it's necessary to be redundant here.

JamesHenry:
@rafaelss95 yep happy to further enhance the templates with relevant stuff

PR #518, opened 2021-05-28
refactor(utils): change return type of fixer utilities

rafaelss95:
Currently, if it isn't possible to determine a fix in utility functions, we simply return an empty `array` (`TSESLint.RuleFix[]`) and this can be a little confuse, as we need to encapsulate the fixers in `arrays` and use `.concat` (which accepts either a single item or an `array`). See https://github.com/angular-eslint/angular-eslint/pull/501#discussion_r641589200.

In this PR, the proposal is to return `undefined` instead and whenever we use it in `report#fix` (which accepts `IterableIterator<RuleFix> | TSESLint.RuleFix | TSESLint.RuleFix[] | null`), we call the` isNotNullOrUndefined` helper to extract only the _non-undefined_ fixes from the `array` and make the compiler happy.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/eqeqeq.ts
+++ b/packages/eslint-plugin-template/src/rules/eqeqeq.ts
@@ -110,10 +110,10 @@ const getFix = ({
   start: number;
   end: number;
   fixer: TSESLint.RuleFixer;
-}): TSESLint.RuleFix | TSESLint.RuleFix[] => {
+}): TSESLint.RuleFix | null => {
   const { source } = getNearestNodeFrom(node, isASTWithSource) ?? {};
 
-  if (!source) return [];
+  if (!source) return null;
 
   return fixer.insertTextAfterRange(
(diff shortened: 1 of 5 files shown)


JamesHenry (changes requested):
Please could we change it to be `null` to represent the intentional "non-useful" value? I feel it is a little more explicit (as an unintentional `undefined` return value is easily done)

rafaelss95:
1. You mean that we should use `| null` instead? If so, no problem. I just opted by `| undefined` to conform to the rest of the methods.

2. While `null` is accepted, `Type '(RuleFix | null)[]' is not assignable to type 'RuleFix | RuleFix[] | IterableIterator<RuleFix> | null'.`, so we need to use a type-guard function to extract only truthy values from an array.

JamesHenry:
Thanks @rafaelss95 that makes sense!

JamesHenry (approved):

PR #520, opened 2021-05-29
feat(builder): add `cacheStrategy` option

### LINKED ISSUE #410: Add support to --cache-strategy option
Please, add [`--cache-strategy` option](https://eslint.org/docs/user-guide/command-line-interface#options) to `ng lint`. Option `--cache` works fine. I have `@angular-eslint/builder` up to `2.1.0`.

**Versions**

```sh
                |___/
    

Angular CLI: 11.2.8
Node: 14.16.0
OS: win32 x64

Angular: 11.2.9
... platform-browser, platform-browser-dynamic, platform-server
... router, service-worker
Ivy Workspace: Yes

@angular-devkit/architect       0.1102.7
@angular-devkit/build-angular   0.1102.7
@angular-devkit/core            11.2.7
@angular-devkit/schematics      11.2.8
@angular/cdk                    11.2.8
@angular/cli                    11.2.8
@angular/fire                   6.1.4
@angular/material               11.2.8
@nguniversal/builders           11.2.1
@nguniversal/express-engine     11.2.1
@schematics/angular             11.2.8
@schematics/update              0.1102.8
rxjs                            6.6.7
typescript                      4.1.5
```
- [X] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

rafaelss95:
Closes #410.

~Waiting for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53485.~

### DIFF
--- a/packages/builder/src/schema.d.ts
+++ b/packages/builder/src/schema.d.ts
@@ -10,6 +10,7 @@ export interface Schema extends JsonObject {
   fix: boolean;
   cache: boolean;
   cacheLocation: string | null;
+  cacheStrategy: 'content' | 'metadata' | null;
   eslintConfig: string | null;
   ignorePath: string | null;
 }
--- a/packages/builder/src/utils/eslint-utils.ts
+++ b/packages/builder/src/utils/eslint-utils.ts
@@ -26,6 +26,7 @@ export async function lint(
     fix: !!options.fix,
     cache: !!options.cache,
     cacheLocation: options.cacheLocation || undefined,
+    cacheStrategy: options.cacheStrategy || undefined,
     /**
      * Default is `true` and if not overridden the eslint.lintFiles() method will throw an error
      * when no target files are found.
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@nrwl/tao": "11.2.10",
     "@nrwl/workspace": "11.2.10",
     "@schematics/angular": "12.0.0",
-    "@types/eslint": "^7.2.0",
+    "@types/eslint": "^7.2.1",
     "@types/eslint-scope": "^3.7.0",
     "@types/jest": "^26.0.15",
     "@types/node": "^10.12.2",
--- a/packages/builder/src/schema.json
+++ b/packages/builder/src/schema.json
@@ -22,6 +22,12 @@
       "type": "string",
       "description": "Path to the cache file or directory."
     },
+    "cacheStrategy": {
+      "type": "string",
+      "description": "Strategy to use for detecting changed files in the cache.",
+      "default": "metadata",
+      "enum": ["metadata", "content"]
+    },
     "force": {
       "type": "boolean",
       "description": "Succeeds even if there was linting errors.",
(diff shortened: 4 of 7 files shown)


packages/builder/src/schema.d.ts
```diff
@@ -10,6 +10,7 @@ export interface Schema extends JsonObject {
   fix: boolean;
   cache: boolean;
   cacheLocation: string | null;
+  cacheStrategy: 'content' | 'metadata' | null;
```
JounQin, line 13: Shouldn't `null` be `undefined` here? Same as `cacheLocation` above.

rafaelss95: The problem is that `Schema` extends [`JsonObject`](https://github.com/angular/angular-cli/blob/37a06a7c37f5b4286d58b475e6e12c86f00fac5b/packages/angular_devkit/core/src/json/interface.ts#L57-L59), which doesn't accept `undefined`.

JounQin: Can optional property `?:` be used?

rafaelss95: No, because it has the same effect of `| undefined`.

JounQin: I mean add it into `JsonObject`.

packages/builder/src/utils/eslint-utils.ts
```diff
@@ -22,10 +22,11 @@ export async function lint(
   const eslint = new projectESLint.ESLint({
     useEslintrc: true,
     overrideConfigFile: eslintConfigPath,
-    ignorePath: options.ignorePath || undefined,
-    fix: !!options.fix,
-    cache: !!options.cache,
-    cacheLocation: options.cacheLocation || undefined,
+    ignorePath: options.ignorePath ?? undefined,
```
JamesHenry, line 25: Why change these to use `??` instead of `||`?

`||` was already doing everything we needed it to, and now technically with your change `0`, `NaN` and `''` will be set preferentially to `undefined` for these values, which is not desired. It may be unlikely that those values make their way through the code paths here, but there needs to be an objective reason to make this change in the first place IMO, and it seems objectively less desirable to me.

JamesHenry (changes requested):

packages/builder/src/utils/eslint-utils.ts
rafaelss95: You're right. I'll revert this 😃

packages/builder/src/utils/eslint-utils.ts
```diff
@@ -22,10 +22,11 @@ export async function lint(
...
+    fix: options.fix,
+    cache: options.cache,
+    cacheLocation: options.cacheLocation ?? undefined,
+    cacheStrategy: options.cacheStrategy ?? undefined,
```
rafaelss95, line 29:
```suggestion
    ignorePath: options.ignorePath || undefined,
    fix: options.fix,
    cache: !!options.cache,
    cacheLocation: !!options.cacheLocation || undefined,
    cacheStrategy: options.cacheStrategy || undefined,
```

JamesHenry (approved):

JamesHenry:
Thanks @rafaelss95!

Issue #523, opened 2021-05-31
[no-output-native] False positive when EventEmitter's generic type is an object with a property that is named like a native event


cypressious:
**Description and reproduction of the issue**

The rule `no-output-native` produces a false positive if the generic type of the `EventEmitter` is a complex object type with at least one property that is named like a native event according to https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/rules/no-output-native.ts#L49-L219.

```ts
// The following all produce a false positive

@Output() foo = new EventEmitter<{ click: string }>();
@Output() foo = new EventEmitter<{ blur: string }>();
@Output() foo = new EventEmitter<{ bar: string, blur: string }>();

// The following don't produce a false positive

@Output() foo = new EventEmitter<{ bar: string }>();
@Output() foo = new EventEmitter<{}>();
@Output() foo = new EventEmitter();
@Output() foo = new EventEmitter<ComplextObject>();

type ComplextObject = { click: string };
```

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@angular-eslint/eslint-plugin`    | `12.1.0` |
| `@typescript-eslint/parser`        | `4.25.0` |
| `TypeScript`                       | `4.2.3` |
| `ESLint`                           | `7.27.0` |
| `node`                             | `v14.17.0` |

```sh
# Please run `npx ng version` in your project and paste the full output here:

Angular CLI: 12.0.2
Node: 14.17.0
Package Manager: yarn 1.22.5
OS: win32 x64

Angular: 12.0.2
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, localize, material
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1200.2
@angular-devkit/build-angular   12.0.2
@angular-devkit/core            12.0.2
@angular-devkit/schematics      12.0.2
@schematics/angular             12.0.2
rxjs                            7.1.0
typescript                      4.2.3

```

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

PR #525, opened 2021-05-31
fix(eslint-plugin): [no-output-on-prefix] correct false positives

rafaelss95:
This PR basically fixes the same problems related in #523, but for `no-output-on-prefix`.

@JamesHenry while doing this I just noticed that we aren't checking the `outputs` metadata property. Example:

```ts
@Component({
  outputs: ['onChange'],
})
class Test {}
```

Should we support it here and in other `*-output-*` rules? If so, I can work on this on another PRs.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-output-on-prefix.ts
+++ b/packages/eslint-plugin/src/rules/no-output-on-prefix.ts
@@ -13,7 +13,7 @@ export default createESLintRule<Options, MessageIds>({
   meta: {
     type: 'suggestion',
     docs: {
-      description: `Ensures that Output bindings, including aliases, are not named "on", nor prefixed with it. See more at ${STYLE_GUIDE_LINK}`,
+      description: `Ensures that output bindings, including aliases, are not named "on", nor prefixed with it. See more at ${STYLE_GUIDE_LINK}`,
       category: 'Best Practices',
       recommended: 'error',
     },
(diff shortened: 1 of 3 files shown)


rafaelss95:
Idk why the snapshot test is failing as I didn't touch the error message.

JamesHenry:
@rafaelss95 I made some major changes to integration tests to allow them to be run in parallel. Those changes combined with Nx Cloud Distributed CI has cut the total time for CI in half which is pretty amazing.

In light of these changes, it might be worth seeing if you can finally now run the integration tests on your local machine:

You can try the following to run them all serially (will take some time but hopefully it will work for you):

```sh
yarn integration-tests
```

You can just add `-u` at the end to run jest in update mode and update snapshots whenever required. I also kept the alias for this:

```sh
yarn update-integration-tests
```

JamesHenry:
And with regards to your comment:

> Idk why the snapshot test is failing as I didn't touch the error message.

It's because you have changed the location of the error message, it is a legitimate failure:

image

JamesHenry:
> while doing this I just noticed that we aren't checking the outputs metadata property. Example:

Hmm yeah interesting... Can you think of any reason why this wasn't done in Codelyzer this whole time?

rafaelss95:
> @rafaelss95 I made some major changes to integration tests to allow them to be run in parallel. Those changes combined with Nx Cloud Distributed CI has cut the total time for CI in half which is pretty amazing.
> 
> In light of these changes, it might be worth seeing if you can finally now run the integration tests on your local machine:
> 
> You can try the following to run them all serially (will take some time but hopefully it will work for you):
> 
> ```shell
> yarn integration-tests
> ```
> 
> You can just add `-u` at the end to run jest in update mode and update snapshots whenever required. I also kept the alias for this:
> 
> ```shell
> yarn update-integration-tests
> ```

Impressive! That's really faster and I also noticed that my node version was 15, which is greather than the maximum suported 14. Everything should be fine now, thanks!



> And with regards to your comment:
> 
> > Idk why the snapshot test is failing as I didn't touch the error message.
> 
> It's because you have changed the location of the error message, it is a legitimate failure:
> 
> image

Danm, I didn't noticed that the location was different 😞 

> > while doing this I just noticed that we aren't checking the outputs metadata property. Example:
> 
> Hmm yeah interesting... Can you think of any reason why this wasn't done in Codelyzer this whole time? I wonder if it was intentional but I'm not sure why that would be the case

I imagine that's the case of just something that was never requested because it's less used than traditional decorators, but 🤷‍♂️

JamesHenry:
Cool, fair enough, the yeah I think we should address the `outputs: []` point too.

I strongly recommend you install https://volta.sh - it's absolutely flawless and node, npm and yarn version management.

All you need to do is CD into a project and it will magically switch your versions of those binaries to the right ones for that project. You can see the pinned versions we have in package.json (that's what it uses as its source of truth for a project, otherwise it defaults to LTS).

We also use volta in CI so that it never gets out of sync with local development

PR #545, opened 2021-06-17
fix(eslint-plugin-template): include more checks for `isHiddenFromScreenReader`

rafaelss95:
This PR aims to add more checks for `isHiddenFromScreenReader`.

Now it considers the following as hidden:

- It has `display: none` or `visibility: hidden` style;
- It has `aria-hidden` or `hidden` attribute;
- It's an `<input type="hidden">`;
- One of its ancestors met one of the following situations above.


packages/eslint-plugin-template/src/utils/is-presentational-role.ts
```diff
@@ -1,10 +0,0 @@
-import type { TmplAstElement } from '@angular/compiler';
-import { PROPERTY } from './constants';
-import { getAttributeValue } from './get-attribute-value';
-
-const presentationRoles = new Set(['presentation', 'none', PROPERTY]);
-
-export function isPresentationRole(node: TmplAstElement): boolean {
```
rafaelss95, line 7: This file was named incorrectly and that's why I changed to `is-presentation-role`.

packages/eslint-plugin-template/src/utils/get-original-attribute-name.ts
```diff
@@ -0,0 +1,63 @@
+import type {
+  TmplAstBoundAttribute,
+  TmplAstBoundEvent,
+  TmplAstTextAttribute,
+} from '@angular/compiler';
+import { BindingType, ParsedEventType } from '@angular/compiler';
+
+// TODO: It seems like `TmplAstBoundAttribute.type` is incorrectly typed at https://github.com/angular/angular/blob/2d1347b2ceef8c1e564c0f7cfe3a0a740b915e80/packages/compiler/src/render3/r3_ast.ts#L64.
+type BoundAttribute = Omit<TmplAstBoundAttribute, 'type'> &
```
rafaelss95, line 9: `TmplAstBoundAttribute.type` is declared as `BindingType` (as you can see [here](https://github.com/angular/angular/blob/2d1347b2ceef8c1e564c0f7cfe3a0a740b915e80/packages/compiler/src/render3/r3_ast.ts#L64)), however its type should be `BoundAttribute`, no?

The same is valid for `TmplAstBoundEvent` below.

packages/eslint-plugin-template/src/utils/get-attribute-value.ts
```diff
@@ -13,35 +15,36 @@ import { PROPERTY } from './constants';
...
 //        ~~~~~~~~~~~~~~~~~~~~~
-export function getAttributeValue<T extends string | number | boolean = string>(
-  node: TmplAstElement,
+export function getAttributeValue(
+  { attributes, inputs }: TmplAstElement,
   attributeName: string,
-): T | string | null | typeof PROPERTY {
-  const attribute = node.attributes.find(
-    (attribute: TmplAstTextAttribute) => attribute.name === attributeName,
+): unknown {
```
rafaelss95, line 21: I used `unknown` here because the generic can lead us to a mistake as you can't really know what type is present in `value`. The best way to check this is checking the result via `typeof` or `instanceof`.

packages/eslint-plugin-template/src/utils/get-original-attribute-name.ts
```diff
@@ -0,0 +1,63 @@
...
+  }>;
+
+/**
+ * Returns the original attribute name, respecting its `__originalType`.
+ * @example
+ * ```html
+ * <!-- The AST returns the name as "display", but we want the original name: "style.display.none" -->
+ * <div [style.display.none]="test"></div>
+ * ```
+ */
```
rafaelss95, line 29: I'm not good on documenting function so please suggest a better one if you want. (Here and elsewhere).

rafaelss95:
@JamesHenry, I was thinking about creating tests for functions like these, however it would be a pain to mock element, attributes, inputs and so on. Do you know a better way to test only functions?

packages/eslint-plugin-template/src/utils/get-original-attribute-name.ts
JamesHenry: Question for the Angular Team that one :)w

rafaelss95: I'll open an issue there asap... for now I've simplified this method as I noticed that we can rely on `.keySpan.name` to get the full name (except in two-way data binding cases, which are already being handled properly).

JamesHenry: Are you wanting this PR to be reviewed now then?

JamesHenry:
Hi @rafaelss95 just checking you saw this: https://github.com/angular-eslint/angular-eslint/pull/545#discussion_r663486992

packages/eslint-plugin-template/src/utils/get-original-attribute-name.ts
rafaelss95: OMG, I totally missed your message here. Sorry about that, as I changed this implementation, the AST point doesn't matter anymore... so you can revise it whenever you want :)

JamesHenry (approved):

PR #558, opened 2021-06-27
docs(eslint-plugin-template): [no-negated-async] improve message

yujiberra:
(Hoping I'm titling this PR correctly)

Having `Use (observable | async) === (false || null || undefined) to check its value instead'` in the message is not great because if someone interprets it literally it's equivalent to `(observable | async) === undefined`, which may not be what is desired. Also, I don't think it makes sense to suggest comparing to null, since that doesn't solve the issue this rule is trying to solve.

As an aside, this rule is a bit weird because it's easy to imagine situations where excluding null from the truth condition would cause more unnecessary rendering than not doing so, e.g.,
`<some-component [dontRenderALotOfStuff]="(observable | async) === false"></some-component>`
But this rule does at least teach people (like me) that the async pipe evaluates to null until the observable emits its first value, so I don't dispute that it has value.


rafaelss95:
> Having `Use (observable | async) === (false || null || undefined) to check its value instead'` in the message is not great because if someone interprets it literally it's equivalent to `(observable | async) === undefined`, which may not be what is desired. Also, I don't think it makes sense to suggest comparing to null, since that doesn't solve the issue this rule is trying to solve.

Excuse me, could you please clarify why comparing to `null` deviates from the rule's purpose?

> As an aside, this rule is a bit weird because it's easy to imagine situations where excluding null from the truth condition would cause more unnecessary rendering than not doing so, e.g.,
> `<some-component [dontRenderALotOfStuff]="(observable | async) === false"></some-component>`

What would you expect the linter to suggest if this example was violating the rule like:

`<some-component [dontRenderALotOfStuff]="!(observable | async)"></some-component>`

yujiberra:
> Excuse me, could you please clarify why comparing to `null` deviates from the rule's purpose?

Certainly - as I understand it, the purpose of the rule prevent unnecessary rendering before the observable emits a value, in a case like
`<*ngIf="!(observable | async)"><big-component></big-component></ngIf>`
where `big-component` always will be rendered until `observable` emits a value. Changing the ngIf condition to `(observable | async) === null` wouldn't fix this. I certainly don't think the linter should have a rule _against_ `(observable | async) === null`, but it does feel weird (to me) to include it as a suggestion.

> What would you expect the linter to suggest if this example was violating the rule like:
> 
> `<some-component [dontRenderALotOfStuff]="!(observable | async)"></some-component>`
This is a difficult question; it's possible that there isn't a good way to improve this case since there's no way for the linter to tell which truth value of `[dontRenderALotOfStuff]` is going to be more expensive. I also don't think the current behavior is that bad though: the rule as it stands serves the purpose of letting developers know that the async pipe returns null until the observable first emits a value, and discourages them from lumping together all falsy values without thinking about the implications.

rafaelss95:
> Certainly - as I understand it, the purpose of the rule prevent unnecessary rendering before the observable emits a value, in a case like
> `<*ngIf="!(observable | async)"><big-component></big-component></ngIf>`
> where `big-component` always will be rendered until `observable` emits a value. Changing the ngIf condition to `(observable | async) === null` wouldn't fix this. I certainly don't think the linter should have a rule _against_ `(observable | async) === null`, but it does feel weird (to me) to include it as a suggestion.

While you're right about the rule purpose, I don't get why you feel awkward the with `=== null`... how would you replace negation in this case then? For me, `null` is the most suitable suggestion for this case, since I want `<big-component />` to be rendered until `observable` emits a _truthy_ value.

Also note that users (devs) are the only ones who can determine which suggestion should be applied depending on the context. We offer the most commonly used values ('' and 0 are not included) ​​in order to help resolve reported issues and maybe save you some time typing.

yujiberra:
> While you're right about the rule purpose, I don't get why you feel awkward the with `=== null`... how would you replace negation in this case then? For me, `null` is the most suitable suggestion for this case, since I want `<big-component />` to be rendered until `observable` emits a _truthy_ value.

Ah right that makes sense, I had forgotten about the case where the observable is usually falsy, in which case `=== null` would minimize unnecessary rendering changes.

I'll update the PR so that it just changes the `=== (false || null || undefined)` part.

packages/eslint-plugin-template/src/rules/no-negated-async.ts
```diff
@@ -25,7 +25,7 @@ export default createESLintRule<Options, MessageIds>({
     schema: [],
     messages: {
       noNegatedAsync:
-        'Async pipe results should not be negated. Use (observable | async) === (false || null || undefined) to check its value instead',
+        'Async pipe results should not be negated. Use (observable | async) === false, (observable | async) === null, or (observable | async) === undefined to check its value instead',
```
rafaelss95, line 28: While you are at it, could you please add the backticks?

```suggestion
        'Async pipe results should not be negated. Use `(observable | async) === false`, `(observable | async) === null` or `(observable | async) === undefined` to check its value instead',
```

yujiberra: Good call, done!

JamesHenry (approved):
Thanks @yujiberra

PR #568, opened 2021-07-04
fix(eslint-plugin): [no-output-rename] handle `getters` and `outputs` metadata property

rafaelss95:
This PR implements the checks for both `getters` and `outputs` metadata property for `no-output-rename` rule.

In addittion to this, I've added suggestion and fix (in case of property name === alias name).


JamesHenry (changes requested):
Conflicts after merging your previous PR

JamesHenry (approved):

PR #570, opened 2021-07-05
refactor(eslint-plugin): [no-outputs-metadata-property] switch to selectors

rafaelss95:
Basically changes the implementation to use selectors. In addition to this, I added some tests to ensure we only report non-computed properties.


JamesHenry (approved):

JamesHenry (changes requested):
Integration test snapshots need updating after changing message

JamesHenry (approved):

PR #581, opened 2021-07-10
feat: add outputFile to builder (#349)

### LINKED ISSUE #349: builder: Add option to write output of ESLint also to file
It would be very cool to also be able to write the output of ESLint to a file to pass this on to other tools.

AFAIK, this should easily be possible by adding an option to the schema and, depending on it's value, also writing a file in, the same as how it's currently logged:
https://github.com/angular-eslint/angular-eslint/blob/4b6abed4a96cb29698a5de871f367974e5a3a79e/packages/builder/src/index.ts#L77

As a nice extra feature, there could also be another option to set the formatter for file output different than console output. This way, one could have an easy to read console output and easy to parse file output.

If someone can confirm that I didn't miss anything, I would be happy to submit a PR for this. 🙂

cstenglein:
This adds the `outputFile` to the schema and thus closes #349

For the code style I leaned on to https://github.com/nrwl/nx/blob/c7a5f6d7dd04f85386465dd0b7d86bcc74daa4ba/packages/linter/src/executors/eslint/lint.impl.ts#L108 with the exception that i use the `recursive` option of `mkdirSync` which doesn't require a custom recursive function. (Thx to @FrozenPandaz for the basic functionality!)

What I'm not sure about if it should default to JSON output for the written file. Currently it writes the specified format to the outputFile which could lead to confusion.

If anything is missing / should be changed please tell me :)

### DIFF
--- a/packages/builder/src/index.ts
+++ b/packages/builder/src/index.ts
@@ -2,6 +2,7 @@ import type { BuilderContext, BuilderOutput } from '@angular-devkit/architect';
 import { createBuilder } from '@angular-devkit/architect';
 import type { ESLint } from 'eslint';
 import path from 'path';
+import fs from 'fs';
 import type { Schema } from './schema';
 import { lint, loadESLint } from './utils/eslint-utils';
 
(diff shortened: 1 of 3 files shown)


cstenglein:
Closing since I was too late and the code was refactored.

Will open another PR with the same changes for the refactored code :P

JamesHenry:
Thank you for contributing @cstenglein. It’s a crazy coincidence that you would try and add this today, because I have been laying the groundwork for it and will be adding it in tomorrow along with a number of other features

JamesHenry:
Thank you, as I say though this is already in hand. I’m grateful for your contribution though!

For future reference for other thing you may work on when adding new features we would also require tests before something would be eligible.

Vielen Dank nochmal!

cstenglein:
Oh sorry, didn't know that!

Perfect, then thank you for the fast reply @JamesHenry ! :)

JamesHenry:
Merged in the feature, thanks again for contributing and sorry for not making it clearer that this was about to land!

PR #566, opened 2021-07-04
fix(eslint-plugin): [no-output-on-prefix] handle `getters` and `outputs` metadata property

rafaelss95:
This PR implements the checks for both `getters` and `outputs` metadata property for `no-output-on-prefix` rule.

PR #582, opened 2021-07-10
fix(eslint-plugin): [no-input-prefix] handle alias and `inputs` metadata property

rafaelss95:
Following the same we did in #566, this PR implements the checks for both alias and `inputs` metadata property for `no-input-prefix` rule.


JamesHenry (approved):
Thanks @rafaelss95!

PR #584, opened 2021-07-10
refactor(eslint-plugin): [no-inputs-metadata-property] switch to selectors

rafaelss95:
Basically changes the implementation to use selectors. In addition to this, I added some tests to ensure we only report non-computed properties.


JamesHenry (changes requested):
Really sorry for causing the conflict with my PR to better support Angular CLI 12.1

JamesHenry (approved):

PR #594, opened 2021-07-12
feat(eslint-plugin): [use-injectable-provided-in] add suggestion

### LINKED ISSUE #236: use-injectable-provided-in should be configurable
Currently, when we enable this rule, we got a lot of fails with `Interceptors` and `Effects` (from Ngrx).

Correct me if I'm wrong, but afaik we can't use `providedIn` for these. With this assumption in mind I'd want to suggest something for this rule:

- For the `Interceptors` case, I guess it must be disabled by default as it's an Angular native feature;
- For the `Effects`, I guess a configuration pattern like this: `{ ignore: /Effects$/ }` would attend this and other possible corner cases.

rafaelss95:
**1st. commit**: adds suggestions that covers the most common values and also fixes some false negatives like when we use `null` as value or when we use a computed argument to `@Injectable` decorator and false positives when we use `'providedIn'` as a `Literal`;
**2nd. commit**: ignore classes that implements `HttpInterceptor` (see #236);
**3rd. commit**: add option to ignore classes given a pattern (see #236).

Closes #236.

### DIFF
--- a/packages/eslint-plugin/src/rules/use-injectable-provided-in.ts
+++ b/packages/eslint-plugin/src/rules/use-injectable-provided-in.ts
@@ -1,13 +1,13 @@
 import type { TSESTree } from '@typescript-eslint/experimental-utils';
 import { createESLintRule } from '../utils/create-eslint-rule';
-import { INJECTABLE_CLASS_DECORATOR } from '../utils/selectors';
 import {
-  AngularClassDecorators,
-  getDecoratorPropertyValue,
-} from '../utils/utils';
+  INJECTABLE_CLASS_DECORATOR,
+  metadataProperty,
+} from '../utils/selectors';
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/rules/use-injectable-provided-in.ts
```diff
@@ -1,41 +1,69 @@
...
       category: 'Best Practices',
       recommended: false,
+      suggestion: true,
     },
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          ignoreClassNamePattern: {
```
rafaelss95, line 24: Not really sure about this name.

JamesHenry:
Thanks @rafaelss95!

PR #601, opened 2021-07-18
fix(eslint-plugin-template): [no-call-expression]: `FunctionCall`s not being reported

rafaelss95:
This PR aims to correct report `FunctionCall`s in `no-call-expression`, e.g.:

```html
<div>{{ getInfo()() }}</div>
<a [href]="test()()">info</a>
```


JamesHenry (changes requested):
This needs a docs regen:

```sh
yarn update-rule-docs
```

PR #606, opened 2021-07-21
fix(eslint-plugin): [no-empty-lifecycle-method] incorrect suggestions and correct reports

### LINKED ISSUE #604: [no-empty-lifecycle-method] also removes everything after `extends` keyword

```JSON
{
  "rules": {
    "@angular-eslint/no-empty-lifecycle-method": ["error"]
  }
}
```

```TS
// your repro code case
abstract class A<T> {
    protected foo(bar: T): { return bar }
}

class B extends A<never> implements onInit() {
    ngOnInit() {} // here comes a reasonable error;
} 
```

Then make **Ctrl + .** to remove an empty lifecycle hook then it also removes `extends A<never>`, which is not needed.

**Versions**

| `@angular-eslint/eslint-plugin`    | `12.2.0` |
| `@typescript-eslint/parser`        | `12.2.0` |
| `ESLint`                           | `7.30.0` |
| `node`                             | `16.1.0` |

```sh


                |___/
    

Angular CLI: 12.1.1
Node: 16.1.0 (Unsupported)
Package Manager: npm 7.12.1
OS: win32 x64

Angular: 12.1.1
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, material, material-moment-adapter

@angular-devkit/architect       0.1201.1
@angular-devkit/build-angular   12.1.1
@angular-devkit/core            12.1.1
@angular-devkit/schematics      12.1.1
@angular/fire                   6.1.5
@angular/flex-layout            12.0.0-beta.34
@schematics/angular             12.1.1
rxjs                            7.2.0
    
```

- [ yes ] I have tried restarting my IDE and the issue persists.
- [ yes ] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

rafaelss95:
In addittion to the fix on suggestion, it now reports literal `MethodDefinition`s and removes interfaces within `MemberExpression`s.

Fixes #604.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts
+++ b/packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts
@@ -58,6 +58,12 @@ export default createESLintRule<Options, MessageIds>({
                 const importDeclarations =
                   getImportDeclarations(node, '@angular/core') ?? [];
                 const interfaceName = getRawText(node).replace(/^ng+/, '');
+                const text = sourceCode.getText();
+                const totalInterfaceOccurrences = getTotalInterfaceOccurrences(
+                  text,
+                  interfaceName,
+                );
+                const totalInterfaceOccurrencesSafeForRemoval = 2;
 
                 return [
                   fixer.remove(node),
@@ -67,12 +73,15 @@ export default createESLintRule<Options, MessageIds>({
                     interfaceName,
                     fixer,
                   ),
-                  getImportRemoveFix(
-                    sourceCode,
-                    importDeclarations,
-                    interfaceName,
-                    fixer,
-                  ),
+                  totalInterfaceOccurrences <=
+                  totalInterfaceOccurrencesSafeForRemoval
+                    ? getImportRemoveFix(
+                        sourceCode,
+                        importDeclarations,
+                        interfaceName,
+                        fixer,
+                      )
+                    : null,
                 ].filter(isNotNullOrUndefined);
               },
             },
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts
```diff
@@ -67,12 +63,6 @@ export default createESLintRule<Options, MessageIds>({
                     interfaceName,
                     fixer,
                   ),
-                  getImportRemoveFix(
```
rafaelss95, line 70: I'm proposing to remove it because we can't guarantee that the import is **only** being used for this class, in other words it might be being used in another(s) class(es) also in the same file and so we could probably be suggesting a code that does not compile. Does that make sense for you, @JamesHenry?

JamesHenry: Can we do a simple string-based check on the file contents for that?

If there is more than one occurrence of the string, we leave it (yes, this _could_ have false positives, e.g. from comments, but still better than never removing it for the common case), if there is only one occurrence (the import) then we remove it?

WDYT?

JamesHenry (changes requested):
See my comment response and there are conflicts from one of your earlier PRs being merged

packages/eslint-plugin/src/rules/no-empty-lifecycle-method.ts
rafaelss95: Indeed! I've submitted this check, PTAL!

JamesHenry (approved):

PR #507, opened 2021-05-25
docs(eslint-plugin-template): [no-positive-tabindex] markdown documentation

rafaelss95:

PR #610, opened 2021-07-30
docs: created docs for rule 'component-max-inline-declarations'

jfgreffier:
Docs for rule component-max-inline-declarations

Helps with https://github.com/angular-eslint/angular-eslint/issues/197


jfgreffier:
@JamesHenry is working on a solution to generate doc automatically https://github.com/angular-eslint/angular-eslint/pull/507#issuecomment-848048082

JamesHenry:
Thanks for your patience on this @jfgreffier - I have just merged the first version of the automated docs generation logic. We will continue to improve on it from here, but for now it addresses this PR.

Thanks again!

PR #615, opened 2021-08-02
chore(eslint-plugin-template): autogenerate rule docs

JamesHenry:
I have some further improvements planned for this, including a website which will take it all to the next level.

The other plugin will come in a follow up PR.


rafaelss95:
Awesome job, @JamesHenry!! 🚀

packages/eslint-plugin-template/src/rules/cyclomatic-complexity.ts
```diff
@@ -27,6 +29,9 @@ export default createESLintRule<Options, MessageIds>({
           maxComplexity: {
             type: 'number',
             minimum: 1,
+            // Used by docs generator, edit with care
+            description: `
+Default is: \`${DEFAULT_MAX_COMPLEXITY}\``.trim(),
```
rafaelss95, line 34: A small question/suggestion:

What if we use `default` (here and in some others) instead?

```suggestion
            default: DEFAULT_OPTIONS.maxComplexity, // `DEFAULT_OPTIONS` object is what we have used in most of the cases.
```

We could use `description` only if there's a needed explanation about the option as you did in `no-duplicate-attributes`.

Note that it also helps with text standardarizion (in some cases we have 'Default is...' and in others 'Default value...'), so we can choose one and put it fixed in the doc-gen:

`Default value: ${ruleConfig.default}`

JamesHenry: That isn't supported by https://github.com/bcherny/json-schema-to-typescript

JamesHenry: @rafaelss95 We can consider forking/extending it in the future, but going with this for now

PR #622, opened 2021-08-05
docs: autogenerate default values as part of options description

JamesHenry:
FYI @rafaelss95 

Now that things are in pretty good shape, this also updates the plugins to point at the docs for individual rules


rafaelss95:
That's awesome! 🚀

PR #665, opened 2021-09-05
fix(eslint-plugin-template): [i18n] fixes some incorrect reports

### LINKED ISSUE #412: [i18n] Provide a way to exclude non-bounded texts
As I commented in https://github.com/angular-eslint/angular-eslint/issues/298#issuecomment-808909754:

We need either create a new option called `ignoreTexts` to check only non-bounded texts or create this option to check all types of texts and thus deprecating the `boundTextAllowedPattern`.

IMHO we should go with the second as I don't see why one would want to differentiate these two kinds of texts, but I'm all ears.

<hr>

Additional info/context:

From #298:

> 1) It would be really useful to be able to skip some simple HTML items that in most cases shouldn't be translated:

```html
<span>&nbsp;</span>
<span>$</span>
```

> I tried to filter out those texts using `boundTextAllowedPattern` but without success and end up rewriting such items as follows:

```html
<ng-container>&nbsp;</ng-container>
<ng-container>$</ng-container>
```

> and setting `ignoreTags: ['ng-container']`. It's not a good solution, though.

> I wonder whether it's possible to achieve that with `boundTextAllowedPattern` or there should be another new option like `ignoreTexts`?

<hr>

From #329:

> Currently there is a pattern ("boundTextAllowedPattern" option) to filter out non-textual content. But it works only for bound text (for example: `<span>+{{data_from_backend}}</span>`). 
https://github.com/angular-eslint/angular-eslint/blob/eeb49d491bb92e470f505a699ef3e12963ff351b/packages/eslint-plugin-template/src/rules/i18n.ts#L252

> But it is not applied to text nodes (for example: `<span>+</span>`).
https://github.com/angular-eslint/angular-eslint/blob/eeb49d491bb92e470f505a699ef3e12963ff351b/packages/eslint-plugin-template/src/rules/i18n.ts#L249

> So there is currently no way to filter out this content. There are currently two options:
> 1. Apply "checkBoundTextAllowedPattern" to the text nodes.
> 2. Add new option ("checkTextAllowedPattern") and apply it to the text nodes.

Also https://github.com/angular-eslint/angular-eslint/issues/400#issuecomment-819950602

<hr>

**Versions**

| `@typescript-eslint       `        | `4.20.0` |
| `TypeScript`                       | `4.1.5` |
| `ESLint`                           | `7.23.0` |
| `node`                             | `14.16.0` |

```sh
Angular CLI: 11.2.6
Node: 14.16.0
OS: darwin x64

Angular: 11.2.8
... language-service, localize, platform-browser
... platform-browser-dynamic, platform-server, router
... service-worker
Ivy Workspace: Yes

@angular-devkit/architect       0.1102.4
@angular-devkit/build-angular   0.1102.6
@angular-devkit/core            11.2.4
@angular-devkit/schematics      11.2.6
@angular/cdk                    11.2.7
@angular/cli                    11.2.6
@angular/material               11.2.7
@schematics/angular             11.2.6
@schematics/update              0.1102.6
rxjs                            6.6.7
typescript                      4.1.5
webpack                         4.46.0
```

### LINKED ISSUE #513: [i18n] Unexpected behaviors

1. `boundTextAllowedPattern` isn't behaving as expected. See this case:
https://github.com/angular-eslint/angular-eslint/blob/b434b513abee0b3bee4db64c8fecfd866324634b/packages/eslint-plugin-template/tests/rules/i18n.test.ts#L204-L215

Here we report a bound text that starts with an allowed character. It should be valid instead;

2. Missing i18n-* attributes are reported duplicately. See this case:

https://github.com/angular-eslint/angular-eslint/blob/b434b513abee0b3bee4db64c8fecfd866324634b/packages/eslint-plugin-template/tests/rules/i18n.test.ts#L167-L178

3. Report ranges could be more assertive. For example, when an i18n attribute is missing an id or the text is missing the i18n attribute, the whole element is reported, when we just need to report the original attribute or text;

4. `<ng-template>` , `ngPlural`, `ngPluralCase` and `ngProjectAs` should be ignored;

5. Numeric values and HTML entities (https://github.com/mgechev/codelyzer/issues/618 and https://github.com/mgechev/codelyzer/issues/952) should be ignored. Examples:

```html
<div ariaselected="0"></div>
<div matBadge="&#8288;">5</div>
<div>732</div>
<div>&nbsp;</div>
```

6. ICUs are not being reported (https://angular.io/guide/i18n#translate-plurals-and-alternate-expressions);

7. Duplicate custom IDs are not being reported (https://angular.io/guide/i18n#define-unique-custom-ids).

rafaelss95:
This PR does the following:

- add `ngProjectAs` directive to the `ignoredAttributes` by default;
- ignores attributes with numeric values such as "0", "1", "2" by default;
- ignore non-texts by default;
- makes the reporting range more assertive, that is, only the attribute or text is reported and no longer the entire element as it currently is;
- visit each selector independently, instead of for-loops and conditionals;
- changes the `messageIds` in hope to better express the intent;
- more tests;
- removes the `Template` visitor  as it does nothing but duplicate error reports, as you can see here for example: https://github.com/angular-eslint/angular-eslint/blob/b434b513abee0b3bee4db64c8fecfd866324634b/packages/eslint-plugin-template/tests/rules/i18n.test.ts#L167-L178
- fixes the `boundTextAllowedPattern` behavior: now if you pass a config like `{ boundTextAllowedPattern: '-' }`, it doesn't report a bound text containing this symbol like it currently does:
https://github.com/angular-eslint/angular-eslint/blob/b434b513abee0b3bee4db64c8fecfd866324634b/packages/eslint-plugin-template/tests/rules/i18n.test.ts#L204-L215

Fixes #412, fixes #513.


rafaelss95:
I've seen this error happens in CI a dozen of times. Is it NX-related...more specifically **https://github.com/nrwl/nx/issues/6370**, right?

No such file or directory error

JamesHenry (approved):

JamesHenry:
Hmm yeah not 100% sure, I've seen it a few times myself but it's always resolved immediately upon rerunning CI.

Are you able to trigger CI reruns?

PR #676, opened 2021-09-09
feat(utils): publicly expose utils related to eslint-plugin

JamesHenry:
FYI @rafaelss95 this is a significant internal refactor. The utils package going beyond testing utils happened sooner than I thought 😁 

I'm fortunate to be dogfooding this at work right now writing some workspace specific lint rules using these utils.

We obviously have v13 coming up too so we can always make some breaking changes to these groupings and import paths if we need to.

I will be following up with a similar PR for the template related utils


rafaelss95:
Nice! I'll try to take a look at this later today or tomorrow to see what could be improved (mainly making things more flexible since they were created for this specific project) to take advantage of the v13 migration window.

Issue #675, opened 2021-09-09
[sort-ngmodule-metadata-arrays] `--fix` not able to correct all the reports once


splincode:
```JSON
{
  "rules": {
    "@angular-eslint/sort-ngmodule-metadata-arrays": "error"
  }
}
```

After run with `--fix` didn't work for my cases:

```ts
// ...

@NgModule({
    imports: [
        CommonModule,
        DeepPathPipeModule,
        DefaultValuePipeModule,
        DragDropModule,
        MatIconModule,
        MergeCssClassesPipeModule,
        SafePipeModule
    ],
    declarations: [
        AutoHeightDirective,
        NgxColumnComponent,
        NgxOptionsComponent,
        TableBuilderComponent,
        TableTbodyComponent,
        TableTheadComponent,
        TableCellComponent,
        TemplateBodyTdDirective,
        TemplateHeadThDirective,
        ObserverViewDirective,
        NgxContextMenuComponent,
        NgxContextMenuItemComponent,
        NgxContextMenuDividerComponent,
        NgxMenuContentComponent,
        NgxEmptyComponent,
        NgxHeaderComponent,
        NgxFooterComponent,
        NgxFilterViewerComponent,
        NgxFilterComponent,
        NgxFilterDirective,
        DragIconComponent,
        NgxSourceNullComponent,
        DisableRowPipe,
        TableSelectedItemsPipe,
        MapToTableEntriesPipe,
        VirtualForDirective,
        GetFreeSizePipe,
        GetClientHeightPipe
    ],
    providers: [WebWorkerThreadService],
    exports: [
        NgxColumnComponent,
        NgxOptionsComponent,
        NgxContextMenuComponent,
        TableBuilderComponent,
        TemplateBodyTdDirective,
        TemplateHeadThDirective,
        NgxContextMenuItemComponent,
        NgxContextMenuDividerComponent,
        NgxMenuContentComponent,
        NgxEmptyComponent,
        NgxHeaderComponent,
        NgxFooterComponent,
        NgxFilterViewerComponent,
        NgxFilterComponent,
        NgxFilterDirective,
        TableSelectedItemsPipe,
        MapToTableEntriesPipe,
        NgxSourceNullComponent
    ]
})
// ...
```

**Versions**

| package                                   | version |
| ----------------------------------        | ------- |
| `@angular-eslint/eslint-plugin`  | 12.3.1 |
| `@typescript-eslint/parser`               | 4.31.0 |
[...]

PR #677, opened 2021-09-10
refactor(eslint-plugin): [sort-ngmodule-metadata-arrays]: properly handle multiple fixes at once

rafaelss95:
Besides addressing the issue #675, the report node is more accurate now and reports nested arrays.

PS: TBH, I'm not sure which scope should be this PR, so I used `refactor` 🤷‍♂️

Closes #677.

### DIFF
--- a/packages/eslint-plugin/src/rules/sort-ngmodule-metadata-arrays.ts
+++ b/packages/eslint-plugin/src/rules/sort-ngmodule-metadata-arrays.ts
@@ -1,10 +1,10 @@
-import { Selectors } from '@angular-eslint/utils';
-import type { TSESTree } from '@typescript-eslint/experimental-utils';
+import { ASTUtils, partition, Selectors } from '@angular-eslint/utils';
+import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 import { ASTUtils as TSESLintASTUtils } from '@typescript-eslint/experimental-utils';
 import { createESLintRule } from '../utils/create-eslint-rule';
 
 type Options = [];
-export type MessageIds = 'sortNgmoduleMetadataArrays';
+export type MessageIds = 'sortNgmoduleMetadataArrays' | 'suggestFix';
 export const RULE_NAME = 'sort-ngmodule-metadata-arrays';
 
 export default createESLintRule<Options, MessageIds>({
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin/src/rules/sort-ngmodule-metadata-arrays.ts
```diff
@@ -16,42 +16,124 @@ export default createESLintRule<Options, MessageIds>({
...
       category: 'Best Practices',
       recommended: false,
+      suggestion: true,
     },
     fixable: 'code',
     schema: [],
     messages: {
       sortNgmoduleMetadataArrays:
         '`NgModule` metadata arrays should be sorted in ASC alphabetical order',
+      suggestFix: 'Sort members (WARNING! comments will probably be messed up)',
```
rafaelss95, line 26: Please, suggest a better message if that's the case.

packages/eslint-plugin/src/rules/sort-ngmodule-metadata-arrays.ts
```diff
@@ -16,42 +16,124 @@ export default createESLintRule<Options, MessageIds>({
...
-          node: unorderedNode,
+          node: firstUnorderedNode,
           messageId: 'sortNgmoduleMetadataArrays',
-          fix: (fixer) => [
-            fixer.replaceText(unorderedNode, nextNode.name),
-            fixer.replaceText(nextNode, unorderedNode.name),
-          ],
+          ...(hasCommentsWithinElements(sourceCode, elements)
+            ? { suggest: [{ messageId: 'suggestFix', fix }] }
+            : { fix }),
```
rafaelss95, line 56: I even tried to use `fix`, even if there were comments, but for god's sake, besides being very costly to maintain such an implementation, it's also very prone to failure.

packages/utils/src/utils.ts
```diff
@@ -65,3 +65,32 @@ export function capitalize<T extends string>(text: T): Capitalize<T> {
 export function withoutBracketsAndWhitespaces(text: string): string {
   return text.replace(/[[\]\s]/g, '');
 }
+
+/**
+ * Split the `array` into two `arrays`, where the first array contains all
```
rafaelss95, line 70:
```suggestion
 * Splits the `array` into two `arrays`, where the first array contains all
```

JamesHenry:

PR #679, opened 2021-09-11
refactor(eslint-plugin-template): [click-events-have-key-events] switch to selectors

rafaelss95:
**1st. commit**: replaces long if blocks in `eslint-plugin-template/utils` with early return to improve readability;
**2nd. commit**: corrects a small mistake where we check on a `NgStyle` if a node is hidden from screenreader;
**3rd. commit**:
- switch implementation to be selector-based;
- improve docs to better describe the intention;
- update tests descriptions and, even if unintentionally, corrects the docs gen of valid cases 😆


JamesHenry:

PR #681, opened 2021-09-13
fix(eslint-plugin-template): [accessibility-alt-text] correct false positives

rafaelss95:
This PR fixes various issues respecting the specs. It also adds suggestions where applicable.


JamesHenry:

PR #683, opened 2021-09-17
fix(eslint-plugin-template): [i18n] ignore empty strings and non-texts within `BoundText` by default

### LINKED ISSUE #682: [template/i18n] reports error for interpolation

```JSON
{
  "rules": {
        "@angular-eslint/template/i18n": [
          "error",
          {
            "checkId": true,
            "checkAttributes": false,
            "ignoreTags": [":svg:title", ":svg:desc", ":svg:tspan"]
          }
        ],
  }
}
```

Interpolation bindings get reported by the rule with `Each element containing text node should have an i18n attribute.` but this is no actual english text in there to extract. This was working ok with 12.3.x so I guess it was caused by https://github.com/angular-eslint/angular-eslint/pull/665
```HTML
<ng-template #errorMessage>
  {{ error.title }}
</ng-template>
```

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `12.4.1` |
| `@angular-eslint/template-parser`         | `12.4.1` |
| `@typescript-eslint/parser`               | `4.28.5` |
| `ESLint`                                  | `7.29.0` |
| `node`                                    | `14.17.0` |

```sh
Angular CLI: 12.2.6
Node: 14.17.0
Package Manager: npm 6.14.13
OS: darwin x64

Angular: 12.2.6
... animations, cli, common, compiler, compiler-cli, core
... elements, forms, language-service, localize
... service-worker

@angular-devkit/architect       0.1202.1
@angular-devkit/build-angular   12.2.6
@angular-devkit/core            12.1.4
@angular-devkit/schematics      12.2.6
@angular/cdk                    12.2.1
@angular/google-maps            12.2.1
@schematics/angular             12.2.6
ng-packagr                      12.2.1
rxjs                            7.3.0
```

rafaelss95:
Fixes #682.

FWIW I missed that because it was named incorrectly since the creation of this rule [here](https://github.com/angular-eslint/angular-eslint/commit/0fd5b25204a09f5288353a425f4ecc186694b5ca) where it says the variable name is `DEFAULT_BOUND_TEXT_ALLOWED_PATTERN`, when it was actually referring to the **non-allowed** pattern, that is, letters from "A" to "Z", and that's why I've confused myself all this time with this 😞.


packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -17,9 +17,9 @@ import {
...
-const DEFAULT_BOUND_TEXT_ALLOWED_PATTERN = /[a-z]/i;
+const DEFAULT_ALLOWED_BOUND_TEXT_PATTERN = RegExp(`[^${PL_PATTERN}]`);
```
rafaelss95, line 22: It should be correct now. The **allowed** text is everything that is **not** in the alphabet and not the other way around.

rafaelss95: It should be correct now. The **allowed** text is everything that is **not** in the alphabet (+ latins characters — as expressed by `PL_PATTERN`) and not the other way around.

packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -457,12 +436,13 @@ function isAttributeAllowed(
...
+  allowedBoundTextPattern: RegExp,
   {
     value: {
       ast: { strings },
     },
   }: StronglyTypedBoundText,
 ) {
-  return boundTextAllowedPattern.test(strings.join('').trim());
+  const text = strings.join('').trim();
+  return !PL_PATTERN.test(text) || allowedBoundTextPattern.test(text);
```
rafaelss95, line 447: We only consider it as an allowed text if it doesn't contain letters or if it was deliberately allowed by the user via `boundTextAllowedPattern`.

packages/eslint-plugin-template/tests/rules/i18n/cases.ts
```diff
@@ -181,15 +190,15 @@ export const invalid = [
       'should fail if `i18n` attribute is missing on element containing bound text',
     annotatedSource: `
       <div>
-        <span>-{{data_from_backend}}</span>
```
rafaelss95, line 184: I've moved this test to the valid section, as it should be valid since there's no letter here.

packages/eslint-plugin-template/tests/rules/i18n/cases.ts
```diff
@@ -46,13 +46,16 @@ export const valid = [
...
   },
+  `<div>-{{data_from_backend}}</div>`,
+  `<div>1{{data_from_backend}}</div>`,
+  `<div>-1{{data_from_backend}}</div>`,
   {
     code: `
-      <div i18n-tooltip="@@tooltip.label" tooltip="This tooltip property is ignored">
-        -{{data_from_backend}}
+      <div>
+        My company untranslatable name{{data_from_backend}}
```
rafaelss95, line 55: Thinking more about it, I think we still should address #412, because one may still needs to exclude some texts from `Text` (not only from `BoundText`).

packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -177,7 +177,7 @@ export default createESLintRule<Options, MessageIds>({
     const parserServices = getTemplateParserServices(context);
     const sourceCode = context.getSourceCode();
     const allowedBoundTextPattern = RegExp(
-      boundTextAllowedPattern ?? DEFAULT_BOUND_TEXT_ALLOWED_PATTERN,
+      boundTextAllowedPattern ?? DEFAULT_ALLOWED_BOUND_TEXT_PATTERN,
```
JamesHenry, line 180: Please can we avoid passing an existing instance of RegExp to the RegExp constructor here? It is only needed for the string coming from configuration.

Additionally please can you add the `new` keyword to the RegExp constructor call, it works without it but I think it is preferable to have it there.

packages/eslint-plugin-template/src/rules/i18n.ts
JamesHenry: Please can you apply this as a comment to the code too? I think it is great to capture your intention there

packages/eslint-plugin-template/tests/rules/i18n/cases.ts
JamesHenry: Sounds good, thanks! I'll create a v13 branch soon as well

JamesHenry (changes requested):

packages/eslint-plugin-template/src/rules/i18n.ts
rafaelss95: > Please can we avoid passing an existing instance of RegExp to the RegExp constructor here? It is only needed for the string coming from configuration.

Oh yeah, for sure, I missed that!

> Additionally please can you add the new keyword to the RegExp constructor call, it works without it because of legacy reasons but I think it is preferable to have it there.

_"because of legacy reasons"_ — Is there any reference where I can find this? Also, is it something related to `RegExp(...)` specifically or, for example, `Error(...)` too? To clarify, I'm not doubting or anything like that, just asking because I've never seen that we shouldn't use it like this because it's legacy and somewhat will be deprecated. I thought it just to work like the the other factories `Boolean(...)`, `String(...)`or even `Error(...)` instead of calling the classes directly `new Boolean(...)`, `new String(...)`, `new Error(...)`.

That being said, of course I can change this without any problem.

JamesHenry: Sorry for not being clearer. What I mean by that is:

Constructors functions such as `RegExp` and `Error` come from a time before the ECMAScript class feature was introduced. They are _functionally_ classes and were they introduced as globals today, they would literally be classes, however classes have different semantics to straight functions so they can never be converted - that would break the web.

As I'm sure you know, when you use a class, you cannot use its constructor without the `new` keyword, you will get a `TypeError`. Were it not for the legacy of the web, these globals would therefore also throw a `TypeError` without the use of the `new` keyword, that's what I meant by my comment above.

Completely anecdotal, but I _feel_ like I see far more cases of `RegExp` and `Error` used with the `new` keyword than without (the same is true for `Array`). Interestingly I feel the opposite about those representing non-object primitive values - `String`, `Boolean`, `Number`. I feel like I never/almost never see those use with `new`.

A fine old mess :D

rafaelss95: Thanks, it's clear now.

is really interesting, taking a look at the `@angular/components` codebase, for example, I find a lot more `Error(...)` instead of `new Error(...)`, while a lot more `new RegExp(...)` than `RegExp(...)` 😅

PR #684, opened 2021-09-17
refactor(eslint-plugin-template): [i18n] remove non-actionable suggestion

rafaelss95:
Currently, the i18n rule has a suggestion that doesn't do anything. It was introduced in https://github.com/angular-eslint/angular-eslint/commit/25234c80b5bca3700d3ca71ab9181a8dc1ff1360 and it basically acts like a hint telling the user to add some config to the `.eslint` file, but IMHO it's unnecessary. We don't do this for any other option in this or other rule. Why should this be treated differently?


JamesHenry:
Yes, fair enough. Maybe in the future we can link to our own rule docs from some of these types of reports which would then include tips on how to address it, we can think more about that...

PR #693, opened 2021-09-20
fix(eslint-plugin): [sort-ngmodule-metadata-arrays] report the correct node

rafaelss95:
We currently report the node that precedes the violation. With this PR, it should be no longer a problem.


JamesHenry (approved):
Awesome, thanks!

PR #694, opened 2021-09-20
fix(eslint-plugin): [sort-ngmodule-metadata-arrays] remove the property restriction

rafaelss95:
Commit 1 is from #693 and that PR should be merged first.

---

We currently have a restriction to search only some properties, but we don't need to. The rule should report all unordered arrays within `NgModule`, whatever the property name is.


JamesHenry (approved):
Great, super clear thank you!

PR #697, opened 2021-09-20
build: update TS to 4.4.3

rafaelss95:
This updates TS to [4.4.3](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/) so we can use new features 🔥.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,6 @@
     "terser": "^4.0.0",
     "ts-jest": "27.0.3",
     "ts-node": "^9.1.1",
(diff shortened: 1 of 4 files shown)


JamesHenry:
@rafaelss95 We usually choose the TypeScript version based on the version that the currently supported major version of Angular is supporting.

It looks like TS 4.4 has been a tricky one to land for Angular: https://github.com/angular/angular/pull/43281 and it is still not in yet so I think we should hold off for now

PR #698, opened 2021-09-20
feat(eslint-plugin-template): [no-restricted-syntax] add rule

### LINKED ISSUE #655: Rule proposal: `template/no-restricted-syntax`
Similar to https://eslint.org/docs/rules/no-restricted-syntax, but for templates. Taking https://github.com/angular-eslint/angular-eslint/issues/308 as an example:

```json
{
  "rules": {
    "@angular-eslint/template/no-restricted-syntax": [
      "error",
      {
        "selector": "Element[name='button']:not(:has(:matches(BoundAttribute, TextAttribute)[name='type']))",
        "message": "Missing an explicit type attribute for button"
      },
      {
        "selector": "Element[name='button'] :matches(BoundAttribute, TextAttribute)[name='type'][value!=/^(button|reset|submit)$/]",
        "message": "Invalid value for button type attribute"
      },
    ]
  }
}
```

rafaelss95:
Closes #655.

<s>Note that the first two commits are from #697.<s>


packages/eslint-plugin-template/docs/rules/no-restricted-syntax.md
```diff
@@ -0,0 +1,63 @@
...
+Disallows specified syntax
+
+- Type: suggestion
+- Category: Stylistic Issues
+
+<br>
+
+## Rule Options
+
+The rule does not have any configuration options.
```
rafaelss95, line 25: Unfortunately, [there's a bug on `json-schema-to-typescript` so it doesn't compile schema well if it contains `oneOf`](https://github.com/bcherny/json-schema-to-typescript/issues/381).

packages/eslint-plugin-template/docs/rules/no-restricted-syntax.md
```diff
@@ -0,0 +1,63 @@
...
+        ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+```
+
+<br>
+
+---
+
+<br>
+
+✅ - Examples of **correct** code for this rule:
```
rafaelss95, line 63: Hmmm .. the generator only grab tests that don't contain options, right? What can we do for this case here since the rule has no defaults + the options are not optional?

packages/eslint-plugin-template/docs/rules/no-restricted-syntax.md
```diff
@@ -0,0 +1,63 @@
...
+
+<br>
+
+## Usage Examples
+
+> The following examples are generated automatically from the actual unit tests within the plugin, so you can be assured that their behavior is accurate based on the current commit.
+
+<br>
+
+❌ - Examples of **incorrect** code for this rule:
```
rafaelss95, line 35: For these examples, it would be great to have the options defined somewhere (perhaps in the form of a comment).

packages/eslint-plugin-template/src/utils/get-loc-from-absolute-source-span.ts
```diff
@@ -0,0 +1,12 @@
+import type { AbsoluteSourceSpan } from '@angular/compiler';
+import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
+
+export function getLocFromAbsoluteSourceSpan(
```
rafaelss95, line 4: Not sure if it's the best name + the best place, but I do think it's a good ideia to have this utility (we can refactor the rules to use it in another PR).

packages/eslint-plugin-template/src/rules/no-restricted-syntax.ts
```diff
@@ -0,0 +1,96 @@
...
+      };
+    }
+
+    return options.reduce<Record<string, TSESLint.RuleFunction>>(
+      (selectors, selectorOrOption) => {
+        const { message, selector } = getNormalizedOptions(selectorOrOption);
+        return {
+          ...selectors,
+          // If the `selector` is already computed (it can happens in case of passing `string` + object format), we just ignore it.
+          ...(!(selector in selectors) && { [selector]: visitor(message) }),
```
rafaelss95, line 78: To be clear, it's preventing the following case:

```ts
{
  "@angular-eslint/template/no-restricted-syntax": [
    "error",
    "BoundAttribute",
    {
      "selector": "BoundAttribute" // note that is a duplicated.
    },
  ]  
}
```

AFAIK, we can't prevent this using `schema`.

---

That being said:

1. Not sure if this comment is clear enough, let me know in case of rephrase needed.
2. Or perhaps would it be better to give a warning?

JamesHenry:
Sorry, this one will need to be updated as I don't think it's best to move ahead with the TS 4.4 one yet

packages/eslint-plugin-template/docs/rules/no-restricted-syntax.md
JamesHenry: Working on it now

packages/eslint-plugin-template/docs/rules/no-restricted-syntax.md
JamesHenry: Working on it now

JamesHenry:
Config/options in docs is now supported as of: https://github.com/angular-eslint/angular-eslint/pull/699

miluoshi:
Hi, is this waiting for [this](https://github.com/bcherny/json-schema-to-typescript/issues/381) to be fixed?

packages/eslint-plugin-template/src/rules/no-restricted-syntax.ts
JamesHenry: This seems to be the opposite behavior to the existing core rule for JS: https://github.com/eslint/eslint/blob/d36f12f71b3e4f9e9552f1054d7a75be4dc03671/lib/rules/no-restricted-syntax.js#L58

The latest definition for that selector will replace previous ones via the Object.assign.

I think it would be better to match the existing rule here, regardless of whether we agree with it or not

JamesHenry (changes requested):
This is excellent, thanks so much @rafaelss95! Just the one point to deal with around multiple selector definitions.

I'm really sorry it took me so long to get this reviewed, but thank you again for all your hard work 🙏

JamesHenry:
I emailed Rafael about 2 months ago asking if wished to continue working on this but sadly have not received a response in that time. I will therefore close this so that it is clear that it is not being worked on and that somebody else can pick this up if it is important to them.

Many thanks to Rafael for his fantastic contributions to angular-eslint to date 🙏

PR #702, opened 2021-09-22
fix(eslint-plugin-template): [i18n] handle `ng-template` properly

### LINKED ISSUE #701: [template/i18n] reports error for ng-template's marked with i18n tag

```JSON
{
  "rules": {
        "@angular-eslint/template/i18n": [
          "error",
          {
            "checkId": true,
            "checkAttributes": false,
            "ignoreTags": [":svg:title", ":svg:desc", ":svg:tspan"]
          }
        ],
  }
}
```

i18n tags on `ng-template` tags get marked as missing i18n e.g. both of these cases are reported as errors (via the latest canary version of angular-eslint):
```HTML
      <ng-template
        [ngIf]="userGroups?.length > 0"
        [ngIfElse]="noTeamsTmpl"
        i18n="@@cu-user-list__no-matching-users-groups"
      >
        No people or teams matched your search
      </ng-template>
      <ng-template i18n="@@cu-user-list__no-matching-users" #noTeamsTmpl>
        No people matched your search
      </ng-template>
```

This error also gets thrown for some libraries (disabling the i18n rule fixes it so it originates somewhere in that rule):

```
Cannot read property 'start' of undefined
```

I presume it's caused by one of these 2 arguments being `undefined`:
https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/rules/i18n.ts#L337
https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/rules/i18n.ts#L348

The stack trace is kind of useless, so I'm struggling to find the exact bit of code causing that error:

```
TypeError: Cannot read property 'start' of undefined
    at Linter._verifyWithoutProcessors (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/linter/linter.js:1194:54)
    at Linter._verifyWithConfigArray (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/linter/linter.js:1276:21)
    at Linter.verify (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/linter/linter.js:1231:25)
    at Linter.verifyAndFix (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/linter/linter.js:1424:29)
    at verifyText (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/cli-engine/cli-engine.js:234:48)
    at CLIEngine.executeOnFiles (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/cli-engine/cli-engine.js:802:28)
    at ESLint.lintFiles (/Users/mattlewis/Code/clickup/frontend/node_modules/eslint/lib/eslint/eslint.js:563:23)
    at Object.<anonymous> (/Users/mattlewis/Code/clickup/frontend/packages/linter/src/executors/eslint/utility/eslint-utils.ts:41:23)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/mattlewis/Code/clickup/frontend/node_modules/tslib/tslib.js:114:62)
```

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `12.5.1-alpha.5` |
| `@angular-eslint/template-parser`         | `12.5.1-alpha.5` |
| `@typescript-eslint/parser`               | `4.28.5` |
| `ESLint`                                  | `7.29.0` |
| `node`                                    | `14.17.0` |

```sh
Angular CLI: 12.2.6
Node: 14.17.0
Package Manager: npm 6.14.13
OS: darwin x64

Angular: 12.2.6
... animations, cli, common, compiler, compiler-cli, core
... elements, forms, language-service, localize
... service-worker

@angular-devkit/architect       0.1202.1
@angular-devkit/build-angular   12.2.6
@angular-devkit/core            12.1.4
@angular-devkit/schematics      12.2.6
@angular/cdk                    12.2.1
@angular/google-maps            12.2.1
@schematics/angular             12.2.6
ng-packagr                      12.2.1
rxjs                            7.3.0
```

rafaelss95:
Fixes #701.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/i18n.ts
+++ b/packages/eslint-plugin-template/src/rules/i18n.ts
@@ -7,8 +7,13 @@ import type {
   TmplAstText,
   TmplAstTextAttribute,
 } from '@angular/compiler';
-import { TmplAstBoundText, TmplAstElement } from '@angular/compiler';
-import type { Message } from '@angular/compiler/src/i18n/i18n_ast';
+import {
+  TmplAstBoundText,
+  TmplAstElement,
+  TmplAstTemplate,
+} from '@angular/compiler';
+import type { I18nMeta, Message } from '@angular/compiler/src/i18n/i18n_ast';
+import type { Node } from '@angular/compiler/src/render3/r3_ast';
 import type { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 import {
   createESLintRule,
@@ -64,13 +69,15 @@ export type MessageIds =
   | 'i18nCustomIdOnElement'
   | 'i18nDuplicateCustomId'
   | 'suggestAddI18nAttribute';
-type StronglyTypedElement = Omit<TmplAstElement, 'i18n'> & {
-  i18n: Message;
+type StronglyTypedI18n<T extends Node> = Omit<T, 'i18n'> & {
+  i18n: I18nMeta;
   parent?: AST;
 };
+type StronglyTypedElement = StronglyTypedI18n<TmplAstElement>;
+type StronglyTypedTemplate = StronglyTypedI18n<TmplAstTemplate>;
 type StronglyTypedTextAttribute = Omit<TmplAstTextAttribute, 'i18n'> & {
-  i18n?: Message;
-  parent: TmplAstElement;
+  i18n?: I18nMeta;
+  parent: TmplAstElement | TmplAstTemplate;
 };
 type StronglyTypedBoundText = TmplAstBoundText & {
   value: ASTWithSource & { ast: Interpolation };
@@ -183,17 +190,22 @@ export default createESLintRule<Options, MessageIds>({
     const allowedTags: ReadonlySet<string> = new Set(ignoreTags);
     const collectedCustomIds = new Map<string, readonly ParseSourceSpan[]>();
 
-    function handleElement({
-      i18n: { customId },
-      name,
-      parent,
-      sourceSpan,
-    }: StronglyTypedElement) {
-      if (allowedTags.has(name) || isElementWithI18n(parent)) {
+    function handleElementOrTemplate(
+      node: StronglyTypedElement | StronglyTypedTemplate,
+    ) {
+      const { i18n, parent, sourceSpan } = node;
+
+      if (
+        isTagAllowed(allowedTags, node) ||
+        isElementWithI18n(parent) ||
+        isTemplateWithI18n(parent)
+      ) {
         return;
       }
 
-      if (!isEmpty(customId)) {
+      const customId = getI18nCustomId(i18n);
+
+      if (customId) {
         const sourceSpans = collectedCustomIds.get(customId) ?? [];
         collectedCustomIds.set(customId, [...sourceSpans, sourceSpan]);
         return;
@@ -210,11 +222,13 @@ export default createESLintRule<Options, MessageIds>({
       i18n,
       keySpan,
       name: attributeName,
-      parent: { name: elementName },
+      parent,
       sourceSpan,
       value,
     }: StronglyTypedTextAttribute) {
-      if (allowedTags.has(elementName)) {
+      const tagName = getTagName(parent);
+
+      if (!tagName || allowedTags.has(tagName)) {
         return;
       }
 
@@ -223,24 +237,24 @@ export default createESLintRule<Options, MessageIds>({
       );
 
       if (checkId && i18n) {
-        const { customId } = i18n;
+        const customId = getI18nCustomId(i18n);
 
-        if (isEmpty(customId)) {
+        if (customId) {
+          const sourceSpans = collectedCustomIds.get(customId) ?? [];
+          collectedCustomIds.set(customId, [...sourceSpans, sourceSpan]);
+        } else {
           context.report({
             messageId: 'i18nCustomIdOnAttribute',
             loc: keyOrSourceSpanLoc,
             data: { attributeName },
           });
-        } else {
-          const sourceSpans = collectedCustomIds.get(customId) ?? [];
-          collectedCustomIds.set(customId, [...sourceSpans, sourceSpan]);
         }
       }
 
       if (
         i18n ||
         !checkAttributes ||
-        isAttributeAllowed(allowedAttributes, elementName, attributeName, value)
+        isAttributeAllowed(allowedAttributes, tagName, attributeName, value)
       ) {
         return;
       }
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -183,17 +191,22 @@ export default createESLintRule<Options, MessageIds>({
...
+        isElementWithI18n(parent) ||
+        isTemplateWithI18n(parent)
+      ) {
         return;
       }
 
-      if (!isEmpty(customId)) {
+      const customId = getI18nCustomId(i18n);
+
+      if (customId) {
```
rafaelss95, line 209: This was one of the causes of having `Cannot read property 'start' of undefined` related in #701 (this message comes from `ESLint` [here](https://github.com/eslint/eslint/blob/0a1a850575ca75db017051abe5e931f0f9c8012b/lib/linter/linter.js#L1236) — as the `Node` doesn't contain a property called `loc` it tries to access `undefined.start` and thus the error doesn't bubble up correctly).

---

Actually `customId` can be `undefined` in some cases, but we were using `Message` when we should use `I18nMeta` instead and check it properly.

packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -306,12 +324,14 @@ export default createESLintRule<Options, MessageIds>({
...
-          handleElement(node);
+        ':matches(Element, Template[tagName="ng-template"])[i18n]'(
+          node: StronglyTypedElement,
+        ) {
+          handleElementOrTemplate(node);
         },
       }),
       ...((checkAttributes || checkId) && {
-        [`Element > TextAttribute[value=${PL_PATTERN}]`](
+        [`:matches(Element, Template[tagName="ng-template"]) > TextAttribute[value=${PL_PATTERN}]`](
```
rafaelss95, line 334: Interestingly the compiler may return `Template` with a different tag name than `ng-template`, when using, for example, an element with structural directive, ex:

```html
<!-- The return here is `Template[tagName="mat-option"]` -->
<mat-option *ngFor="let item of items">{{ item.value }}</mat-option>
```

packages/eslint-plugin-template/tests/rules/i18n/cases.ts
```diff
@@ -173,6 +209,20 @@ export const invalid = [
...
+  //     <ng-template dir="Hello"></ng-template>
+  //                  ~~~
+  //   `,
+  //   messageId: i18nAttribute,
+  //   data: { attributeName: 'dir' },
+  //   annotatedOutput: `
+  //     <ng-template dir="Hello" i18n-dir></ng-template>
+  //                  ~~~
+  //   `,
+  // }),
```
rafaelss95, line 225: I'm having problems with this and other commented tests. I don't know if I'm missing something, but apparently attributes in `ng-template` aren't parsing correctly.

This is the result from `ASTExplorer`:

Template parser with attributes

Note that the `attribute` doesn't come as `TextAttribute` as usual. I've tested with `Angular` and this is the result:

Angular with attributes

packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -183,17 +191,22 @@ export default createESLintRule<Options, MessageIds>({
...
+
+      if (
+        isTagAllowed(allowedTags, node) ||
+        isElementWithI18n(parent) ||
+        isTemplateWithI18n(parent)
+      ) {
         return;
       }
 
-      if (!isEmpty(customId)) {
```
rafaelss95, line 196: This was one of the causes of having `Cannot read property 'start' of undefined` related in #701 (this message comes from `ESLint` [here](https://github.com/eslint/eslint/blob/0a1a850575ca75db017051abe5e931f0f9c8012b/lib/linter/linter.js#L1236) — as the `Node` doesn't contain a property called `loc` it tries to access `undefined.start` and thus the error doesn't bubble up correctly).

---

Actually `customId` can be `undefined` in some cases, but we were using `Message` when we should use `I18nMeta` instead and check it properly.

packages/eslint-plugin-template/tests/rules/i18n/cases.ts
JamesHenry: @rafaelss95 I looked in to this just now and I think it was just a case of missing visitor keys. I have addressed it here: https://github.com/angular-eslint/angular-eslint/pull/968

JamesHenry:
@rafaelss95 I think it might be revisiting this whole PR afresh once https://github.com/angular-eslint/angular-eslint/pull/968 has been merged

JamesHenry:
Ok further update, that PR to alter the visitor keys was more fiddly than I thought, and the only way I could reliably not cause regressions in the existing i18n rule was to only temporarily patch the visitor keys within the node preprocessing logic.

I am guessing most if not all of this PR needs to be applied in order to arrive at the ultimate fix...

JamesHenry:
I just tried it and applying the visitor keys update does cause all your commented out tests here to pass, but there are still a couple of regressions that it introduces as well

JamesHenry:
This led me to discover something very interesting, some of the cases we currently have are actually conceptually invalid:

E.g.

```
    description:
      'should fail if `i18n-*` attribute is missing with nested elements',
    annotatedSource: `
      <p>Lorem ipsum <em i18n="@@dolor">dolor</em> sit amet.</p>
         ~~~~~~~~~~~~                              ^^^^^^^^^
    `,
    messages: [
      {
        char: '~',
        messageId: i18nAttributeOnIcuOrText,
        suggestions: [
          {
            messageId: suggestAddI18nAttribute,
            output: `
      <p i18n>Lorem ipsum <em i18n="@@dolor">dolor</em> sit amet.</p>
```

If you take that suggested fix of:

```html
      <p i18n>Lorem ipsum <em i18n="@@dolor">dolor</em> sit amet.</p>
```

and parse it with the Angular compiler, it will raise a `ParseError` against it:

image

So this raises a much broader topic that we don't currently check for `ParseError`s from the Angular compiler and throw accordingly. I will address that in its own PR now.

Will be interesting to see which of our existing tests produce invalid results!

JamesHenry:
Surfaced a couple of issues, I merged it here: https://github.com/angular-eslint/angular-eslint/pull/969

jaufgang:
@rafaelss95, @JamesHenry, has this PR been abandoned?  It's been over a year since issue #701 was reported, and an `<ng-template>` with an i18n attribute will still get a false-positive lint error from this rule.

JamesHenry:
I emailed Rafael about 2 months ago asking if wished to continue working on this but sadly have not received a response in that time. I will therefore close this so that it is clear that it is not being worked on and that somebody else can pick this up if it is important to them.

Many thanks to Rafael for his fantastic contributions to angular-eslint to date 🙏

Issue #596, opened 2021-07-14
Enhancement: automatically update every installed @angular-eslint package during `ng update @angular-eslint/schematics`


alebx:
When I'm trying to update the @angular-eslint dependencies via @angular-eslint/schematics the dependencies are not updated.

**Description and reproduction of the issue**
```sh
ng update @angular-eslint/schematics
Using package manager: 'npm'
Collecting installed dependencies...
Found 99 dependencies.
Fetching dependency metadata from registry...
    Updating package.json with dependency @angular-eslint/schematics @ "12.3.0" (was "12.2.2")...
  UPDATE package.json (9173 bytes)
✔ Packages successfully installed.

npm outdated               
Package                                 Current  Wanted  Latest  Location
@angular-eslint/builder                  12.2.2  12.3.0  12.3.0  webapp
@angular-eslint/eslint-plugin            12.2.2  12.3.0  12.3.0  webapp
@angular-eslint/eslint-plugin-template   12.2.2  12.3.0  12.3.0  webapp
@angular-eslint/template-parser          12.2.2  12.3.0  12.3.0  webapp
```

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@angular-eslint/schamtics`          | `12.2.2` |
| `ESLint`                           | `7.30.0` |
| `node`                             | `14.17.0` |

```sh
# Please run `npx ng version` in your project and paste the full output here:

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 12.1.1
Node: 14.17.0
Package Manager: npm 6.14.13
OS: linux x64

Angular: 12.1.1
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, localize, material
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1201.1
@angular-devkit/build-angular   12.1.1
@angular-devkit/core            12.1.1
@angular-devkit/schematics      12.1.1
@schematics/angular             12.1.1
ng-packagr                      12.1.2
rxjs                            6.6.7
[...]

PR #770, opened 2021-10-26
feat(eslint-schematics): update ng-update to include packageGroup

bryantlikes:
adding package group for ng update to address #596


JamesHenry:
Thanks a lot for bringing this to my attention @bryantlikes, I have included it in the v13 branch so the first prerelease of v13 I did yesterday includes this (and the last couple of packages that were missed off which it is possible a user could have installed if they are developing their own lint rules).

Thanks again!

PR #780, opened 2021-11-02
feat: angular-eslint v13

JamesHenry:
NOTE: We will publish prereleases from this branch and it will be subject to breaking changes right up until it is merged into master and v13.0.0 is cut.

Therefore please be aware that if you use the prereleases you will be responsible for navigating any breaking changes.

- [x] Angular v13 support
- [x] eslint v8 and typescript-eslint v5 support


JamesHenry:
The first prerelease has been published, anybody wanting to migrate to v13 can run:

```sh
npx ng update @angular-eslint/schematics --next
```

**After** they have already upgraded to Angular v13

PR #781, opened 2021-11-02
feat: eslint 8 support

JamesHenry:
This will be combined with the Angular 13 changes as it is also a breaking change


JamesHenry:
🟢  so merging into v13 branch

PR #804, opened 2021-11-15
feat(i18n): option to require description for i18n metadata

adbaran1:
This pull request adds the option to enforce the requirement of a description for i18n metadata. Per the Angular guide, "To translate a text message accurately, provide additional information or context for the translator." This is especially helpful when working in projects that contain large amounts of translatable texts.

P.S.
- If this feature is sound, would it be worth adding the option to require the meaning for i18n metadata? If so, in this PR, or a follow-up? Maybe something like: `"requireMetadata": ["meaning", "description"]`
- The logic for this feature adds a conditional statement with the existing logic for checking the ID. Even though all of the tests are still passing, please provide any feedback on the changes to `handleElement(...)`.

Any/all feedback welcome. 😄

### DIFF
--- a/packages/eslint-plugin-template/src/rules/i18n.ts
+++ b/packages/eslint-plugin-template/src/rules/i18n.ts
@@ -58,6 +58,7 @@ type Options = [
     readonly checkText?: boolean;
     readonly ignoreAttributes?: readonly string[];
     readonly ignoreTags?: readonly string[];
+    readonly requireDescription?: boolean;
   },
 ];
 export type MessageIds =
@@ -66,7 +67,8 @@ export type MessageIds =
   | 'i18nCustomIdOnAttribute'
   | 'i18nCustomIdOnElement'
   | 'i18nDuplicateCustomId'
-  | 'suggestAddI18nAttribute';
+  | 'suggestAddI18nAttribute'
+  | 'i18nMissingDescription';
 type StronglyTypedElement = Omit<TmplAstElement, 'i18n'> & {
   i18n: Message;
   parent?: AST;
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -196,17 +206,27 @@ export default createESLintRule<Options, MessageIds>({
...
-        return;
-      }
-
       const loc = parserServices.convertNodeSourceSpanToLoc(sourceSpan);
-      context.report({
-        messageId: 'i18nCustomIdOnElement',
-        loc,
-      });
+
+      if (!requireDescription) {
```
adbaran1, line 211: Perhaps this could be simplified...

alex-okrushko: Yes, I don't think you need the extra `if ... else ...` here. It could be that both customId and description are required and missing, so I assume we'd want to report both.

adbaran1: Without the `if ... else ...`, we get test results like the following:
```
    Expected value to strictly be equal to:
      1
    Received:
      2

    Message:
      Should have 1 error but had 2: [
      {
        ruleId: 'i18n',
        severity: 1,
        message: 'Missing custom ID on element. See more at https://angular.io/guide/i18n#manage-marked-text-with-custom-ids',
        line: 2,
        column: 7,
        nodeType: null,
        messageId: 'i18nCustomIdOnElement',
        endLine: 2,
        endColumn: 26
      },
      {
        ruleId: 'i18n',
        severity: 1,
        message: 'Missing i18n description on element. See more at https://angular.io/guide/i18n-common-prepare#i18n-metadata-for-translation',
        line: 2,
        column: 7,
        nodeType: null,
        messageId: 'i18nMissingDescription',
        endLine: 2,
        endColumn: 26
      }
    ]
```

This is because the `checkId` option is `true` by default. Because of this, the two avenues I considered were:
1. Add the conditional in the `handleElement` logic to separate the handling of IDs and other metadata
2. Add in the docs to set `checkId` to `false` if you want to require a description

Option 1 is a lot more user-friendly, and option 2 will become even more problematic if/when we add an option to make the meaning required.

alex-okrushko (approved):
Nice!

packages/eslint-plugin-template/src/rules/i18n.ts
alex-okrushko: (this was addressed offline)

JamesHenry (approved):
Thanks a lot @adbaran1 - I always feel like an imposter when reviewing these i18n rules, because I have never actually used the native i18n solution myself, but the PR looks good and I trust @alex-okrushko so let's merge this! :)

PR #807, opened 2021-11-17
feat: add package group for ng update

kyubisation:
`ng update` checks the "ng-update"."packageGroup" field in the package.json for related packages. If defined these will also be updated to the resolved version: https://github.com/angular/angular-cli/blob/master/docs/specifications/update.md#library-developers
This simplifies the update procedure; Consumers will only have to execute `ng update @angular-eslint/schematics`, which will update all installed `@angular-eslint/*` packages.

### DIFF
--- a/packages/schematics/package.json
+++ b/packages/schematics/package.json
@@ -18,6 +18,14 @@
   ],
   "schematics": "./dist/collection.json",
   "ng-update": {
+    "packageGroup": [
+      "@angular-eslint/builder",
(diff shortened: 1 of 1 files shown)


JamesHenry (approved):
I had already applied this in the v13 branch which is in prerelease, but we can go ahead and merge this as I am going to publish another 12 minor first

kyubisation:
Sorry, I hadn't noticed that. Thank you anyway for the merge. 😃

PR #808, opened 2021-11-17
feat(builder): expose nx executor without ng-compat layer

bgotink:
This exposes the nx executor directly for usage within nx, rather than running the command via two compatibility layers (the executor is turned into a builder & nx needs to run in an AngularCLI-like mode to execute builders).

This is similar to how it's set up in `@nrwl/linter` ([source](https://github.com/nrwl/nx/blob/d7af894681787432ff0ddfeff905d2517a93f333/packages/linter/executors.json)).


JamesHenry:
@bgotink Can you please explain to me the rationale behind this PR?

If you are using Nx you should use the Nx Executor exposed by `@nrwl/linter` (which I also created), if you are not using Nx, you should use the Angular CLI builder exposed by `@angular-eslint/builder`.

(The fact `@nrwl/devkit` is used here is a simply a convenience for me as the author as I personally find it more intuitive than `@angular-devkit`).

bgotink:
`@nrwl/linter` has a dependency on `@nrwl/jest`, using that package forces me to pull in a bunch of jest dependencies that I don't need. For comparison: adding `@angular-eslint/builder` to an empty repo installs 49 packages, adding `@nrwl/linter` instead installs a whopping 364 packages.

As a sidenote:

I'm not using the angular CLI nor nx but a separate tool that has support for nx executors & ng builders, but which doesn't use angular.json/workspace.json. That means the nx compat layer doesn't work, as it tries to find a configuration file that doesn't exist.

If I were using `nx`, the current setup would require `nx` to run an angular architect so it can run the builder, which then runs the builder. That leads to e.g. reading the workspace configuration multiple times (even though the executor here doesn't use the configuration at all).

JamesHenry:
@bgotink Ok interesting, please can you share more about this tool you mention?

Given this is a trivial addition, I am ok with adding it but please note that this is _not_ an officially supported use-case and I will not be adding any kind of integration tests for it and it could therefore break at any time.

bgotink:
@JamesHenry Thanks! Chances of something like this breaking are low, and I understand this isn't something you want to officially support. I'm just happy I'll be able to use this package without workaround!

The tool I mentioned is a side project of mine, a CLI called [sn](https://github.com/snuggery/snuggery/tree/main/packages/snuggery) with some features I was missing from `ng`:

- support for yarn 2/3 in PnP mode in the globally installed command
- support for running any target name as command, instead of being limited to `build`, `serve` and a few others. For example: at work I run `sn update-icons` instead of `ng run icons:update-icons`.
- an update mechanism loosely based on `nx`
- a bunch of documentation commands for builders, schematics, and migrations.
  <details>
  <summary>
   Example: <code>sn help builder @angular-eslint/builder:lint</code>
  </summary>

  ```
  Builder `lint` of package `@angular-eslint/builder`

  Run ESLint over a TypeScript project

  ━━━ Properties: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  - `eslintConfig`
    The name of the ESLint configuration file.
    Type: `string`

  - `fix`
    Fixes linting errors (may overwrite linted files).
    Default value is `false`
    Type: `boolean`

  - `cache`
    Only check changed files.
    Default value is `false`
    Type: `boolean`

  - `cacheLocation`
    Path to the cache file or directory.
    Type: `string`

  - `outputFile`
    File to write report to instead of the console.
    Type: `string`

  - `cacheStrategy`
    Strategy to use for detecting changed files in the cache.
    Default value is `"metadata"`
    Possible values:
    - `"metadata"`
    - `"content"`

  - `force`
    Succeeds even if there was linting errors.
    Default value is `false`
    Type: `boolean`

  - `quiet`
    Report errors only.
    Default value is `false`
    Type: `boolean`

  - `maxWarnings`
    Number of warnings to trigger nonzero exit code.
    Default value is `-1`
    Type: `number`

  - `silent`
    Hide output text.
    Default value is `false`
    Type: `boolean`

  - `lintFilePatterns` (required)
    One or more files/dirs/globs to pass directly to ESLint's lintFiles() method.
    Default value is `[]`
    Type: `string[]`

  - `format`
    ESLint Output formatter (https://eslint.org/docs/user-guide/formatters).
    Default value is `"stylish"`
    Type: `string`

  - `ignorePath`
    The path of the .eslintignore file.
    Type: `string`

  - `noEslintrc`
    The equivalent of the --no-eslintrc flag on the ESLint CLI, it is false by
    default
    Default value is `false`
    Type: `boolean`

  - `rulesdir`
    The equivalent of the --rulesdir flag on the ESLint CLI, it is an empty array
    by default
    Default value is `[]`
    Type: `string[]`

  - `resolvePluginsRelativeTo`
    The equivalent of the --resolve-plugins-relative-to flag on the ESLint CLI
    Type: `string`
  ```

  </details>

PR #840, opened 2021-11-19
feat!: add original component filename to inline-template filename

abierbaum:
This PR updates the code to use the name of the component file in the name of any created inline-template filenames.  This allows overrides to be created to adjust the lint rules used for inline-templates based upon the containing filename.

The goal of this change is to allow configurations to do things like apply different eslint rules for spec/test file inline templates than they do for standard templates.

I _think_ it would work something like this, but have not been able to fully test it and I am looking for feedback from the maintainers on this idea.

```
{
  "root": true,
  "ignorePatterns": ["projects/**/*"],
  "overrides": [
    {
      "files": ["*.ts"],
      "parserOptions": {
        "project": ["tsconfig.json", "e2e/tsconfig.json"],
        "createDefaultProgram": true
      },
      "extends": [
        "plugin:@angular-eslint/recommended",
        "plugin:@angular-eslint/template/process-inline-templates",
        "plugin:prettier/recommended"
      ],
      "rules": {}
    },
    // Apply to all html except spec html
    {
      "files": ["*.html"],
      "excludedFiles": ["*spec*.html"],
      "extends": ["plugin:@angular-eslint/template/recommended"],
      "rules": {
         "@angular-eslint/template/i18n": "error",
      }
    },
    // Apply these rules to spec html files
    {
      "files": ["*spec*.html"],
      "extends": ["plugin:prettier/recommended"],
      "rules": {
         "@angular-eslint/template/i18n": "off",
      }
    }
  ]
}
```

### DIFF
--- a/packages/eslint-plugin-template/src/processors.ts
+++ b/packages/eslint-plugin-template/src/processors.ts
@@ -1,4 +1,5 @@
 import ts from 'typescript';
+import { basename } from 'path';
 
 const rangeMap = new Map();
 
@@ -155,7 +156,8 @@ export function preprocessComponentFile(
         continue;
       }
 
-      const inlineTemplateTmpFilename = `inline-template-${++id}.component.html`;
+      const baseFilename = basename(filename);
+      const inlineTemplateTmpFilename = `inline-template-${baseFilename}-${++id}.component.html`;
 
       const start = templateProperty.initializer.getStart();
       const end = templateProperty.initializer.getEnd();
--- a/packages/eslint-plugin-template/tests/processors.test.ts
+++ b/packages/eslint-plugin-template/tests/processors.test.ts
@@ -135,7 +135,7 @@ describe('extract-inline-html', () => {
         ).toEqual([
           input,
           {
-            filename: 'inline-template-1.component.html',
+            filename: 'inline-template-test.component.ts-1.component.html',
             text: '<h1>Hello, A!</h1>',
           },
         ]);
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin-template/src/processors.ts
```diff
@@ -155,7 +156,8 @@ export function preprocessComponentFile(
         continue;
       }
 
-      const inlineTemplateTmpFilename = `inline-template-${++id}.component.html`;
+      const baseFilename = basename(filename);
+      const inlineTemplateTmpFilename = `inline-template-${baseFilename}-${++id}.component.html`;
```
abierbaum, line 160: I tried to keep this filename as consistent with the previous structure as possible so I left `.component` in place although it may look redundant in the final filenames.

abierbaum:
@JamesHenry or other maintainers, any thoughts on this approach?  Is there some other way to accomplish the same thing?

JamesHenry:
Thanks a lot for this @abierbaum, I'm very sorry for the delay in reviewing this. It's a breaking change and so has only just become eligible for release with v14 upon us (I nevertheless should have reiterated that we follow Angular's breaking change schedule before now, so sorry for that).

It seems reasonable to me to take this approach, so we can get this into v14.

Please can you change the branch your are PRing into to be `next-major-release/v14` instead of main?

JamesHenry (changes requested):
See comment

JamesHenry:
@abierbaum I was able to change the target branch, but unfortunately in looking into this the implementation was incomplete and it caused the integration tests to fail.

I have included a complete version of your change in the Angular 14 PR directly here: https://github.com/angular-eslint/angular-eslint/pull/1004/commits/89ec1b2fd29b2df1addde1a0e3f65a93cdd2a667

Thank you very much again for taking the initiative to submit this PR, it's a good change 👍

PR #844, opened 2021-11-23
feat(eslint-plugin-template): add require-localize-metadata rule

adbaran1:
Similar to #804, this pull request adds the option to enforce the requirement of a description for i18n metadata when using `$localize` in component code.

TODO:
- [x] Address all questions/comments below
- [x] Documentation

***Pull request is in draft mode; please see my comments below!***

FYI @JamesHenry , I know that you haven't used Angular i18n, but perhaps you could chime in on the creation of this new rule? See my comments for require-localize-metadata.ts.


packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+];
+
+const DEFAULT_OPTIONS: Options[number] = {
+  requireDescription: false,
+};
+
+const VALID_LOCALIZED_STRING_WITH_DESCRIPTION = new RegExp(/`:[^|]*:.*`/);
+
+// TODO: The following values are also used in eslint-plugin-template/i18n, and should likely be centralized
```
adbaran1, line 16: What is a good location for these values?

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    function testTaggedMessageToken(token: TSESTree.Token) {
+      if (requireDescription) {
+        // Test for i18n description only
+        testTaggedMessageTokenWithDescription(token);
+      }
+    }
+
+    return {
+      Program(programNode: TSESTree.Program) {
```
adbaran1, line 72: I'm still learning the ropes with ASTs, but I don't have a great feeling about using `TSESTree.Program`. Is there a better way?

`$localize` can be used in many different ways; see the test cases.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,87 @@
...
+  {
+    readonly requireDescription?: boolean;
+  },
+];
+
+const DEFAULT_OPTIONS: Options[number] = {
+  requireDescription: false,
+};
+
+const VALID_LOCALIZED_STRING_WITH_DESCRIPTION = new RegExp(/`:[^|]*:.*`/);
```
adbaran1, line 14: Any thoughts on this RegExp?

packages/eslint-plugin/tests/rules/require-localize-metadata/cases.ts
```diff
@@ -0,0 +1,82 @@
...
+      return $localize\`:An introduction header for this sample:Hello i18n!\`;
+    `,
+    options: [{ requireDescription: true }],
+  },
+  {
+    code: `
+      someFunction($localize\`:An introduction header for this sample:Hello i18n!\`);
+    `,
+    options: [{ requireDescription: true }],
+  },
```
adbaran1, line 35: Extra test cases: meaning and/or ID with and without description.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
adbaran1: Discovered that we can check `TaggedTemplateExpression`; will refactor for this.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,103 @@
...
+        if (
+          requireDescription &&
+          !!identifierName &&
+          identifierName === '$localize' &&
+          !!templateElement
+        ) {
+          testLocalizeTemplateElement(templateElement);
+        } else {
+          return;
+        }
```
aokrushk, line 99: Else part is not needed.

```suggestion
        }
```

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,103 @@
...
+        taggedTemplateExpression: TSESTree.TaggedTemplateExpression,
+      ) {
+        const identifierName = (
+          taggedTemplateExpression.tag as TSESTree.Identifier
+        ).name;
+        const templateElement = taggedTemplateExpression.quasi.quasis[0];
+
+        if (
+          requireDescription &&
+          !!identifierName &&
```
aokrushk, line 92: this condition is not needed because `identifierName === '$localize'` covers it.

```suggestion
```

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,103 @@
...
+        testLocalizeTemplateElementWithDescription(templateElement);
+      }
+    }
+
+    return {
+      TaggedTemplateExpression(
+        taggedTemplateExpression: TSESTree.TaggedTemplateExpression,
+      ) {
+        const identifierName = (
+          taggedTemplateExpression.tag as TSESTree.Identifier
```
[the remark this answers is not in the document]
adbaran1: Because `tag` is of type `LeftHandSideExpression`, which is:
```
export declare type LeftHandSideExpression = ArrayExpression | ArrayPattern | ArrowFunctionExpression | CallExpression | ClassExpression | FunctionExpression | Identifier | JSXElement | JSXFragment | LiteralExpression | MemberExpression | MetaProperty | ObjectExpression | ObjectPattern | SequenceExpression | Super | TaggedTemplateExpression | ThisExpression | TSAsExpression | TSNonNullExpression | TSTypeAssertion;
```

packages/eslint-plugin/src/rules/require-localize-metadata.ts
JamesHenry: I would remove the comment, I think having the small bit of repetition is fine and allows for them to remain independent and be more easily refactored if the structure of Angular guides changes in future

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,100 @@
...
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      requireLocalizeMetadata: `$localize tagged messages should contain a description. See more at ${STYLE_GUIDE_LINK_METADATA_FOR_TRANSLATION}`,
+    },
+  },
+  defaultOptions: [DEFAULT_OPTIONS],
+  create(context, [{ requireDescription }]) {
+    function reportRequireLocalizeMetadataError(
```
JamesHenry, line 53: Please can you refactor this implementation to just inline everything in the Node handler method?

There is very little code here but the dedication to wrapping every step into functions hurts readability in my opinion because of how it changes the flow of the document.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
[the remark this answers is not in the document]
JamesHenry: @adbaran1 It would be better to narrow the type through control flow analysis by actually checking the shape of the tag to check that it is an identifier.

You can use the utility `TSESLintASTUtils.isIdentifier(...)` from `@typescript-eslint/experimental-utils` to do exactly what you need and TypeScript will pick up on the types through the control flow analysis.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -0,0 +1,100 @@
...
+    docs: {
+      description:
+        'Ensures that $localize tagged messages contain a description.',
+      recommended: false,
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          requireDescription: {
```
JamesHenry, line 39: I don't understand what purpose this option serves given the whole rule is dedicated to this purpose?

This is further backed up by the fact you don't have tests for when it is disabled.

I would think that if someone does not want to require the description they would simply not use this rule right?

JamesHenry (changes requested):
Thanks a lot for this @adbaran1, I'm really sorry for how long it has taken me to get back to you on this.

I moved countries a couple of times so things have been very chaotic.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
adbaran1: @JamesHenry no worries, and I hope things are less chaotic now 😄 . The reason for this set up is to be able to support custom ID and meaning, which are the other two i18n metadata parameters (see [i18n metadata for translation](https://angular.io/guide/i18n-common-prepare#i18n-metadata-for-translation)). This pull request is only to support description for the time being, which is why the implementation looks the way it does (i.e. having functions like `testLocalizeTemplateElementWithDescription`).

With that all said, let me know what you think about the overall approach.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
adbaran1: See my comment below about eventually being able to support custom ID and meaning.

aokrushk: Makes sense @adbaran1 👍

aokrushk (approved):

alex-okrushko:
@JamesHenry Looks like Adrian has addresses all the comments. Anything else needs to be done before it's merged?
Thanks!

JamesHenry:
Thanks again for your patience on this @adbaran1 @alex-okrushko.

I think the description of the overall rule matching the intent behind a subset of its intended features is still confusing but I will amend that post-merge

JamesHenry (approved):

adbaran1:
Thank you @JamesHenry. Please let us know how we can assist you in making it all better. We can look to add support for Angular i18n meaning and IDs to make the overall setup more clear ASAP.

PR #868, opened 2022-01-12
feat(i18n): add checkDuplicateId option to i18n rule

### LINKED ISSUE #867: [i18n] add option to disable duplicate id check

Currently the i18n rule will always report errors when custom ids are used multiple times.

However there may be cases where this is not desired, for example when using common labels throughout the application

```html
<button type="button" i18n="@@common.cancel-label">Cancel</button>

<!-- later in the same file-->

<button type="button" i18n="@@common.cancel-label">Cancel</button>
```

Angular already reports duplicate ids which do not have the same content. So an option should be introduced to disable the duplicate id check.

```JSON
{
  "rules": {
    "@angular-eslint/template/i18n": ["warn"]
  }
}
```

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `13.0.1` |
| `@angular-eslint/template-parser`         | `13.0.1` |
| `@typescript-eslint/parser`               | `4.33.0` |
| `ESLint`                                  | `7.32.0` |
| `node`                                    | `14.18.2` |

marcuskrahl:
This adds an option to the i18n linting rule in order to oppress duplice id reports

fix #867


JamesHenry:
Thanks @marcuskrahl!

PR #892, opened 2022-02-08
fix(i18n): No longer throws exception when compiler returns null i18n description

### LINKED ISSUE #891: i18nMissingDescription rule throws exception if compiler returns null i18n description
 
Currently, the angular compiler is stripping the i18n description when the i18n attribute is already nested within another i18n under certain conditions. I encountered this when I ran the lint rule that was nested within another i18n *and* within an ng-template (structural directive also triggers the issue).

I was able to reproduce the issue using an annotated failure case:
```
export const invalid = [
    description: 'should fail if there is a nested `i18n-*` attribute within an ng-template.', 
    annotatedSource: `
    <span i18n="outer i18n" > 
      <ng-template>
        <div>  
          <span i18n="nested i18n descriptions are stripped within a ng-template">inner</span> 
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        </div>
      </ng-template>
    </span>    
    `,
    messageId: i18nMissingDescription,
    options: [{checkId: false,requireDescription: true }], 
  }),
];
```

Note, a nested i18n is consider a compilation issue, and the angular compiler doesn't seem to return location information that matches the annotated location, at least in this case. I suspect this might be an issue for any templates with compilation issues, which is probably just the way it is. So, you can use the above failure case to reproduce the null description issue, but we will not be able to keep it in the angular-eslint test suite.

When I originally encountered this issue running the linter against my actual code, it produced a "TypeError: Cannot read properties of undefined (reading 'trim')" error output (with a not terribly helpful stack trace), and the linter stopped execution. This left me with very little info on where the issue occurred in my code, and it took me a while to track down what was causing the linter to crash. If I was running with AOT set to true or if my text editor ran the linter against the template, the issue would've been flagged as a compilation issue and quickly resolved, fwiw.

I will submit a PR for this shortly.

```JSON
{
  "rules": {
    "@angular-eslint/template/<rule>": ["<setting>"]
  }
}
```

```HTML
// your repro code case
```

**Versions**
Reproduced in latest version & 12.7.0 of angular-eslint. Angular 12.2.7, Angular 13.x (I think). Let me know if more versioning info needed, but this is likely sufficient.

```sh


```

jamjarvi:
Closes #891.

Note, I attempted to include a test case for this specific issue, but the returned location info was always out of alignment with the  annotated test case location, I believe because my test case always contained a template compilation issue.


packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -422,7 +422,7 @@ function isBooleanLike(value: string): value is 'false' | 'true' {
 }
 
 function isEmpty(value: string) {
-  return value.trim().length === 0;
+  return value == null || value.trim().length === 0;
```
adbaran1, line 425: Would it be worth to have something like the following?
```js
if (!!value) {
  return value.trim().length === 0;
}

return true;
```

jamjarvi: I suppose that could save us some unnecessary calls to ```trim```, so it might be slightly more performant.

PR #928, opened 2022-02-23
feat(eslint-plugin-template): [button-has-type] add rule

chernodub:
closes #823


chernodub:
Hi @JamesHenry, I'd appreciate if you could look into this PR. So that we could continue working on it or, at least, close it. Thanks!

packages/eslint-plugin-template/tests/rules/button-has-type/cases.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { MessageIds } from 'packages/eslint-plugin-template/src/rules/button-has-type';
+import { INVALID_TYPE_DATA_KEY } from 'packages/eslint-plugin-template/src/rules/button-has-type';
+
+// https://www.w3.org/TR/html401/interact/forms.html#adef-type-BUTTON
+// The types are intentionally duplicated in order to de-couple the tests from the code
+const VALID_BUTTON_TYPES: readonly string[] = ['button', 'submit', 'reset'];
+const missingType: MessageIds = 'missingType';
+const invalidType: MessageIds = 'invalidType';
+
+const validCasesWithTextAttribute = VALID_BUTTON_TYPES.map(
```
JamesHenry, line 11: Please remove these abstractions in favour of duplication and keep everything inline in the valid cases array. Folks shouldn't have to parse test cases as they read or update them, it adds unnecessary mental overhead and harms the readability of the tests

packages/eslint-plugin-template/src/rules/button-has-type.ts
```diff
@@ -0,0 +1,126 @@
...
+function isTypeAttributePresentInElement({
+  inputs,
+  attributes,
+}: TmplAstElement): boolean {
+  return [...inputs, ...attributes].some(
+    ({ name }) => name === TYPE_ATTRIBUTE_NAME,
+  );
+}
+
+function getInvalidButtonTypeIfPresent(
```
JamesHenry, line 87: I found this difficult to read with there being nested closures which do not leverage any aspect of their outer scope. Additionally `lookIn` is not a well established semantic prefix (unlike `get` or `find` for example) so it's way more open to personal interpretation - I wouldn't have necessarily expected a function starting with `lookIn` to return a value, for example...

Speaking of returning a value, you are directly returning a `TextAttribute` value to represent a type of `InvalidButtonTypeInfo` in `lookInTextAttributes` instead of creating its more specific shape (unlike what you do in the `lookInBoundAttributes` function just below).

In summary then I think keeping it simple would be preferable here, e.g. something like:

```ts
function getInvalidButtonTypeIfPresent(
  element: TmplAstElement,
): InvalidButtonTypeInfo | null {
  const invalidTextAttribute = element.attributes.find(
    ({ name, value }) =>
      name === TYPE_ATTRIBUTE_NAME && !VALID_BUTTON_TYPES.includes(value),
  )
  if (invalidTextAttribute) {
    return {
      value: invalidTextAttribute.value,
      sourceSpan: invalidTextAttribute.sourceSpan,
    };
  }

  for (const { name, value, sourceSpan } of element.inputs) {
    // Intentionally ignore the property binding by looking for literal primitives only
    // in order to avoid type-checking for the property, e.g. check that it's 'submit', 'button', etc.
    if (
      name === TYPE_ATTRIBUTE_NAME &&
      value instanceof ASTWithSource &&
      value.ast instanceof LiteralPrimitive &&
      !VALID_BUTTON_TYPES.includes(value.ast.value)
    ) {
      return {
        value: value.ast.value,
        sourceSpan,
      };
    }
  }

  return null;
}
```

packages/eslint-plugin-template/src/rules/button-has-type.ts
```diff
@@ -0,0 +1,126 @@
...
+import {
+  createESLintRule,
+  getTemplateParserServices,
+} from '../utils/create-eslint-rule';
+
+export const RULE_NAME = 'button-has-type';
+export const INVALID_TYPE_DATA_KEY = 'type';
+export type MessageIds = 'invalidType' | 'missingType';
+
+interface InvalidButtonTypeInfo {
```
JamesHenry, line 22: I really don't think these comments add any value here (especially the source span) and they can be removed

JamesHenry (changes requested):
I'm very sorry for the delay in getting back to you on this @chernodub and thank you so much for taking the initiative and contributing this rule.

Please take a look at the review comments. You will also need to run `yarn update-rule-docs` in order to generate the necessary rule documentation to accompany this rule.

chernodub:
Thanks a lot for a valuable feedback @JamesHenry! 🙏

I've updated the code following your comments (agreed to all the points you mentioned). I've also rebased the changes on `origin/master` branch, so it's ready for merging.

JamesHenry (approved):

JamesHenry:
Thanks a lot @chernodub! I'll release this on v13 so you're not forced to upgrade to v14 of Angular to use it

PR #968, opened 2022-04-03
fix(parser): ensure attributes of ng-template elements get preprocessed

JamesHenry:
I took a snapshot of the current behavior and then this is the diff when applying this change:

image


JamesHenry:
This has ended up forming a much broader discussion around the i18n rule. Holding off on this for now until I hear back from @rafaelss95

PR #969, opened 2022-04-03
feat(parser): propagate parse errors from angular compiler

JamesHenry:


JamesHenry:
@rafaelss95 The rogue `"` within one of the valid test cases was causing it to fail to pick up on a legitimate failure within the source.

I am not sure whether the current behavior was intentional or not, but I documented it and made it much more explicit by adding some extra tests. If we want to talk about changing the behavior to match your original test contents (minus the extra `"`) then we definitely can, but I felt that making the current behavior more explicit was the better way to proceed for now.

packages/eslint-plugin-template/tests/rules/click-events-have-key-events/cases.ts
```diff
@@ -17,12 +17,11 @@ export const valid = [
     code: '<cui-button (click)="onClick()"></cui-button>',
   },
   {
-    // It should work when element has aria-hidden.
+    // It should work when element has a static value for aria-hidden.
     code: `
-        <div (click)="onClick()" aria-hidden"></div>
```
JamesHenry, line 22: @rafaelss95 See my comment about this rogue `"`

packages/eslint-plugin-template/tests/rules/accessibility-alt-text/cases.ts
```diff
@@ -12,9 +12,9 @@ export const valid = [
   '<object aria-labelledby="id1">',
   '<object>Meaningful description</object>',
   '<object title="An object">',
-  '<area aria-label="foo"></area>',
-  '<area aria-labelledby="id1"></area>',
-  '<area alt="This is descriptive!"></area>',
+  '<area aria-label="foo" />',
```
JamesHenry, line 15: Area is a void element and cannot have a closing tag

PR #988, opened 2022-04-20
feat(i18n): include i18n-* with the require description rule

smithpb:
This pull request is to expand on the existing rule to require a description on the i18n attribute. With these changes, the i18n-* will also require a description.

### DIFF
--- a/.github/workflows/auto-close-dependabot-prs-targeting-fixtures.yml
+++ b/.github/workflows/auto-close-dependabot-prs-targeting-fixtures.yml
@@ -11,6 +11,10 @@ on:
     paths:
       - 'packages/integration-tests/fixtures/**'
 
+# Need to give extended permissions to secrets.GITHUB_TOKEN
+permissions:
+  pull-requests: write
+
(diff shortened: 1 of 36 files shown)


alex-okrushko:
Overall looks good. The only concern that I have is that we are getting into a 3-rd level of `if` nesting. Anything we can do about that?

smithpb:
With the nature of the types of checks being performed, refactoring to consolidate if-statements I feel would make things needlessly complicated and difficult to read.

smithpb:
@JamesHenry I'm submitting a PR to hopefully get merged. Thanks.

JamesHenry (approved):
Thank you @smithpb!

JamesHenry (changes requested):
Sorry, actually looks like a docs regen is needed:

```sh
yarn update-rule-docs
```

and commit the result

JamesHenry:
Thanks again @smithpb - just FYI I updated the PR title (and squash commit) to better reflect the scope and change here

PR #1000, opened 2022-05-19
fix(eslint-plugin-template): [eqeqeq] update suggest message

amayer42:
Fix an issue where the suggest message for the `eqeqeq` rule suggests that you replace the "expected" value with the "actual" value instead of the other way around.

## Before
screenshot of issue

## After
screenshot of fix


JamesHenry (approved):
Thanks @amayer42!

amayer42:
And thank you for the approval @JamesHenry!

PR #1056, opened 2022-06-22
feat(eslint-plugin-template): [quotes] add rule

ShacharHarshuv:
Hey,
I added a rule to enforce use of double / single quotes in angular template. 
For example, by default, this is valid:
```html
<div class="hello"></<div>
```
This is not:
```html
<div class='hello'></div>
```

This is my first time contributing for this library and I don't have much experience with writing ESLint rules so feel free to let me know if I implemented something in a less ideal / unconventional way, and I'll fix it. I tried my best to imitate the coding style of the existing rules.


sharshuv-quotient:
@rafaelss95 Can you / someone else please review? Thanks 🙏

jerone:
There is a third option to consider which is valid HTML5: [no quotes](https://dev.w3.org/html5/spec-LC/syntax.html#unquoted).

Does this also work with quotes inside the expression?

ShacharHarshuv:
> There is a third option to consider which is valid HTML5: [no quotes](https://dev.w3.org/html5/spec-LC/syntax.html#unquoted).

Good point. I'll revise the auto-fixer so it can handle no-quotes (and add them as necessary).

> Does this also work with quotes inside the expression?

Do you mean something like that:
```html
<div [ngClass]="'col-' + colNum">
```
?
If yes then this will pass this rule with `"quotesType": "double"` as I'm only checking the outer quotes. 
I think if we want to enforce quotes inside the expression it should be a different rule, as technically those quotes are javascript code, not html.

jerone:
> Do you mean something like that:

Yes, exactly. Or an static example:
```html
<div title="Hello's world"></div>
```

The autofixer will replace the outer quotes, creating invalid HTML and/or breaking the inner value/expression. 
I would disable the autofixer with a simple check if the value/expression contains an opposite double or single quote. This would still keep the warning/error for the developer to fix manually.
In my opinion, the autofixer should only fix what is really fixable without creating more errors.

ShacharHarshuv:
I see what you mean. I think it should be possible to replace only the correct quotes (start and end) because the AST marks the span of the "value". I'll look into it, and if I can't find a viable solution I'll disable the autofixer in that case.

jerone:
> I see what you mean. I think it should be possible to replace only the correct quotes (start and end) and escape any characters a necessary:
> 
> ```
> <div title='Hello\'s world'></div>
> ```

That would indeed work for `TextAttribute`. 

Don't forget about already escaped values.

PR #605, opened 2021-07-21
feat(eslint-plugin-template): [attributes-order] add rule

nickbullock:
closes #127 

State: rule and tests are ready, fixer is not ready.
Let's discuss default order here.

PR #1066, opened 2022-06-24
feat(eslint-plugin-template): [attributes-order] add rule with fixer

### LINKED ISSUE #127: Rule proposal: template/attributes-order
It makes sense for structural directives like `*ngIf` to be first in the list of attributes:
```
<div [ngClass]="item.classes" my-decorator *ngIf="item.isVisible" [my-decorator-extras]="item.extras">
  {{ item.text | translate }}
</div>
```

I don't promise anything, but if I have the time, I'll look into it this month.

P.S. Some issue templates could be useful. I'll also look into this.

P.S.2 Thanks for all the hard work!

pmccloghrylaing:
Continuation of #605

Fixes #127


lachieh:
@rafaelss95 @JamesHenry Any updates? Looks like the original issue for this was closed (#127). Would be great to get this rule in.

packages/eslint-plugin-template/src/rules/attributes-order.ts
```diff
@@ -0,0 +1,361 @@
...
+  alphabetical: false,
+  order: [...DEFAULT_ORDER],
+};
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'layout',
+    docs: {
+      description: 'Ensures that html attributes are sorted correctly',
```
JamesHenry, line 82:
```suggestion
      description: 'Ensures that HTML attributes and Angular bindings are sorted based on an expected order',
```

packages/eslint-plugin-template/src/rules/attributes-order.ts
```diff
@@ -0,0 +1,361 @@
...
+            minItems: DEFAULT_OPTIONS.order.length,
+            uniqueItems: true,
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      attributesOrder:
+        'Attributes order is incorrect, expected {{expected}} instead of {{actual}}',
```
JamesHenry, line 109:
```suggestion
        'The element's attributes/bindings did not match the expected order: expected {{expected}} instead of {{actual}}',
```

packages/eslint-plugin-template/src/utils/is-on-same-location.ts
```diff
@@ -0,0 +1,14 @@
+import type {
+  TmplAstBoundAttribute,
+  TmplAstBoundEvent,
+} from '@angular-eslint/bundled-angular-compiler';
+
+export function isOnSameLocation(
```
JamesHenry, line 6: How come this utility was singled out to be promoted to its own file?

JamesHenry (changes requested):
I'm very sorry for the delay in reviewing this @pmccloghrylaing.

I feel that the language of "incorrect/correct" is too strong for this rule, as this is subjective with no runtime behaviour impact, so let's please use the suggested edits to make it more about "expectations" not being met.

You will also need to update the branch to the latest.

Many thanks!

JamesHenry:
@pmccloghrylaing sorry one of my suggestions made via the github UI was borked because of the single quote usage within the string, you'll need to fix up locally

packages/eslint-plugin-template/src/utils/is-on-same-location.ts
pmccloghrylaing: Can't recall a reason - I'll move it to `attributes-order.ts`

JamesHenry:
Thanks @pmccloghrylaing this LGTM now, we just need to generate the rule docs then this can be merged.

You just need to run `yarn update-rule-docs` and commit the result

pmccloghrylaing:
@JamesHenry - that's done. 

Only issue I had is the relative path generated on Windows uses `\` instead of `/`.

JamesHenry:
Thanks again @pmccloghrylaing!

PR #1100, opened 2022-07-27
feat(eslint-plugin-template): [accessibility-role-has-required-aria] add rule

sandikbarr:
- Adds a new accessibility rule for checking roles to have the required `aria-` properties.
  - Ported from [eslint-plugin-jsx-a11y role-has-required-aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-has-required-aria-props.md)
- The accessibility-role-has-required-aria rule works as follows:
  - Find DOM elements in angular templates with a role attribute
  - Check role for missing aria- props using definitions from 'aria-query'
  - Check an element's semantic role and implied aria- using definitions from 'axobject-query'
  - Report violation with role and required aria- props
  - Provide suggested fix to remove the role attribute


packages/eslint-plugin-template/src/rules/role-has-required-aria-props.ts
```diff
@@ -0,0 +1,95 @@
...
+  createESLintRule,
+  getTemplateParserServices,
+} from '../utils/create-eslint-rule';
+import { getDomElements } from '../utils/get-dom-elements';
+import { toPattern } from '../utils/to-pattern';
+import { isSemanticRoleElement } from '../utils/is-semantic-role-element';
+
+type Options = [];
+export type MessageIds = 'roleHasRequiredAria' | 'suggestRemoveRole';
+export const RULE_NAME = 'role-has-required-aria-props';
```
sandikbarr, line 17: I ported this rule from [eslint-plugin-jsx-a11y role-has-required-aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-has-required-aria-props.md) and used the same name, but I've considered dropping "-props" from the name and just calling it "role-has-required-aria" since props is a React convention. Let me know if that are opinions one way or another.

JamesHenry:
Hi @sandikbarr 👋 

Long time no speak! Hope you are doing well. I'm really sorry for how long it's taken to review this.

In order to match our existing pattern for these kinds of rules, let's please go for `accessibility-role-has-required-aria`.

Based on existing work by Rafael and others I think we wanted to lazy load `axobject-query` only if it was needed. I won't apply that as a blocker here though, I can tackle that in a follow up

JamesHenry (changes requested):
See comment regarding rule name

sandikbarr:
Yes! Hello @JamesHenry! No worries about the time. This helped us track down some issues and verify others, so thought I'd share it back with the rest of the accessibility rules. 
* Renamed the rule to `accessibility-role-has-required-aria`
* Updated the util to lazy load `axobject-query` and also added _some_ local typing to the AXObjects being used to help with readability.

Thanks so much for the feedback!

packages/eslint-plugin-template/src/rules/role-has-required-aria-props.ts
sandikbarr: renamed to `accessibility-role-has-required-aria`

JamesHenry (approved):

JamesHenry:
Thanks @sandikbarr!

PR #1101, opened 2022-07-28
fix(eslint-plugin-template): [click-events-have-key-events]: handle additional outputs

sandikbarr:
Fixes a bug where the `click-events-has-keyboard-events` rule is allowing elements to pass as valid that have outputs/events in addition to their (click) events without an associated with key event because as it loops through the outputs, any additional node output is resetting `hasClick` back to false and the rule returns with `!hasClick`.


packages/eslint-plugin-template/tests/rules/click-events-have-key-events/cases.ts
```diff
@@ -6,7 +6,11 @@ const messageId: MessageIds = 'clickEventsHaveKeyEvents';
 export const valid = [
   {
     // It should work when click events are associated with key events.
-    code: '<div (click)="onClick()" (keyup)="onKeyup()"></div>',
+    code: `
+        <div (click)="onClick()" (keyup)="onKeyup()"></div>
+        <div (keyup)="onKeyup()" (click)="onClick()"></div>
```
sandikbarr, line 11: Additionally found that the order of click and key events shows as invalid if the key event came before the click event. As a workaround, we're putting the key event after the click event, but based on the initial finding here with additional outputs, any event/output after the click event could show a false positive.

JamesHenry (approved):

JamesHenry:
Thanks @sandikbarr!

PR #1104, opened 2022-08-04
feat(eslint-plugin-template): [no-duplicate-attributes] Add option to ignore properties

ssams:
This introduces a new option `ignore` on the `@angular-eslint/template/no-duplicate-attributes` rule to ignore reports from selected duplicate attributes.

### Rationale

In some special cases presence of duplicate attributes may be required, this allows to keep the rule enabled whilst being capable of not generating errors/warnings on accepted duplicates.

For example, in our project we regularly need to apply both a static text attribute `class="foo"` together with a bound `[class]=dynamicValue`. In our project this is necessary to ensure the static class is always present, also including it in the dynamic property handled via the `ngClass` would cause it to appear only after all inputs have been updated, which in some cases can cause visual flickering due to the toggling of the class by the directive. Hence we need to be able to ignore reports on the `class` attribute, but want to keep using the rule for others.


JamesHenry (changes requested):
Thank you @ssams. I'm not super familiar with the behaviour of the framework in the example you give, but adding this option is low effort and low risk so let's press ahead with it.

Please remove the different data types for `ignore`, an array of strings is all that is needed and simplifies both the usage and implementation

ssams:
ok, I've changed the type to an array as requested

JamesHenry:
Thanks a lot @ssams - we just need to regenerate the docs now that the options have changed.

You just need to run `yarn update-rule-docs` and commit the results

ssams:
> Thanks a lot @ssams - we just need to regenerate the docs now that the options have changed.
> 
> You just need to run `yarn update-rule-docs` and commit the results

done :-)

JamesHenry (approved):
Thanks a lot @ssams!

PR #1118, opened 2022-08-25
chore: add description to invalid tests

jerone:
Allow the tests description to be used in tests output.

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>

image
   </td>
	<td>

image
</td>
</table>

_Tip: the valid tests can be extended with an `name` to also show a description._


JamesHenry:
Thank you @jerone!

PR #1134, opened 2022-09-16
feat(eslint-plugin-template): [accessibility-interactive-supports-focus] add rule

sandikbarr:
- Adds a new accessibility rule to check that interactive elements are focussable.
  - Ported from [eslint-plugin-jsx-a11y interactive-supports-focus](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/interactive-supports-focus.md)
  - This rule ensures that DOM elements with interaction handlers (click, keydown, keyup, keypress) can receive focus either because the element is interactive by default (button, input, etc.) or by adding a tabindex.
- This rule is intended to be used in conjunction with the template rule [click-events-have-key-events](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/docs/rules/click-events-have-key-events.md)
  - click-events-have-key-events ensures that when DOM elements are given click handlers that are not interactive by default that they also have a corresponding key event; however, elements that are not interactive by default also do not receive keyboard focus without tabindex.
  - Also, see my fix for [click-events-have-key-events](https://github.com/angular-eslint/angular-eslint/pull/1101)

The accessibility-interactive-supports-focus rule works as follows:
- Checks only native HTML DOM elements
- Looks for a tabindex attribute or input
- Looks for interactive outputs, click, keyup, keydown, or key press
- If the element has no interactive outputs, is disabled, aria-disabled, has a presentation role, or is hidden from the screen reader, it returns as valid
- If the element has interactive outputs, does not have a tab index, is not `contenteditable`, and is not an interactive element by default or is not explicitly assigned a non interactive role, it returns an error that the element must be able to receive focus
  - The fix is to either use a DOM element that is interactive by default, or
  - Add a tabindex to the element to make it focusable. However, it still may lack semantic meaning, so using semantic elements should be preferred.


packages/eslint-plugin-template/src/rules/interactive-supports-focus.ts
```diff
@@ -0,0 +1,86 @@
...
+  defaultOptions: [],
+  create(context) {
+    return {
+      Element$1(node: TmplAstElement) {
+        const elementType = node.name;
+        if (!getDomElements().has(elementType)) {
+          return;
+        }
+
+        const tabIndex = [...node.attributes, ...node.inputs].find(
```
adbaran1, line 43: It looks like `tabIndex` is not used until the second conditional. Let's move it to line 66?

packages/eslint-plugin-template/tests/rules/interactive-supports-focus/cases.ts
```diff
@@ -0,0 +1,254 @@
+import { convertAnnotatedSourceToFailureCase } from '@angular-eslint/utils';
+import type { MessageIds } from '../../../src/rules/interactive-supports-focus';
+
+const messageId: MessageIds = 'interactiveSupportsFocus';
+
+export const valid = [
+  // no interactive outputs
+  { code: '<div></div>' },
+
```
adbaran1, line 9: In the description, you mentioned that an element is valid if it is disabled. So what if a user ends up with something like `<a disabled>Click me</a>`? Is this considered valid? Do we want a test for it?

packages/eslint-plugin-template/tests/rules/interactive-supports-focus/cases.ts
```diff
@@ -0,0 +1,254 @@
...
+      <div role="tab" tabindex="0" (click)="onClick()"></div>
+      <div role="textbox" tabindex="0" (click)="onClick()"></div>
+    `,
+  },
+];
+
+export const invalid = [
+  // aria-hidden="false"
+  convertAnnotatedSourceToFailureCase({
+    description: 'should fail not hidden from screen reader',
```
adbaran1, line 153: Nit: can we make these descriptions unique?

packages/eslint-plugin-template/tests/rules/interactive-supports-focus/cases.ts
sandikbarr: That's a good catch. I will add a valid test case for the disabled attribute.
The `disabled` attribute is only valid on `<button>`, `<fieldset>`, `<input>`, `<optgroup>`, `<option>`, `<select>`, and `<textarea>` elements. All of those elements pass the IsInteractiveElement test and would receive focus by default when not disabled, but adding disabled to those elements would remove them from the focus ring.

packages/eslint-plugin-template/src/rules/accessibility-interactive-supports-focus.ts
```diff
@@ -0,0 +1,88 @@
...
+        const tabIndex = [...node.attributes, ...node.inputs].find(
+          (attr) => attr.name === 'tabindex',
+        );
+
+        if (
+          interactiveOutput &&
+          !tabIndex &&
+          !isInteractiveElement(node) &&
+          !isNonInteractiveRole(node) &&
+          !isContentEditable(node)
```
sandikbarr, line 73: Update to support `contenteditable`: I was realizing I had used only `(click)` events in my test cases and updated some to use `keyup`, `keydown`, and `keypress` events and realized those key events might be used for `contenteditable` which is focusable and interactive by default.

packages/eslint-plugin-template/src/configs/all.json
```diff
@@ -3,6 +3,7 @@
   "rules": {
     "@angular-eslint/template/accessibility-alt-text": "error",
     "@angular-eslint/template/accessibility-elements-content": "error",
+    "@angular-eslint/template/accessibility-interactive-supports-focus": "error",
```
sandikbarr, line 6: I have updated the rule name to follow the `accessibility-` name prefix convention.

JamesHenry:
@adbaran1 are you happy with the current state of this PR?

adbaran1:
> @adbaran1 are you happy with the current state of this PR?

Looks good to me! Thank you for making the adjustments @sandikbarr!

adbaran1 (approved):

JamesHenry (approved):
Thanks @sandikbarr and @adbaran1!

PR #1162, opened 2022-09-28
feat(eslint-plugin-template): [no-inline-styles] add rule

pumano:
It's (subjectively) good practice to separate styles from the view layout, when possible. This rule detects inline styles in angular templates and help to keep all css in one place (for example in .css files, or use tailwindcss classes). Also inline styles has different priority in CSS and that cause some bugs. 

I also found some rule https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-inline-styles.md but that rule for JSX, and I created own using @angular-eslint tooling and in using angular-eslint style.

Previously I create that rule as custom rule for my nx workspace rules (for our team needs), but have idea to share it with community.


pumano:
@JamesHenry could you check this PR please :)

JamesHenry:
Thinking out load, should this rule also prohibit `ngStyle` usage (based on my understanding of your intent with it)?

Maybe that could be configured via an option on the rule? 🤔 WDYT?

pumano:
> Thinking out load, should this rule also prohibit `ngStyle` usage (based on my understanding of your intent with it)?
> 
> Maybe that could be configured via an option on the rule? 🤔 WDYT?

Nice idea @JamesHenry, but I have no experience to create rule with params, I will try to add this tomorrow, that's really needed

jerone:
And what about the `style` prefix: https://angular.io/guide/class-binding#binding-to-a-single-style

```html
<nav [style.background-color]="expression"></nav>
```

Aldo, I would actually allow this. 
Maybe also enabling this case via an option.

pumano:
> Thinking out load, should this rule also prohibit `ngStyle` usage (based on my understanding of your intent with it)?
> 
> Maybe that could be configured via an option on the rule? 🤔 WDYT?

`ngStyle` support added via options

jerone:
I understand this might be a bit out of scope for the first iteration of this feature. Might be a good addition for later. 
Anyways, some information for future iterations:

> Looks like it's not possible to block it, because we can't obtain that is `style` binding

It looks like there is an property `BoundAttribute.keySpan.details` that contains the full string. Maybe we can find `BoundAttribute` that start with `style.*` value.
image


> Only possible do it via array of all css properties for forbid it

It appears that [eslint-plugin-css](https://github.com/ota-meshi/eslint-plugin-css) is [using](https://github.com/ota-meshi/eslint-plugin-css/blob/68da1eb550953427a10e497ec8ee44f85d931aa4/lib/rules/no-unknown-property.ts#L3) such list: [known-css-properties](https://github.com/known-css/known-css-properties).

pumano:
@jerone thank you for your findings, I will try to add it too.

pumano:
@jerone check please, I add support for `[style.cssproperty]` style binding too.

packages/eslint-plugin-template/tests/rules/no-inline-styles/cases.ts
```diff
@@ -0,0 +1,135 @@
...
+  '<input type="text">',
+  '<input type="image" alt="This is descriptive!">',
+  '<input type="image" aria-label="foo">',
+  '<input type="image" aria-labelledby="id1">',
+  {
+    code: `<input [ngStyle]="{ 'background-color': '#fff' }">`,
+    options: [{ allowNgStyle: true }],
+  },
+  {
+    code: `<input [style.background-color]="'#fff'">`,
```
jerone, line 27: Besides dash-case, I would also add an test for camelCase as mentioned [here](https://angular.io/guide/class-binding#binding-to-a-single-style).

```suggestion
    code: `<input [style.backgroundColor]="'#fff'">`,
```

pumano:
@jerone I apply your suggestion

pumano:
@JamesHenry all cases are implemented

packages/eslint-plugin-template/src/rules/no-inline-styles.ts
```diff
@@ -0,0 +1,129 @@
...
+export type MessageIds = 'noInlineStyles';
+export const RULE_NAME = 'no-inline-styles';
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Using html inline styles using style attribute is prohibited, use classes instead.',
```
JamesHenry, line 29: You have written this `description` as if it's a lint error, which is not what we want to do. The `description` is only used to describe why the rule exists, not give instructions/imply it is the "right" thing to do.

```suggestion
        'Disallows the use of inline styles in HTML templates',
```

JamesHenry:
@pumano as well as the above you'll please need to sync up with latest main

pumano:
@JamesHenry I apply your suggestion and rebase from main, but somehow without fast forward(

JamesHenry (approved):
Thanks a lot @pumano!

PR #1207, opened 2022-11-17
fix(no-input-rename): allow input aliases that match the directive na…

### LINKED ISSUE #978: [no-input-rename] False positive when directive apply on specific selector
eslint reports an error when I rename an `@Input` to match the directive name but the directive is applied only on specific element like `img[myDirective]`

```JSON
{
  "rules": {
    "@angular-eslint/no-input-rename": [true]
  }
}
```

```TS
@Directive({
  selector: 'img[appPicture]',
})
export class ImgPictureDirective {
  @Input('appPicture') picture: Picture; // <-- ESLint: Input bindings should not be aliased

  @HostBinding() get src(): string {
    return this.picture.src;
  }
}
```

The error disapears if I change my directive selector to only `[appPicture]`

**Versions**

| `@angular-eslint/eslint-plugin`    | `13.2.0` |
| `@typescript-eslint/parser`        | `13.2.0` |
| `ESLint`                           | `8.7.0` |

```sh
Angular CLI: 13.2.6
Node: 16.14.2
Package Manager: npm 8.5.0
OS: linux x64

Angular: 13.2.7

@angular-devkit/architect       0.1302.6
@angular-devkit/build-angular   13.2.6
@angular-devkit/core            13.2.6
@angular-devkit/schematics      13.2.6
@angular/cdk                    13.2.6
@angular/cli                    13.2.6
@angular/flex-layout            13.0.0-beta.38
@angular/material               13.2.6
@schematics/angular             13.2.6
ng-packagr                      13.3.0
rxjs                            7.4.0
typescript                      4.5.5
```

adbaran1:
…me applied to an element

This pull request enhances the `no-input-rename` rule to allow input aliases that match the directive name applied to an element.

Example:
```ts
@Directive({
  selector: 'img[fooDirective]'
})
class Test {
  @Input('fooDirective') foo: Foo;
}
```

Resolves #978 

**Note:** No new suggestions have been added for this enhancement because the rule already allows input aliases that match the directive name (when not applied to an element).

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-input-rename.md
+++ b/packages/eslint-plugin/docs/rules/no-input-rename.md
@@ -365,6 +365,38 @@ class Test {
 }
 ```
 
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/no-input-rename": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```ts
+@Directive({
+  selector: 'img[fooDirective]',
+})
+class Test {
+  @Input('notFooDirective') foo: Foo;
+         ~~~~~~~~~~~~~~~~~
+}
+```
+
 </details>
 
 <br>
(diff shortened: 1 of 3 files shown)


JamesHenry (approved):
Thanks a lot @adbaran1!

PR #1217, opened 2022-11-20
feat(eslint-plugin-template): [no-call-expression] add allowList option

### LINKED ISSUE #1055: [no-call-expression] Add support for including/excluding specific functions
I am trying to avoid using `form.get(...)` in my templates and force my team to use `form.controls.` in the HTML templates.
And I came across this rule - which is pretty awesome.

But I realized I get a lot of false positives - because some library maintainers recommend using their functions in the template:
https://ngneat.github.io/transloco/docs/translation-in-the-template#structural-directive

And they have said that the function is optimized and created to be used in templates efficiently (via memoization etc)

So, my request is - can we enable/disable certain calls based on the name of the function ?
Maybe a `includeNames`/`excludeNames` optional configuration for this rule ?

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `13.0.1` |
| `@angular-eslint/template-parser`         | `13.0.1` |
| `@typescript-eslint/parser`               | `5.29.0` |
| `ESLint`                                  | `8.18.0` |
| `node`                                    | `16.13.2` |

```sh

$ npx ng version

                |___/


Angular CLI: 14.0.0
Node: 16.13.2
Package Manager: npm 8.1.2
OS: linux x64

Angular: 14.0.0
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

@angular-devkit/architect       0.1400.0
@angular-devkit/build-angular   14.0.0
@angular-devkit/core            14.0.0
@angular-devkit/schematics      14.0.0
@schematics/angular             14.0.1
rxjs                            7.5.5
typescript                      4.7.3

```

- [ ] I have tried restarting my IDE and the issue persists.

sandikbarr:
Resolves #1055

Adds `allowList` option to the no-call-expression template rule to configure a list of function names allowed to be called.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/no-call-expression.ts
+++ b/packages/eslint-plugin-template/src/rules/no-call-expression.ts
@@ -1,10 +1,14 @@
-import type { Call } from '@angular-eslint/bundled-angular-compiler';
+import type { AST, Call } from '@angular-eslint/bundled-angular-compiler';
 import { TmplAstBoundEvent } from '@angular-eslint/bundled-angular-compiler';
 import { ensureTemplateParser } from '@angular-eslint/utils';
 import { createESLintRule } from '../utils/create-eslint-rule';
 import { getNearestNodeFrom } from '../utils/get-nearest-node-from';
 
-type Options = [];
+type Options = [
+  {
+    readonly allowList?: readonly string[];
+  },
+];
 export type MessageIds = 'noCallExpression';
 export const RULE_NAME = 'no-call-expression';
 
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/no-call-expression.ts
```diff
@@ -35,6 +51,15 @@ export default createESLintRule<Options, MessageIds>({
 
         if (isChildOfBoundEvent) return;
 
+        if (
```
adbaran1, line 54: Would it make sense to create a helper function for this logic, like seen in other rules? Something like `functionNameIsAllowed` could make this easier to understand. What do you think?

sandikbarr: 👍 good call, would keep this concise. I will update.

abaran30 (approved):

PR #1234, opened 2022-11-24
feat(eslint-plugin-template): [i18n] option to require i18n meaning

abaran30:
This pull request adds the option to `requireMeaning` when adding i18n metadata in templates.


JamesHenry (approved):
Thanks!

PR #1235, opened 2022-11-25
feat(eslint-plugin): [require-localize-metadata] option to require me…

abaran30:
…aning

This pull request adds the option to `requireMeaning` when using `$localize` to translate text in TypeScript.


JamesHenry (changes requested):
Otherwise LGTM

abaran30:

JamesHenry:
Thanks again, @abaran30!

P.S. Please don't force push to PRs which are already in review, as it causes the history to be nuked. In this case I had said I was otherwise happy with the PR, so I could have quickly taken a look at the diff between that point and now. As it is I had to look from scratch because the history was gone. We squash merge anyway so there is really no downside to having a slightly "unclean" history on the PR

PR #1248, opened 2022-12-07
fix(eslint-plugin): [no-inputs-metadata-property] do not report on di…

### LINKED ISSUE #1238: [no-inputs-metadata-property] Angular Host Directive inputs error
When using the new Angular 15 Directive Composition API and specifying inputs, angular-eslint reports an error.
 **Use `@Input` rather than the `inputs` metadata property (https://angular.io/styleguide#style-05-12) @angular-eslint/no-inputs-metadata-property**

Reproduction Repository https://github.com/Marcarrian/angular-15-eslint-issue

```TS
@Component({
  selector: 'app-my-component',
  template: `<span>MyComponent</span>`,
  hostDirectives: [{
    directive: MyDirective,
    inputs: ['myDirectiveSomeInput: someInput'], // errors here with @angular-eslint/no-inputs-metadata-property
  }],
  standalone: true,
})
export class MyComponent {

  @Input()
  public someInput = false;
}

@Directive({
  selector: '[appMyDirective]',
  standalone: true,
})
export class MyDirective {

  @Input()
  public myDirectiveSomeInput = false;
}
```

**Versions**

| `@angular-eslint/eslint-plugin`    | `15.1.0` |
| `@typescript-eslint/parser`        | `5.44.0` |
| `ESLint`                           | `8.28.0` |
| `node`                             | `16.16.0` |

```sh
Angular CLI: 15.0.1        
Node: 16.16.0              
Package Manager: npm 8.11.0
OS: win32 x64              

Angular: 15.0.1
... animations, cli, common, compiler, compiler-cli, core, forms

@angular-devkit/architect       0.1500.1
@angular-devkit/build-angular   15.0.1
@angular-devkit/core            15.0.1
@angular-devkit/schematics      15.0.1
@schematics/angular             15.0.1
rxjs                            7.5.7
typescript                      4.8.4
```

abaran30:
…rective composition API

Resolve #1238

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-inputs-metadata-property.md
+++ b/packages/eslint-plugin/docs/rules/no-inputs-metadata-property.md
@@ -489,6 +489,105 @@ class Test {}
 class Test {}
 ```
 
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/no-inputs-metadata-property": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```ts
+@Component({
+  selector: 'qx-menuitem',
+  hostDirectives: [{
+    directive: CdkMenuItem,
+    inputs: ['cdkMenuItemDisabled: disabled'],
+  }]
+})
+class Test {}
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/no-inputs-metadata-property": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```ts
+@Component({
+  selector: 'qx-menuitem',
+  'hostDirectives': [{
+    directive: CdkMenuItem,
+    inputs: ['cdkMenuItemDisabled: disabled'],
+  }]
+})
+class Test {}
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/no-inputs-metadata-property": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```ts
+@Component({
+  selector: 'qx-menuitem',
+  ['hostDirectives']: [{
+    directive: CdkMenuItem,
+    inputs: ['cdkMenuItemDisabled: disabled'],
+  }]
+})
+class Test {}
+```
+
 </details>
 
 <br>
--- a/packages/eslint-plugin/src/rules/no-inputs-metadata-property.ts
+++ b/packages/eslint-plugin/src/rules/no-inputs-metadata-property.ts
@@ -1,5 +1,6 @@
-import { Selectors } from '@angular-eslint/utils';
+import { ASTUtils, Selectors } from '@angular-eslint/utils';
 import type { TSESTree } from '@typescript-eslint/utils';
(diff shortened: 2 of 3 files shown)


JamesHenry (changes requested):
Please could you add a comment around why we are doing this in a similar way to how I did with no-input-rename?

packages/eslint-plugin/docs/rules/no-inputs-metadata-property.md
```diff
@@ -489,6 +489,105 @@ class Test {}
...
+<br>
+
+#### ✅ Valid Code
+
+```ts
+@Component({
+  selector: 'qx-menuitem',
+  hostDirectives: [{
+    directive: CdkMenuItem,
+    inputs: ['cdkMenuItemDisabled: disabled'],
```
Marcarrian, line 519: nit: remove the input renaming as it is not relevant for this rule.

packages/eslint-plugin/src/rules/no-inputs-metadata-property.ts
```diff
@@ -29,6 +30,26 @@ export default createESLintRule<Options, MessageIds>({
...
         node: TSESTree.Property,
       ) {
+        const ancestorMayBeHostDirectiveAPI = node.parent?.parent?.parent;
+
+        if (
+          ancestorMayBeHostDirectiveAPI &&
+          ASTUtils.isProperty(ancestorMayBeHostDirectiveAPI)
+        ) {
+          const hostDirectiveAPIPropertyName = 'hostDirectives';
+
```
Marcarrian, line 40: a short comment similar to https://github.com/angular-eslint/angular-eslint/pull/1231/files#diff-34e0047dec14410dccd7fee53b69b0edeb6c62c8b6d4480a4d6144e95aa323b5 might be useful.

packages/eslint-plugin/src/rules/no-inputs-metadata-property.ts
```diff
@@ -29,6 +30,26 @@ export default createESLintRule<Options, MessageIds>({
       } ${Selectors.metadataProperty(METADATA_PROPERTY_NAME)}`](
         node: TSESTree.Property,
       ) {
+        const ancestorMayBeHostDirectiveAPI = node.parent?.parent?.parent;
```
Marcarrian, line 33: question: I see a difference between the implementation here https://github.com/angular-eslint/angular-eslint/pull/1231/files#diff-34e0047dec14410dccd7fee53b69b0edeb6c62c8b6d4480a4d6144e95aa323b5. What's the reason for it?

Marcarrian (commented):
Looks good to me! Some nitpicky comments and questions.
Oh and I just saw JamesHenry's comment after leaving mine.

packages/eslint-plugin/src/rules/no-inputs-metadata-property.ts
abaran30: Only a slight difference 😄 . So in `no-input-rename`, the targeted selector is `Selectors.INPUTS_METADATA_PROPERTY_LITERAL`, while the targeted selector for this rule is `${
        Selectors.COMPONENT_OR_DIRECTIVE_CLASS_DECORATOR
      } ${Selectors.metadataProperty(METADATA_PROPERTY_NAME)}`. Although either of the two selectors *could* work for this rule, the main reason why I kept the latter selector is because when we do report a lint violation, we want to highlight the `inputs` keyword like shown in the invalid test cases. When using the former selector, we'd end up highlighting the value bound to `inputs`, which I think is less helpful.

abaran30:

JamesHenry (approved):

PR #1250, opened 2022-12-07
fix(eslint-plugin): [component-selector] enhance check for prefix and…

### LINKED ISSUE #318: kebab-case option warns of single words in component-selector
While creating the docs for the rule [component-selector](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/rules/component-selector.ts), I found an error message different from what I expected

#### Expected message

`The selector should be prefixed by one of the prefixes: 'app,toh' (https://angular.io/guide/styleguide#style-02-07) eslint@angular-eslint/component-selector`

#### Received message

`The selector should be kebab-case (https://angular.io/guide/styleguide#style-05-02) eslint@angular-eslint/component-selector`

#### Configuration
```json
"rules": {
    "@angular-eslint/component-selector": [
        "warn",
        {
            "type": "element",
             "style": "kebab-case",
             "prefix": ["app", "toh"]
        }
    ]
}
```

#### Code
```ts
@Component({
  selector: 'root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {}
```

Digging through the code, I found this to be a [deliberate decision](https://github.com/angular-eslint/angular-eslint/blob/c44cd5d043a3d203efd1faaed4cbfee2c9ac2e9d/packages/eslint-plugin/tests/rules/component-selector.test.ts#L285-L296). Why is it? I mean, a single word like `app` is a valid kebab-cased word (from what I known).

abaran30:
...kebab-case

Resolves #318 

This pull request contains the following changes to accommodate the discussion had in #318:
- Add new error message in rule `component-selector` for invalid prefix ***and*** case
- Add test cases to verify the examples posted in the issue

### DIFF
--- a/packages/eslint-plugin/src/rules/pipe-prefix.ts
+++ b/packages/eslint-plugin/src/rules/pipe-prefix.ts
@@ -70,8 +70,8 @@ export default createESLintRule<Options, MessageIds>({
         const allowPrefixesExpression = prefixes.join('|');
         const prefixValidator = SelectorUtils.SelectorValidator.prefix(
           allowPrefixesExpression,
-          'camelCase',
         );
+        const styleValidator = SelectorUtils.SelectorValidator.camelCase;
 
         let nameValue;
 
@@ -88,7 +88,7 @@ export default createESLintRule<Options, MessageIds>({
           return;
         }
 
-        if (!prefixValidator.apply(this, [nameValue])) {
+        if (!prefixValidator.apply(this, [nameValue]) || !styleValidator) {
           context.report({
             node: nameSelector,
             messageId: 'pipePrefix',
--- a/packages/eslint-plugin/src/rules/pipe-prefix.ts
+++ b/packages/eslint-plugin/src/rules/pipe-prefix.ts
@@ -70,8 +70,8 @@ export default createESLintRule<Options, MessageIds>({
         const allowPrefixesExpression = prefixes.join('|');
         const prefixValidator = SelectorUtils.SelectorValidator.prefix(
           allowPrefixesExpression,
+          'camelCase',
         );
-        const styleValidator = SelectorUtils.SelectorValidator.camelCase;
 
         let nameValue;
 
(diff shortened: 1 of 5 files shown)


abaran30:
TODO: this change allows for selectors like `sgggggggg-foo-bar`; not ideal... Thank you @sandikbarr for catching this!

packages/eslint-plugin/src/rules/pipe-prefix.ts
```diff
@@ -88,7 +88,7 @@ export default createESLintRule<Options, MessageIds>({
           return;
         }
 
-        if (!prefixValidator.apply(this, [nameValue])) {
+        if (!prefixValidator.apply(this, [nameValue]) || !styleValidator) {
```
sandikbarr, line 91: Does this styleValidator need to be called as a function with the nameValue?

packages/eslint-plugin/tests/rules/component-selector/cases.ts
```diff
@@ -302,9 +312,9 @@ export const invalid = [
         })
         class Test {}
       `,
-    messageId: messageIdStyleFailure,
+    messageId: messageIdStyleAndPrefixFailure,
```
abaran30, line 315: Note: the reported error for this invalid case has changed to the new error message. Although it's not incorrect, the former is more concise for this particular case. I think the only way to retain the original message is to move the style logic inside the prefix validator to its own validator (which I did attempt in a previous commit). However, this will result in a more complex conditional in the rule when determining which message to report.

LMK your thoughts!

sandikbarr: This seems acceptable to me as a non breaking change while addressing the original issue.

sandikbarr (approved):
thanks for digging into this!

abaran30:

JamesHenry (approved):
Thank you @abaran30 (and @sandikbarr)! And sorry for the delay in reviewing

Issue #898, opened 2022-02-12
RuleTester should support valid/invalid test cases with `filename`


ramarivera:
**Description and reproduction of the issue**

When trying to create custom rules and test them, if the rules depend on calling `context.getFilename` or `context.getPhysicalFilename` it will always return `<input`> inside the rule.

From what I can see, this is due to `RuleTester` overriding `filename` when parsing the tests [here](https://github.com/angular-eslint/angular-eslint/blob/2000c0c63f7b2c4d67c387cce4be8d039885da40/packages/utils/src/rules-tester.ts#L71).

ramarivera:
Any clue, pointer, tips on how to solve this? 
Should I try to raise a PR on my own?

JamesHenry:
@ramarivera please could you open a draft PR against this repo which shows what you are trying to accomplish and what is currently not possible?

I think by using this repo as a base for the repro it will allow me to easily see what you are referring to and will give you all the necessary setup you need.

Simply create a new fake rule which minimally repros what you are trying to achieve in your custom rule along with the unit tests

ramarivera:
Hey, sorry but I don't have access to the code where I was manually patching the RuleTester.

But this type of tests should reproduce the issue IIRC https://github.com/dolsem/eslint-plugin-filename-rules/blob/master/test/lib/match.test.js#L14

The idea is that there are eslint rules which you can define to check for specific file names or similar. I had wrote a rule which would only run on `*.spec.ts` files for example

JamesHenry:
@ramarivera My request is still the same please => https://github.com/angular-eslint/angular-eslint/issues/898#issuecomment-1318594176

The best way to proceed is to try and reproduce against this repo. This seems to be an issue you have knowledge of and care about, so you are in the best position to address it. I will happily review a PR from you on this once you establish a failing case!

ramarivera:
> @ramarivera My request is still the same please => https://github.com/angular-eslint/angular-eslint/issues/898#issuecomment-1318594176
> 
> The best way to proceed is to try and reproduce against this repo. This seems to be an issue you have knowledge of and care about, so you are in the best position to address it. I will happily review a PR from you on this once you establish a failing case!

Hey @JamesHenry I'm unfortunately not working anymore with angular so my priority for this has become lower (plus, I remember I had an ugly workaround).
I will try to get back to this later this week, but please understand I might not get to it

Thanks for your time

JamesHenry:
[...]

PR #1259, opened 2022-12-13
fix(utils): use test case filename when specified

nathan-xiao1:
Addresses #898
- For valid and invalid test cases, use the test case's filename if specified, otherwise, use `this.filename`


JamesHenry:
Thanks @nathan-xiao1!

Issue #1232, opened 2022-11-24
Improving/replacing `sort-ngmodule-metadata-arrays`


JamesHenry:
In light of the number of issues reported with this rule I wanted to bring them together in one place in order to inform a greatly improved implementation of sorting `imports` metadata.

- Now that standalone components are stable, the rule should allow for sorting imports in `Component` metadata as well as `NgModule`s
  - This is an important point and almost certainly means that **a new rule should be created to replace `sort-ngmodule-metadata-arrays`** which should be deprecated in favour of it*
(Originally reported here: https://github.com/angular-eslint/angular-eslint/issues/1133)

- Sorting of `imports` does not work when some have `.forRoot()` (Originally reported here: https://github.com/angular-eslint/angular-eslint/issues/927)

- The rule implementation needs to be completely refactored in order to allow for reliable auto-fixing. I won't cover this in detail as this is done on #675 but essentially the way ESLint auto-fixing works means we cannot do what we do for other rules and simply apply a fix to an item we detect as being in the wrong place. There could easily be more than 10 items in the wrong place, but ESLint would give up at this point.
(Originally reported here: https://github.com/angular-eslint/angular-eslint/issues/675)

In light of this last point, the new rule instead needs to focus on extracting the relevant chunks of a file manually (not using granular AST selectors) and fixing the final form in one action.

This is the approach taken by this import sorting plugin for example: https://github.com/lydell/eslint-plugin-simple-import-sort/blob/main/src/imports.js

---

*The new rule will be called `@angular-eslint/sort-imports-metadata`

abaran30:
FYI @JamesHenry I started looking into solutions for this issue. I have a draft PR up [here](https://github.com/angular-eslint/angular-eslint/pull/1303), though it only contains boilerplate code for the new rule at the moment. Will continue to play around with it when I can.

Nosfistis:
Any news on this rule?

abaran30:
I closed my draft PR (can always refer back to it whenever necessary) that attempted to resolve this issue because we discovered that there are cases where the order of imports matters.

One case (as I commented in the draft PR) is when simulating a data server using the [In-memory Web API](https://github.com/angular/angular/tree/main/packages/misc/angular-in-memory-web-api). The docs explicitly state, "Always import the `HttpClientInMemoryWebApiModule` *after* the `HttpClientModule` to ensure that the in-memory backend provider supersedes the Angular version." Sorting the imports would break this.

[...]

PR #1303, opened 2023-02-10
feat(eslint-plugin): new rule sort-imports-metadata

abaran30:
This pull request introduces the new rule `@angular-eslint/sort-imports-metadata` as described in #1232.


packages/eslint-plugin/src/rules/sort-imports-metadata.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+  defaultOptions: [
+    {
+      locale: DEFAULT_LOCALE,
+    },
+  ],
+  create(context /* [{ locale }] */) {
+    const selectors = [
+      `${Selectors.COMPONENT_CLASS_DECORATOR} Property[key.name!="deps"] > ArrayExpression`,
+      `${Selectors.MODULE_CLASS_DECORATOR} Property[key.name!="deps"] > ArrayExpression`,
```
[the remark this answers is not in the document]
abaran30: Done! Also made it to check for the `imports` key only.

abaran30:
Update: I added a first version implementation of the new rule and some basic tests for sanity. Perhaps the logic could be more efficient. @rubiesonthesky and @JamesHenry please take a peek when you can and LMK what you think!

P.S. I'll look to add applicable tests from `sort-ngmodule-metadata-arrays` to ensure those are working as expected.

packages/eslint-plugin/tests/rules/sort-imports-metadata/cases.ts
```diff
@@ -0,0 +1,146 @@
...
+      ],
+    })
+    export class TestComponent { }
+    `,
+    messageId,
+    annotatedOutput: `
+    @Component({
+      standalone: true,
+      selector: 'test-component',
+      imports: [aModule, bModule, cModule, dModule],
```
abaran30, line 141: How exactly do we want to handle imports listed per line?

abaran30:
Update on how comments are handled: currently, they get stripped away and we're left with the import items only. It is possible to extract the comments and re-insert them where they belong, but it will require a bit of logic to do so.

Will need to spend some time playing around with it.

packages/eslint-plugin/tests/rules/sort-imports-metadata/cases.ts
```diff
@@ -0,0 +1,183 @@
...
+    })
+    export class TestComponent { }
+    `,
+    messageId,
+    annotatedOutput: `
+    @Component({
+      standalone: true,
+      selector: 'test-component',
+      imports: [
+        // Line comment above aModule
```
abaran30, line 162: I still don't fully understand if/when an `annotatedOutput` also requires the `~~~` annotations found in `annotatedSource`.

packages/eslint-plugin/tests/rules/sort-imports-metadata/cases.ts
```diff
@@ -0,0 +1,183 @@
...
+      cModule,
+      dModule
+    ],
+  })
+  export class TestComponent { }
+  `,
+];
+
+export const invalid = [
+  // TODO: This test fails for two reason:
```
abaran30, line 47: This `TODO` starts creeping into formatting territory, which we should limit/avoid.

packages/eslint-plugin/tests/rules/sort-imports-metadata/cases.ts
```diff
@@ -0,0 +1,183 @@
...
+  // -         cModule,
+  // -         dModule,
+  // +                 aModule,
+  // +                 bModule,
+  // +                 cModule,
+  // +                 dModule,
+  //
+  // Second error is that the fix receives additional characters at the end of the array:
+  //   dModule,
+  // ],·················
```
abaran30, line 60: Still trying to figure out how these characters come about.

packages/eslint-plugin/src/rules/sort-imports-metadata.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        // Sort the imports
+        importEntriesWithComments.sort((importA, importB) =>
+          importA.importName.localeCompare(importB.importName, locale),
+        );
+
+        const sortedImportsWithCommentsString = `[\n${getSortedImportsWithCommentsAsString(
+          importEntriesWithComments,
+          whitespacePrefix,
+        )}${' '.repeat(nodeForFixer.parent?.loc.start.column as number)}]`;
```
abaran30, line 144: This is a bit too magical and in formatting territory...

abaran30:
@JamesHenry updated with the following actions:
- Rebased with latest main
- Pulled some of the auto-fixer logic into a function (passing the `context` instead of `sourceCode`)
- Added applicable tests from existing rule `sort-ngmodule-metadata-arrays`

Ready for review.

P.S. if we still want to test this with a real/large codebase, LMK how we can create a separate release containing this rule to test (I would have to find or create a codebase to work with).

abaran30:
TODO: This rule currently doesn't handle `CallExpression` imports, e.g. `RouterModule.forChild(...)`.

abaran30:
Thank you for sharing @rubiesonthesky! This will work for the check if the imports are sorted. However, adapting the auto-fixer to support this, as it's currently implemented, will get a bit tricky since it essentially builds the new imports list as a string. Will have to play around with it.

VictorienTardif:
Hello, is this PR dropped?
Is there any alternative?
Thx

joergbaier:
It seems a bit dogmatic to exclude rules that improve the consistency and readability of Angular projects. Angular is already a heavily opinionated framework. A rule like this considers the context of Angular code and applies a standard to it. Many new and existing Angular projects will use this package by default, why not provide them with consistency?

These rules fall in the same category:

- `@angular-eslint/template/attributes-order`
- https://github.com/angular-eslint/angular-eslint/pull/1320

Should another tool (prettier plugin) be doing these instead? Applying an order doesn't really "belong" to formatting tools either.

abaran30:
Apologies for the lack of activity here... So we recently discovered that there's at least one case where the order of imports matters for functionality, and that is when simulating a data server using the [In-memory Web API](https://github.com/angular/angular/tree/main/packages/misc/angular-in-memory-web-api). The docs explicitly state, "Always import the `HttpClientInMemoryWebApiModule` *after* the `HttpClientModule` to ensure that the in-memory backend provider supersedes the Angular version." Sorting the imports would break this.

We'll see if there are any other cases like these (feel free to share if you know of any). However, in my opinion as of now, we should not have a rule that would have to dance around certain exceptions like these in order to work for most/all users.

abaran30:
Closing this PR, will add a comment to the issue (#1232).

PR #1317, opened 2023-02-24
feat: add prefer-standalone-component rule

thomasmarcelin:
This rule should suggest using standalone components in Angular.

I want to test the rule locally but I can't figure out how to do this, any help will be much appreciated. Thanks :)


Friendseeker:
Took me a while to figure out how to test the rule locally!

Screenshot 2023-03-05 at 5 24 36 PM

In Webstorm the above run config would work. (Test file is the corresponding `spec.ts`, Suite name is name of the rule.,

thomasmarcelin:
> Took me a while to figure out how to run unit test locally too!
> 
> Screenshot 2023-03-05 at 5 24 36 PM
> 
> In Webstorm the above run config would work to run an individual unit test. (Test file is the corresponding `spec.ts`, Suite name is name of the rule). Before you run the individual test, you may want to execute `yarn test` to to rebuild the project).
> 
> If you don't want to use Webstorm, use `yarn test` to run all test at once should also suffices. It runs reasonably fast, and it shows the log for each individual test.

I didn't have any issue running the unit test, i meant testing my angular-eslint rule against a real project with something like `npm link`

But thank you for the tip anyway :)

packages/eslint-plugin/src/rules/prefer-standalone-component.ts
```diff
@@ -0,0 +1,89 @@
...
+type Options = [];
+export type MessageIds = 'preferStandaloneComponent' | 'suggestAddStandalone';
+export const RULE_NAME = 'prefer-standalone-component';
+const METADATA_PROPERTY_NAME = 'standalone';
+const IS_STANDALONE = 'true';
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
```
JamesHenry, line 19: I don't think this should be a suggestion, it should just be an autofix.

Suggestions are intended for cases where there is potentially ambiguity around the resolution to an issue, and the user needs to factor in the context of their specific code - the lint rule can't know in 100% of cases what the right thing to do is.

Here, however, the resolution is very clear and unambiguous - the standalone property needs to be added.

If you want to have a mixture of standalone and non-standalone components you simply wouldn't want to use this rule.

packages/eslint-plugin/tests/rules/prefer-standalone-component/cases.ts
```diff
@@ -0,0 +1,57 @@
...
+export const valid = [
+  `
+    @Component({
+      standalone: true,
+    })
+    class Test {}
+  `,
+];
+
+export const invalid = [
```
JamesHenry, line 16: Please could you add some richer, more real-world Component metadata examples, just for completeness (it would also make it easier to see how the fixing affects Component metadata which already has multiple lines)

JamesHenry (changes requested):
Thanks a lot @thomasmarcelin! Please the comments inline

packages/eslint-plugin/src/rules/prefer-standalone-component.ts
thomasmarcelin: I have changed the meta type to `problem` instead of `suggestion` to avoid ambiguous rule usage (I could not find the `autofix` you are mentioning, is it something I need to set in the rule meta?)

packages/eslint-plugin/tests/rules/prefer-standalone-component/cases.ts
thomasmarcelin: I have added several valid & invalid tests cases with `selector`, `template` and other properties to also test multi-line compatibility

packages/eslint-plugin/tests/rules/prefer-standalone-component/cases.ts
```diff
@@ -0,0 +1,165 @@
...
+        class Test {}
+`,
+    messageId,
+    suggestions: [
+      {
+        messageId: suggestAddStandalone,
+        output: `
+        @Component({
+        
+        standalone: true,selector: 'my-selector',
```
thomasmarcelin, line 154: Is the rule supposed to also ensure the code is indented ? I assumed it was `prettier` or another code formatter job

thomasmarcelin:
> Thanks a lot @thomasmarcelin! Please see the comments inline

Thanks for the review, I have updated the PR 👍

packages/eslint-plugin/src/rules/prefer-standalone-component.ts
JamesHenry: I’m sorry for the confusion - suggestion is actually correct here. 

This comment should have been against the hasSuggestions: true bit - it should instead use fixable: ‘code’ to enable auto fixing

sorry on mobile so not easy to format the updates there 

please revert back to suggestion as the meta type and sorry again for the confusion

thomasmarcelin: No problem, I have reverted the type to `suggestion` and changed the `hasSuggestions: true` to `fixable: code` instead. I'm relatively new to the eslint rules, apologies for the confusion 😄

packages/eslint-plugin/tests/rules/prefer-standalone-component/cases.ts
JamesHenry: I agree!

packages/eslint-plugin/src/rules/prefer-standalone-component.ts
```diff
@@ -0,0 +1,89 @@
...
+          ASTUtils.isLiteral(standalone) &&
+          standalone.value === true
+        ) {
+          return;
+        }
+
+        context.report({
+          node: nodeToReport(node),
+          messageId: 'preferStandaloneComponent',
+          suggest: [
```
JamesHenry, line 51: This should not be suggest.

Please check out https://eslint.org/docs/latest/extend/custom-rules and other rules that have autofixing, there is a lot of precedent in the codebase

packages/eslint-plugin/src/rules/prefer-standalone-component.ts
```diff
@@ -0,0 +1,89 @@
...
+    type: 'suggestion',
+    docs: {
+      description: `Ensures component \`${METADATA_PROPERTY_NAME}\` property is set to \`${IS_STANDALONE}\` in the component decorator`,
+      recommended: false,
+    },
+    fixable: 'code',
+    schema: [],
+    messages: {
+      preferStandaloneComponent: `The component \`${METADATA_PROPERTY_NAME}\` property should be set to \`${IS_STANDALONE}\``,
+      suggestAddStandalone: `Add \`${METADATA_PROPERTY_NAME}: ${IS_STANDALONE}\` in your component decorator`,
```
JamesHenry, line 28: Please remove the word suggest to avoid confusions with rules which make suggestions

packages/eslint-plugin/tests/rules/prefer-standalone-component/cases.ts
```diff
@@ -0,0 +1,165 @@
...
+  convertAnnotatedSourceToFailureCase({
+    description:
+      'should fail when a component does not have the standalone property in the decorator',
+    annotatedSource: `
+        @Component({})
+        ~~~~~~~~~~~~~~
+        class Test {}
+      `,
+    messageId,
+    suggestions: [
```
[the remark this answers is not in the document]
thomasmarcelin: Thanks, I have changed the test cases accordingly

abaran30 (approved):

JamesHenry:
Thanks a lot for this @thomasmarcelin! I will do a final review tomorrow

JamesHenry (approved):

JamesHenry:
Thanks again!

thomasmarcelin:
> Thanks again!

Thank you for your time 👋

Issue #995, opened 2022-05-12
[sort-lifecycle-methods] Sort Angular Lifecycle Methods


lengauermario:
It would be nice, if the lifecycle methods could be sorted after constructor and before other methods.

Actual they threated as normal method: 
```
constructor() {}

ngAfterContentChecked(): void {}

ngAfterContentInit(): void {}

ngAfterViewChecked(): void {}

ngAfterViewInit(): void {}

ngDoCheck(): void {}

ngOnChanges(changes: SimpleChanges): void {}

public deleteUsers(): void {}

public loadUsers(): void {}

ngOnDestroy(): void {}

ngOnInit(): void {}
```

Expected behavior of lint:

```
constructor() {}

ngOnChanges(changes: SimpleChanges): void {}

ngOnInit(): void {}

ngDoCheck(): void {}

ngAfterContentInit(): void {}

ngAfterContentChecked(): void {}

ngAfterViewInit(): void {}

ngAfterViewChecked(): void {}

ngOnDestroy(): void {}

public deleteUsers(): void {}

public loadUsers(): void {}
```

Samuel-Beslogic:
Sorting them by the same order they are called in would be awesome.

sc0tt5:
I submitted a few updates, requested review and draft removal.

Samuel-Therrien-Beslogic:
Is this closed by https://github.com/angular-eslint/angular-eslint/pull/1320 or is there more to do ?

PR #1320, opened 2023-03-06
feat(eslint-plugin): [sort-lifecycle-methods] add rule

Friendseeker:
Implements feature requested by https://github.com/angular-eslint/angular-eslint/issues/995


Friendseeker:
TODO: `@Input() set` methods should be allowed to be declared before lifecycle methods.

fandy:
Any updates on this? ❤️

Friendseeker:
> Any updates on this? ❤️ 

Sorry. I have been especially busy. I will work on it if I have spare time (If I... will have any spare time).

packages/utils/src/eslint-plugin/ast-utils.ts
```diff
@@ -91,6 +91,16 @@ export const angularLifecycleInterfaceKeys = objectKeys(
 );
 export const angularLifecycleMethodKeys = objectKeys(AngularLifecycleMethods);
```
abaran30, line 93: Would you be able to include a comment containing the URL to the lifecycle event sequence?

https://angular.io/guide/lifecycle-hooks#lifecycle-event-sequence

packages/eslint-plugin/src/rules/sort-lifecycle-methods.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        for (let i = 1; i < declaredMethods.length; ++i) {
+          const previous = declaredMethods[i - 1];
+          const current = declaredMethods[i];
+          if (!isLifecycleMethod(previous) && isLifecycleMethod(current)) {
+            context.report({
+              node: previous.key,
+              messageId: 'nonLifecycleMethodBeforeLifecycleMethod',
+            });
+          }
```
jerone, line 66: You could limit to one `for` loop by adding an extra condition here.

Pseudo-code:
```
let previousLifecycleMethod = null; // Declare outside loop.
if (isLifecycleMethod(current) && previousLifecycleMethod != null) {
  if (isBefore(current, previousLifecycleMethod)) {
    context.report({
       node: current.key,
       messageId: 'lifecycleMethodsNotSorted',
    });
  }
  previousLifecycleMethod = current;
}
```

JamesHenry:
Converted to draft to more accurately reflect its current state

sc0tt5:
@Friendseeker @JamesHenry I would like to contribute to this PR if possible. Please advise.

Friendseeker:
> @Friendseeker @JamesHenry I would like to contribute to this PR if possible. Please advise.

Hi @sc0tt5! Sorry that it took me so long but I finally regained some spare time. Would definitely appreciate you to work on this PR to wrap it up!

First off, would you say my below comment is reflective on what this feature should be? When I first worked on this feature, I did not consider all the cases, and I am still not sure if all cases are considered in the comment below.

> TODO:
> 
> * `@Input() set` methods should be allowed to be declared before lifecycle methods.
> * Constructor should be before lifecycle methods
> 
> Ordering should be:
> 
> * `@Inputs()`
> * Constructor
> * Lifecycle methods
> * Non-Lifecycle methods

sc0tt5:
> 

Hello @Friendseeker! No worries. I was delighted to find someone had taken the initiative to work this in. I will double check the ordering and anything we might be missing, then reconfirm with you soon. Give me until later today/tonight. 

Then once we're on the same page, I'll send over my contribution sometime soon. How should I contribute? I assume PR into your branch?

Friendseeker:
> Hello @Friendseeker! No worries. I was delighted to find someone had taken the initiative to work this in. I will double check the ordering and anything we might be missing, then reconfirm with you soon. Give me until later today/tonight.
> 
> Then once we're on the same page, I'll send over my contribution sometime soon. How should I contribute? I assume PR into your branch?

Really appreciate, @sc0tt5 ! Feel free to take as much time as you need! Better to consider all cases than rushing it and having to redo things afterwards.

PR into my branch sounds great!

sc0tt5:
> 

@JamesHenry  please consider removing the draft status and review the recent updates that I submitted via @Friendseeker PR.

packages/eslint-plugin/src/rules/sort-lifecycle-methods.ts
sc0tt5: @jerone I consider this. However the first loop considers all methods, checks the constructor placement and non-lifecycle methods. The second loop targets only the lifecycle methods and checks sequence. I added comments for clarity.

packages/utils/src/eslint-plugin/ast-utils.ts
sc0tt5: Done. See below.

packages/eslint-plugin/tests/rules/sort-lifecycle-methods/cases.ts
```diff
@@ -0,0 +1,190 @@
...
+        ngAfterViewChecked(): void {}
+        ngOnDestroy(): void {}
+        ~~~~~~~~~~~
+        constructor() {}
+        doSomethingElse(): void {}
+      }
+      `,
+    messageId: lifecycleMethodBeforeConstructor,
+  }),
+];
```
abaran30, line 190: What's the expected behavior if more than one lifecycle hook is in the incorrect order?

abaran30:
> > @Friendseeker @JamesHenry I would like to contribute to this PR if possible. Please advise.
> 
> Hi @sc0tt5! Sorry that it took me so long but I finally regained some spare time during weekends (for which I can use to wrap up this PR). Would definitely appreciate you to work on this PR to wrap it up!
> 
> First off, would you say my below comment is reflective on what this feature should be? When I first worked on this feature, I did not consider all the cases, and I am still not sure if all cases are considered in the comment below.
> 
> > TODO:
> > 
> > * `@Input() set` methods should be allowed to be declared before lifecycle methods.
> > * Constructor should be before lifecycle methods
> > 
> > Ordering should be:
> > 
> > * `@Inputs()`
> > * Constructor
> > * Lifecycle methods
> > * Non-Lifecycle methods
> 
> Also, if order is incorrect, which methods should be flagged? Should the entire class be flagged inside? I still don't have a good idea on that problem, and would really appreciate your feedback.

I'm curious why there's a consideration for where the Lifecycle method implementations are, e.g. before or after the constructor and other methods. IMO, I'd expect this rule to only care about the sort order of the Lifecycle methods and not where they are implemented, even though I can imagine most developers preferring some order like suggested in the quoted comment. Perhaps I'm missing something from the discussions or established expectations?

Friendseeker:
> > > @Friendseeker @JamesHenry I would like to contribute to this PR if possible. Please advise.
> > 
> > 
> > Hi @sc0tt5! Sorry that it took me so long but I finally regained some spare time during weekends (for which I can use to wrap up this PR). Would definitely appreciate you to work on this PR to wrap it up!
> > First off, would you say my below comment is reflective on what this feature should be? When I first worked on this feature, I did not consider all the cases, and I am still not sure if all cases are considered in the comment below.
> > > TODO:
> > > 
> > > * `@Input() set` methods should be allowed to be declared before lifecycle methods.
> > > * Constructor should be before lifecycle methods
> > > 
> > > Ordering should be:
> > > 
> > > * `@Inputs()`
> > > * Constructor
> > > * Lifecycle methods
> > > * Non-Lifecycle methods
> > 
> > 
> > Also, if order is incorrect, which methods should be flagged? Should the entire class be flagged inside? I still don't have a good idea on that problem, and would really appreciate your feedback.
> 
> I'm curious why there's a consideration for where the Lifecycle method implementations are, e.g. before or after the constructor and other methods. IMO, I'd expect this rule to only care about the sort order of the Lifecycle methods and not where they are implemented, even though I can imagine most developers preferring some order like suggested in the quoted comment. Perhaps I'm missing something from the discussions or established expectations?

I did some thinking and I personally think you are right. I overcomplicated the problem. The rule should have one single responsibility only, being the lifecycles methods themselves are in order. Other method's ordering do not matter.

@sc0tt5 What do you think?

sc0tt5:
@Friendseeker yes, sounds good to me. @abaran30 I'm happy to revert changes referencing the constructor, or wait on updates from you before making any changes. Please let me know.

packages/eslint-plugin/tests/rules/sort-lifecycle-methods/cases.ts
Friendseeker: @sc0tt5 Should we flag all inversions? Or can we find a more optimal & deterministic way to flag methods?

abaran30:
@sc0tt5 let's make the rule focus on the sorting of lifecycle methods only. 🙂

packages/eslint-plugin/tests/rules/sort-lifecycle-methods/cases.ts
abaran30: I'm assuming that with the following example...
```ts
        ngOnInit(): void {}
        ngOnChanges(): void {}
        ngAfterContentInit() {}
        ngDoCheck() {}
```
...this rule will report twice: first on `ngOnChanges` and then on `ngDoCheck`. If my assumption is correct, I think this approach is appropriate.

Friendseeker: Just want to clarify, in the example

```javascript
        ngOnInit() {}
        ngOnChanges() {}
        ngAfterContentInit() {}
        ngDoCheck() {}
```

There's two inversion pairs `(ngOnInit, ngOnChanges)` and `(ngAfterContentInit, ngDoCheck)`. The agreed rule is for each inversion pair, we flag the second element in the pair.

abaran30: Thank you for clarifying. Makes sense to me 🙂.

One thing to keep in mind with this approach and if there will be an auto-fixer for this rule in the future... The ESLint fix "process will repeat up to 10 times, or until no more fixable problems are found," as documented [here](https://eslint.org/docs/latest/extend/custom-rules#applying-fixes).

Should be okay for this rule because there are only 8 Angular lifecycle events, at least for now.

bevrard:
The proposal with sort and position was just what I was looking for ;-) How could we do it if it's not included with this rule?

packages/eslint-plugin/tests/rules/sort-lifecycle-methods/cases.ts
abaran30: If possible, maybe let's play around with test cases that check for multiple errors? See [here](https://github.com/angular-eslint/angular-eslint/blob/main/packages/utils/src/convert-annotated-source-to-failure-case.ts#L77)

Friendseeker:
> The proposal with sort and position was just what I was looking for ;-) How could we do it if it's not included with this rule?

I say likely via a combination of multiple rules.

jerone:
> > The proposal with sort and position was just what I was looking for ;-) How could we do it if it's not included with this rule?
> 
> Likely via a combination of multiple rules.
> 

Sounds more like an overload for [`@typescript-eslint/member-ordering`](https://typescript-eslint.io/rules/member-ordering).

packages/eslint-plugin/tests/rules/sort-lifecycle-methods/cases.ts
Friendseeker: @sc0tt5 What is your opinion on the inversion based flagging?

sc0tt5:
I like where we landed with this, only focused on sort order of the lifecycle methods. It's concise and serves it's purpose within the Angular scope of things. I did test with [@typescript-eslint/member-ordering](https://typescript-eslint.io/rules/member-ordering) and as @bevrard mentioned it would be nice to somehow have these positioned, which is out of scope here. It doesn't seem possible with [@typescript-eslint/member-ordering](https://typescript-eslint.io/rules/member-ordering). I will probably try a custom rule in my project.

JamesHenry (approved):
Thanks everyone for their hard work and consideration on this one! I think we can ship it

PR #1322, opened 2023-03-10
feat(eslint-plugin-template): [self-closing-tags] add rule

aleesaan:
Implements the feature proposed in [this discussion](https://github.com/angular-eslint/angular-eslint/discussions/1318).

Let me know If there's a better way to check that an element has no content, and thanks for the awesome plugin!


packages/utils/src/eslint-plugin-template/native-elements.ts
JamesHenry, line 1: Please use the existing `getDomElements()` utility that we have and do not introduce a new list for us to maintain separately.

You can have a look in https://github.com/angular-eslint/angular-eslint/blob/6abcc3ab0d3000d00ab668d2f6776e0f851b0a41/packages/eslint-plugin-template/src/rules/accessibility-role-has-required-aria.ts for precedent on how to select native elements (and here you want to do the inverse, and only select non-native)

packages/eslint-plugin-template/tests/rules/self-closing-tags/cases.ts
```diff
@@ -0,0 +1,63 @@
...
+export const valid = [
+  '<my-component type="text" [name]="foo">With some content</my-component>',
+  '<my-component />',
+  `
+    <my-component
+      type="text"
+      [name]="foo"
+      [items]="items" />
+  `,
+  '<img />',
```
JamesHenry, line 13: Please add a non self-closing native element example

packages/eslint-plugin-template/src/rules/self-closing-tags.ts
```diff
@@ -0,0 +1,74 @@
...
+      [MESSAGE_ID]:
+        'Use self-closing tags for elements with a closing tag but no content.',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const parserServices = getTemplateParserServices(context);
+
+    return {
+      Element$1({
```
JamesHenry, line 35: Please see comment about applying the filter of non-native elements in the selector

packages/eslint-plugin-template/docs/rules/self-closing-tags.md
```diff
@@ -0,0 +1,237 @@
...
+  - ../../tests/rules/self-closing-tags/cases.ts
+
+  In order to update this file, it is therefore those files which need to be updated, as well as potentially the generator script:
+  - ../../../../tools/scripts/generate-rule-docs.ts
+
+-->
+
+<br>
+
+# `@angular-eslint/template/self-closing-tags`
```
JamesHenry, line 16: To match existing conventions, please go for `prefer-self-closing-tags` as the rule name

JamesHenry (changes requested):
Thanks a lot for this @aleesaan! Please see the comments inline

packages/utils/src/eslint-plugin-template/native-elements.ts
aleesaan: Thank you so much for the feedback! I was in fact looking for this kind of suggestions 😅 I pushed the fixes, let me know how it looks.

JamesHenry:
Thanks a lot @aleesaan after that one final change and syncing up with the latest this is ready to merge!

aleesaan:
Fixed the last things and rebased! Let me know if it's all looking food :)

JamesHenry (approved):

Issue #238, opened 2020-12-19
Consistent naming for accessibility rules


dev054:
I saw that some accessibility rules starts with `accessibility-*`, but others not, like:

- click-events-have-key-events
- mouse-events-have-key-events
- no-autofocus
- no-distracting-elements
- no-positive-tabindex

I'd want to propose to naming them consistently starting with `accessibility` or even create a subpackage like `eslint-plugin-template-accessibility`.

dev054:
Not related to accessibility, but I think `prefer-on-push-component-change-detection` could be better spelled as `prefer-component-on-push-change-detection` to be in sync with `prefer-*` rules.

bbarry:
I am not sure if a separate suppackage is warranted but I think a config that includes all of them would be good.

rafaelss95:
@JamesHenry Now that Angular v13 is closer and we'll probably have a major version shortly thereafter, do you have any thoughts on this?

rafaelss95:
Indeed this prefix doesn't seem like a good idea. However, today we already have 21 template rules, most of them (more specifically, 11) focused on accessibility, not to mention the rules that we can possibly port from the https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/master/docs/rules (TBH, I miss  the _role_ related rules here).

With that in mind, would a new package called something like `@angular-eslint/accessibility` or `@angular-eslint/template-accessibility` be such a weird idea?

IMHO, it'd be good to separate which rules/utilities are geared exclusively to Angular (basically its styleguide or best practices) and those geared towards accessibility in a more general context.

In any case, I'll be fine either way, as long as we keep a standard.

JamesHenry:
v16 is now the latest release and the `accessibility-` prefix has been stripped from the rules which had it. Accessibility related rules can be easily discovered via the new accessibility config

See the release notes for me: https://github.com/angular-eslint/angular-eslint/releases/tag/v16.0.0

PR #1361, opened 2023-04-28
feat(es-plugin-template): create `accessibility` config

mijelin:
As suggested in https://github.com/angular-eslint/angular-eslint/issues/238#issuecomment-911594948, create a new `accessibility` preset so that each accessibility rule doesn't have to be enabled individually.


JamesHenry:
Thank you @mijelin.

The `accessibility` config will be available as part of the upcoming v16 release this week. There are some tooling aspect this PR didn't include, and I am making changes to some existing rules which need to be taken into account by the new config

PR #1398, opened 2023-05-30
fix(eslint-plugin): [valid-aria] tristate "mixed" value not handled

### LINKED ISSUE #1397: [valid-aria] - tristate "mixed" value not handled

I get a linter error when i use `aria-checked="mixed"` or `aria-pressed="mixed"`.

The value `mixed` exists : 
- checked: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#values
- pressed : https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed#values

```JSON
{
  "rules": {
    "@angular-eslint/template/accessibility-valid-aria": "error"
  }
}
```

```HTML
<label class="checkbox">
   <input
      class="checkbox-input"
      type="checkbox"
      [checked]="condition"
      (click)="toggleAll()"
    />
    <span
      class="checkbox-mixed"
      role="checkbox"
      aria-checked="mixed"></span>
</label>
```

**Suggested Change**

After investigating the aria-query library and specifically the `tristate` types, I think the mixed type should be checked in `tristate` case in the `isValidAriaPropertyValue` function.

```ts
function isMixed(value: unknown): value is 'mixed' {
  return typeof value === 'string' && value === 'mixed';
}

function isTristate(value: unknown): boolean {
  return isMixed(value) || isBooleanLike(value) || isNil(value);
}
```

And then, add it to the switch case : https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/rules/valid-aria.ts#L157-L158

```ts
    case 'tristate':
      return isTristate(attributeValue);
```

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `15.2.1` |
| `@angular-eslint/template-parser`         | `15.2.1` |
| `@typescript-eslint/parser`               | `5.57.1` |
| `ESLint`                                  | `8.37.0` |
| `node`                                    | `18.12.1` |

image

mbriand-lucca:
## Description

Handle mixed value for tristate type in `valid-aria` rule

----

fixes #1397


JamesHenry (changes requested):
Thanks @mbriand-lucca, please add a test which covers this case

JamesHenry (approved):

JamesHenry:
Many thanks @mbriand-lucca!

PR #1407, opened 2023-05-31
fix: allow the combination of the "map binding" and "static value" use of the `style` and `class` attributes

### LINKED ISSUE #1406: [no-duplicate-attributes] - Allow the combination of the "map binding" and "static value" use of the `style` and `class` attributes

The rule `no-duplicate-attributes` should allow the combination of the "map binding" and "static value" use of the `style` and `class` attributes. According to the Angular documentation (https://angular.io/guide/style-precedence#style-precedence), Angular merges both attributes which means their combined use is valid.

```JSON
{
  "rules": {
    "@angular-eslint/template/no-duplicate-attributes":  "error"
  }
}
```

```HTML
// Valid :
<div class="bar" [class]="dynamic"></div>
<div style="color: blue" [style]="styleExpression"></div>

// Invalid :
<div class="bar" class="dynamic"></div>
<div style="color: blue" style="background-color: red"></div>
```

**Suggested Change**

Allow the combination of the "map binding" and "static value" use of the `style` and `class` attributes.

**Versions**

| package                                  | version   |
| ---------------------------------------- | --------- |
| `@angular-eslint/eslint-plugin-template` | `15.2.1`  |
| `@angular-eslint/template-parser`        | `15.2.1`  |
| `@typescript-eslint/parser`              | `5.57.1`  |
| `ESLint`                                 | `8.37.0`  |
| `node`                                   | `18.12.1` |

image

mbriand-lucca:
## Description

Allow the combination of the "map binding" and "static value" use of the `style` and `class` attributes : 

My proposition is not the best, I found another solution but in `angular/compiler` library, the `TmplAstTextAttribute` type doesn't contain the `type` attribute instead of other types (`TmplAstBoundEvent` and `TmplAstBoundAttribute`). 

If type was available, we could simple do : 

```ts
function findDuplicates<
  TAttributeType extends
    | TmplAstBoundEvent
    | TmplAstBoundAttribute
    | TmplAstTextAttribute,
>(elements: readonly TAttributeType[]): readonly TAttributeType[] {
  // According to the Angular documentation (https://angular.io/guide/style-precedence#style-precedence)
  // Angular merges both attributes which means their combined use is valid
  const mapAndStaticBindingAttributesAllowed = ['class', 'style'];
  return elements.filter((element) => {
    return elements.some((otherElement) => {
      if (
        mapAndStaticBindingAttributesAllowed.includes(
          getOriginalAttributeName(element),
        )
      ) {
        return (
          otherElement !== element &&
          getOriginalAttributeName(otherElement) ===
            getOriginalAttributeName(element) &&
          otherElement.type !== element.type
        );
      } else {
        return (
          otherElement !== element &&
          getOriginalAttributeName(otherElement) ===
            getOriginalAttributeName(element)
        );
      }
    });
  });
}
```

----

fixes #1406

### DIFF
--- a/packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
+++ b/packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
@@ -62,10 +62,52 @@ export default createESLintRule<Options, MessageIds>({
 
     return {
       Element$1({ inputs, outputs, attributes }: TmplAstElement) {
-        const duplicateInputsAndAttributes = findDuplicates([
-          ...inputs,
-          ...attributes,
-        ]);
+        // According to the Angular documentation (https://angular.io/guide/style-precedence#style-precedence)
+        // Angular merges both attributes which means their combined use is valid
+        const mapAndStaticBindingAttributesAllowed = ['class', 'style'];
+
+        let duplicateInputsAndAttributes: (
+          | TmplAstBoundEvent
+          | TmplAstBoundAttribute
+          | TmplAstTextAttribute
+        )[];
+
+        const inputsIgnored = inputs.filter((input) =>
+          mapAndStaticBindingAttributesAllowed.includes(
+            getOriginalAttributeName(input),
+          ),
+        );
+
+        if (inputsIgnored?.length > 0) {
+          const attributesIgnored = attributes.filter((attr) =>
+            mapAndStaticBindingAttributesAllowed.includes(
+              getOriginalAttributeName(attr),
+            ),
+          );
+          const inputsNotIgnored = inputs.filter(
+            (input) => !inputsIgnored.includes(input),
+          );
+          const attributesNotIgnored = attributes.filter(
+            (attr) => !attributesIgnored.includes(attr),
+          );
+          const ignoreDuplicated = [
+            ...findDuplicates(inputsIgnored),
+            ...findDuplicates(attributesIgnored),
+          ];
+          const notIgnoredDuplicates = [
+            ...findDuplicates(inputsNotIgnored),
+            ...findDuplicates(attributesNotIgnored),
+          ];
+          duplicateInputsAndAttributes = [
+            ...ignoreDuplicated,
+            ...notIgnoredDuplicates,
+          ];
+        } else {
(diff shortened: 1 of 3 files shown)


JamesHenry (changes requested):
My immediate reaction is that I think this would be better suited to be an option on the rule, rather than a change in the rule's default behaviour.

It's called `no-duplicate-value` after all, so I think banning duplication outright makes sense as a baseline, and then enabling some nuance via flag(s) makes sense too.

mbriand-lucca:
> My immediate reaction is that I think this would be better suited to be an option on the rule, rather than a change in the rule's default behaviour.
> 
> It's called `no-duplicate-value` after all, so I think banning duplication outright makes sense as a baseline, and then enabling some nuance via flag(s) makes sense too.

Done ;).

I'm not sure about the flag option name, is it ok? Another idea? 

I think it will be nice to get the angular links somewhere to understand which attribute is allowed and why, are you ok with that? I thought in the description flag option.

packages/eslint-plugin-template/src/rules/no-duplicate-attributes.ts
```diff
@@ -40,6 +42,11 @@ export default createESLintRule<Options, MessageIds>({
             default: DEFAULT_OPTIONS.allowTwoWayDataBinding,
             description: `Whether or not two-way data binding is allowed as an exception to the rule.`,
           },
+          allowAngularStylePrecedenceDuplicates: {
+            type: 'boolean',
+            default: DEFAULT_OPTIONS.allowAngularStylePrecedenceDuplicates,
+            description: `Whether or not angular style precedence is allowed as an exception to the rule.`,
```
JamesHenry, line 48:
```suggestion
            description: `Whether or not angular style precedence is allowed as an exception to the rule. See https://angular.io/guide/style-precedence#style-precedence`,
```

JamesHenry (dismissed):

JamesHenry (dismissed):

JamesHenry (approved):

JamesHenry:
Thanks again @mbriand-lucca!

PR #1427, opened 2023-06-16
feat(eslint-plugin-template): add a11y ESLint rule headings-consecutive-order

imonikaimonika:
add accessibility ESLint rule headings-consecutive-order

WCAG link: https://www.w3.org/WAI/tutorials/page-structure/headings/


abaran30:
FYI @sandikbarr

packages/eslint-plugin-template/src/rules/headings-consecutive-order.ts
```diff
@@ -0,0 +1,51 @@
...
+      headingsOrder: 'Headings number sequence must not be skipped',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const parserServices = getTemplateParserServices(context);
+    let headingLastIndex: number;
+
+    return {
+      [`Element$1[name=/^h\\d/]`](element: TmplAstElement) {
```
sandikbarr, line 29: Might be good to filter this to only known heading elements similar to the `elements-content` rule:
```suggestion
      [`Element$1[name=/^(h1|h2|h3|h4|h5|h6)$/`](element: TmplAstElement) {
```

packages/eslint-plugin-template/tests/rules/headings-consecutive-order/cases.ts
```diff
@@ -0,0 +1,53 @@
...
+const messageId: MessageId = 'headingsOrder';
+
+export const valid = [
+  // valid examples from https://www.w3.org/WAI/tutorials/page-structure/headings/
+  {
+    // organize passages on text
+    code: `<h1>Heading 1</h1>
+            <h2>Heading 2</h2>
+            <h2>Heading 2</h2>
+            <h2>Heading 2</h2>`,
```
sandikbarr, line 13: I wonder if it would be worthwhile to ignore heading elements inside an `<ng-template>` since templates are rendered separately and not typically in the order they are written in the code.

imonikaimonika: Thanks, yes, technically it is not expected to follow the ordering..

packages/eslint-plugin-template/src/rules/headings-consecutive-order.ts
```diff
@@ -0,0 +1,86 @@
...
+            return;
+          }
+
+          const headingCurrentIndex = +element.name.slice(1);
+
+          if (
+            headingLastIndex &&
+            headingCurrentIndex - headingLastIndex !== 1 &&
+            headingCurrentIndex !== 1 &&
+            headingCurrentIndex > headingLastIndex
```
sandikbarr, line 69: Is the `headingCurrentIndex !== 1` necessary? Seems `headingCurrentIndex` could not be both `1` && `> headingLastIndex`
```suggestion
            headingCurrentIndex > headingLastIndex
```

sandikbarr (commented):
Thanks for the addition to ignore headings in `ng-template` 👍  
I added a few additional suggestions.
Appreciate the contribution!

packages/eslint-plugin-template/src/rules/headings-consecutive-order.ts
imonikaimonika: Thanks for the reviews, Sandi :)

sandikbarr (approved):

abaran30:
@imonikaimonika would you be able to generate the docs for his rule with `yarn update-rule-docs`, please?

packages/eslint-plugin-template/tests/rules/headings-consecutive-order/cases.ts
```diff
@@ -0,0 +1,63 @@
+import { convertAnnotatedSourceToFailureCase } from '@angular-eslint/utils';
+import { MessageId } from '../../../src/rules/headings-consecutive-order';
```
abaran30, line 2: As reported by CI, please use `import type` here.

packages/eslint-plugin-template/src/rules/headings-consecutive-order.ts
```diff
@@ -0,0 +1,82 @@
...
+            headingCurrentIndex - headingLastIndex !== 1 &&
+            headingCurrentIndex > headingLastIndex
+          ) {
+            const loc = parserServices.convertNodeSourceSpanToLoc(
+              element.sourceSpan,
+            );
+
+            context.report({
+              loc,
+              node: element as any,
```
abaran30, line 73: I don't think the `node` property is necessary and that `loc` and `messageId` should be enough. LMK if this isn't the case. 🙂

packages/eslint-plugin-template/src/rules/headings-consecutive-order.ts
```diff
@@ -0,0 +1,82 @@
...
+export type MessageId = 'headingsOrder';
+
+export default createESLintRule<[], MessageId>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      // link: https://www.w3.org/WAI/tutorials/page-structure/headings/
+      description:
+        'Ensures that headings have consecutive order from largest to lowest h1 -> h6',
```
abaran30, line 20: We're looking to follow a convention where a11y rules contain the prefix `[Accessibility]` in the description (example [here](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/rules/valid-aria.ts#L30C12-L30C12)). Please add that here.

abaran30 (changes requested):

packages/eslint-plugin-template/tests/rules/headings-consecutive-order/cases.ts
imonikaimonika: Thanks!

packages/eslint-plugin-template/src/configs/accessibility.json
```diff
@@ -12,6 +12,7 @@
     "@angular-eslint/template/no-distracting-elements": "error",
     "@angular-eslint/template/role-has-required-aria": "error",
     "@angular-eslint/template/table-scope": "error",
-    "@angular-eslint/template/valid-aria": "error"
+    "@angular-eslint/template/valid-aria": "error",
+    "@angular-eslint/template/headings-consecutive-order": "error"
```
sandikbarr, line 16: I wonder if similar to the recommended ruleset, we should wait to add new rules to the accessibility ruleset until the next major version release. The accessibility ruleset is new to v16, but if we add a new rule on a minor version, people currently using it on their Angular 16 projects could potentially get new lint violations with a minor version update without having changed their configuration, which we should try to avoid.

If/when we do add it here, we should ensure it is alphabetized in this ruleset.

sandikbarr:
@imonikaimonika got few questions to understand more about your project(s) setup and what problems this rule would help solve-
* typically, I think of headings as being spread out among several components, but personally, I think my experience is more from developing to requirements with under use of headings rather than proper use. Do you structure your templates to have headings organized in a single template so order can be checked?
* generally, I find integration/e2e testing that can check the entire rendered page better at checking for headings as a whole. For example, axe-core [headings-order](https://dequeuniversity.com/rules/axe/4.7/heading-order?application=RuleDescription)
* Since it would be a newly established lint rule (to the best of my knowledge not pre-existing in any of the other accessibility lint rulesets,) we do wonder about adding it here or if you could use it as more of a custom/in-house rule.

packages/eslint-plugin-template/src/configs/accessibility.json
imonikaimonika: Don't users manually add the rules they choose in the .eslintrc.json ?
If this file makes it automatically enabled rule I can remove, but I think those rules are added manually so still if people don't know about it, they won't add it (or if they don't want to).

I can alphabetize it , sure

imonikaimonika:
> @imonikaimonika got few questions to understand more about your project(s) setup and what problems this rule would help solve-
> 
> * typically, I think of headings as being spread out among several components, but personally, I think my experience is more from developing to requirements with under use of headings rather than proper use. Do you structure your templates to have headings organized in a single template so order can be checked?
> * generally, I find integration/e2e testing that can check the entire rendered page better at checking for headings as a whole. For example, axe-core [headings-order](https://dequeuniversity.com/rules/axe/4.7/heading-order?application=RuleDescription)
> * Since it would be a newly established lint rule (to the best of my knowledge not pre-existing in any of the other accessibility lint rulesets,) we do wonder about adding it here or if you could use it as more of a custom/in-house rule.


@sandikbarr , let me answer for each one:

-  yes, my experience is also spreading headings among components, most of the time, and yes, there are such cases when skipping headings order will remain uncaught by the eslint rule, since the gap will be in different components (ex.: one component with main heading h1 and then nested component jumps to h3); However, there could be also cases of this skip happening in a single html template, perhaps even more so of some specific products/websites. It is still an a11y bug and I think helpful to be caught whenever possible. 
- The axe-core I just saw - seems to  check for that problem, yes, perhaps even better - but.. me for instance hear for it the 1st time, it is a separate tool, I think people will only use it if their focus is an a11y bug-fix campaign rather in their daily work.. and do the majority even has awareness that this is an accessibility issue to find the tool and even use it.. My point is , with the ESLint which every dev uses, this rule as part of it can not only solve some occurrences , but also raise awareness this is a problem .
- Thanks for noticing the novelty - yes it is "not pre-existing in any of the other accessibility lint rulesets" - that's what I have come up with - a new rule - that was my goal hah. Well, I have solved really many such bugs and that's how I got acquainted that this is breaking a11y guidelines - and I thought it can be automated - and there we go - it is working and pointing to actual a11y bugs - even if in not all possible implementations. Even so, I wouldn't say it is a custom/ in-house rule.. It is a wide accessibility issue and even semantic html, so it even solves more than 1 problems.. I have even provided official WCAG docs with actual examples copied from it.

packages/eslint-plugin-template/src/configs/accessibility.json
sandikbarr: This accessibility ruleset is now being configured for Angular projects with the `ng add @angular-eslint/schematics` schematic, so yes, rules in this accessibility.json files will be added to projects that use the accessibility ruleset configuration, which is now included by default.

packages/eslint-plugin-template/src/rules/headings-consecutive-order.ts
```diff
@@ -0,0 +1,81 @@
...
+          if (
+            headingLastIndex &&
+            headingCurrentIndex - headingLastIndex !== 1 &&
+            headingCurrentIndex > headingLastIndex
+          ) {
+            const loc = parserServices.convertNodeSourceSpanToLoc(
+              element.sourceSpan,
+            );
+
+            context.report({
```
jerone, line 71: Is it correct that **only the first** wrongly header instance is being reported, when there might be multiple?

sandikbarr:
> 
>     * yes, my experience is also spreading headings among components, most of the time, and yes, there are such cases when skipping headings order will remain uncaught by the eslint rule, since the gap will be in different components (ex.: one component with main heading h1 and then nested component jumps to h3); However, there could be also cases of this skip happening in a single html template, perhaps even more so of some specific products/websites. It is still an a11y bug and I think helpful to be caught whenever possible.
> 

agreed, skipping heading levels in individual template files is still an a11y bug.

>     * The axe-core I just saw - seems to  check for that problem, yes, perhaps even better - but.. me for instance hear for it the 1st time, it is a separate tool, I think people will only use it if their focus is an a11y bug-fix campaign rather in their daily work.. and do the majority even has awareness that this is an accessibility issue to find the tool and even use it.. My point is , with the ESLint which every dev uses, this rule as part of it can not only solve some occurrences , but also raise awareness this is a problem .
> 

In the spirit of using the best/right tool for the job, I personally recommend testing heading level on a fully rendered page, which is the purpose of axe-core. 

>     * Thanks for noticing the novelty - yes it is "not pre-existing in any of the other accessibility lint rulesets" - that's what I have come up with - a new rule - that was my goal hah. Well, I have solved really many such bugs and that's how I got acquainted that this is breaking a11y guidelines - and I thought it can be automated - and there we go - it is working and pointing to actual a11y bugs - even if in not all possible implementations. Even so, I wouldn't say it is a custom/ in-house rule.. It is a wide accessibility issue and even semantic html, so it even solves more than 1 problems.. I have even provided official WCAG docs with actual examples copied from it.

Teaching accessibility is not the purpose of Angular ESLint. There has been [past discussion about creating a separate package for angular-eslint/accessibility](https://github.com/angular-eslint/angular-eslint/issues/238#issuecomment-911797812). My own understanding of the rules accessibility rules that exist in angular-eslint today are for parity with other well known eslint packages. Creating a separate plugin for this and any other new accessibility rules not related to that feature parity would be a better direction.

PR #1430, opened 2023-06-23
feat: generator to generate skeleton for new lint rules

abaran30:
This pull request introduces an Nx generator to generate a skeleton for new lint rules.

### DIFF
--- a/packages/nx-plugin/generators.json
+++ b/packages/nx-plugin/generators.json
@@ -2,5 +2,11 @@
   "$schema": "http://json-schema.org/schema",
   "name": "nx-plugin",
   "version": "0.0.1",
-  "generators": {}
+  "generators": {
+    "rule": {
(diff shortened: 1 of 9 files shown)


abaran30:
@JamesHenry I thought that this might be helpful to have, even if only a little helpful 🙂. It's pretty rudimentary and could definitely use enhancements moving forward. One immediate area of improvement is the generator adding the new rule to the package exports. Currently, the generator uses string replacement and the imports/exports don't get sorted even when formatted.

LMK what you think.

FYI @sandikbarr

JamesHenry (approved):
Seems like a good idea to have, thanks! As you say we can incrementally improve it, so I will merge for now

JamesHenry:
Note that this was a `chore` not a feat, because it is only relevant for the repo and should not appear in the changelogs

PR #1469, opened 2023-08-05
fix(eslint-plugin): export require-localize-metadata rule

amikheychik:
Prevents the `Definition for rule '@angular-eslint/require-localize-metadata' was not found` ESLint error.

Also, fix sorted order for the `prefer-standalone-component` rule.

### DIFF
--- a/packages/eslint-plugin/src/configs/all.json
+++ b/packages/eslint-plugin/src/configs/all.json
@@ -29,6 +29,7 @@
     "@angular-eslint/prefer-output-readonly": "error",
     "@angular-eslint/prefer-standalone-component": "error",
     "@angular-eslint/relative-url-prefix": "error",
+    "@angular-eslint/require-localize-metadata": "error",
     "@angular-eslint/sort-ngmodule-metadata-arrays": "error",
     "@angular-eslint/use-component-selector": "error",
     "@angular-eslint/use-component-view-encapsulation": "error",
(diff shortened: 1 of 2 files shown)


abaran30:
@amikheychik thanks very much for catching this!

It looks like something went wrong with the code formatting check (see [here](https://github.com/angular-eslint/angular-eslint/actions/runs/5772396750/job/15659745424?pr=1469)). Would you be able to take a look?

FYI @JamesHenry let's also add this rule to the all config when appropriate.

abaran30 (approved):

amikheychik:
@abaran30 I've updated the formatting, should be good now.

I've also added the rule to the `configs/all.json`, as it should contain all existing rules.

abaran30:
Thank you @amikheychik!

I'm going to check and make sure that it'd be okay to update the all config in this PR. We wouldn't want to surprise users of the all config unexpectedly with this rule on the next release.

amikheychik:
@abaran30 the problem is that without this rule in the `all.json` [tests fail](https://github.com/angular-eslint/angular-eslint/actions/runs/5772396750/job/15659745377) (`configs › all › should contain all of the rules from the plugin`)

BTW, this bug is present since v13.2.0 when the rule was introduced. Would you like me to backport it to v13, v14, and v15?

JamesHenry:
Yeah new rules should always be added to the all config, it’s expected and can’t wait until major releases. I think users wanting to enable all rules all the time is a pretty strange idea personally but the option is there for those that do

abaran30 (approved):

amikheychik:
@JamesHenry would you publish v15, v14, and v13 patch versions if I backport this patch there?

PR #1470, opened 2023-08-09
feat(eslint-plugin-template): [no-target-blank] add rule

pumano:
Add rule about adding rel="noreferrer" to links protects your site's users against having the site you've linked to potentially hijacking the browser (via rogue JS).

Details about vulnerability: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

Why without noopener? 
`noopener noreferrer` is redundant, since `noreferrer` includes the functionality of `noopener`. [html.spec.whatwg.org/multipage/links.html#link-type-noreferrer](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer)

Workaround: all current versions of major browsers now automatically use the behavior of `rel="noopener"` for any `target="_blank"` link, nullifying this issue.

But this rule help to properly add `rel="noreferrer"` while `target="_blank"` used, check proper spelling, and also works against older browsers (which can be used for hijacking).


json-derulo:
IMHO we don't need this rule in 2023. The affected browser versions are well out of official Angular browser support, see the [docs](https://angular.io/guide/browser-support). Also since Angular 15, ES2022 code is produced by default. Recent Angular applications may not even work in such old browsers due to missing features of ES2022.

json-derulo:
@pumano I totally understand the issue. If this was a plugin for HTML files in general, I think it would make sense to add such a rule. But this plugin only works in Angular templates. The current version of this plugin is only compatible with Angular 16. It is very unlikely that projects using Angular 16 support browsers affected by this problem. Therefore, IMO it doesn't make sense to add this rule now. Two years ago it would have made sense.

JamesHenry:
Thank you very much for contributing, @pumano! I really appreciate you looking to give back to angular-eslint.

On balance, given the vulnerability this looks to avoid is only relevant in browsers which Angular already does not support, angular-eslint does not need to maintain a rule for this.

Do of course feel free to use our published utilities to make this available to yourself via a custom ESLint plugin, and thank you again for your contribution to the community.

PR #1472, opened 2023-08-10
fix(eslint-plugin-template): fix fixer of inline templates

### LINKED ISSUE #1464: [prefer-self-closing-tags] auto-fix results in invalid HTML

Starting with **16.1.0**, the [new auto-fix for prefer-self-closing-tags](https://github.com/angular-eslint/angular-eslint/pull/1322) results in invalid HTML in our templates.  It looks like the column calculation is off by one.  

```JSON
{
  "rules": {
    "@angular-eslint/template/prefer-self-closing-tags": "error"
  }
}
```

**Before**

```html
<app-datepicker data-testid="date" formControlName="date" [maxDate]="today" required></app-datepicker>
```

**After**

```html
<app-datepicker data-testid="date" formControlName="date" [maxDate]="today" require />>
```

All of our projects are private, so I cannot provide a full repro, however it may help to know that we only use inline templates with our components.

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `16.1.0` |
| `@angular-eslint/template-parser`         | `16.1.0` |
| `@typescript-eslint/parser`               | `16.1.0` |
| `ESLint`                                  |`8.45.0` |
| `node`                                    | `18.16.1` |

json-derulo:
Currently, the rule fixer does not return correct results for inline templates. It applies to all template rules with a fixer, see https://github.com/angular-eslint/angular-eslint/issues/1464#issuecomment-1673008115. This is fixed by this PR.

Closes #1464 

Note: I wanted to add some tests, but I'm not sure where to start. I performed manual tests of all the cases described in the linked issue.

### DIFF
--- a/packages/eslint-plugin-template/src/processors.ts
+++ b/packages/eslint-plugin-template/src/processors.ts
@@ -257,7 +257,7 @@ export function postprocessComponentFile(
             message.endLine + rangeData.lineAndCharacter.start.line;
 
           if (message.fix) {
-            const startOffset = rangeData.range[0];
+            const startOffset = rangeData.range[0] + 1;
             message.fix.range = [
               startOffset + message.fix.range[0],
               startOffset + message.fix.range[1],
--- a/packages/eslint-plugin-template/tests/processors.test.ts
+++ b/packages/eslint-plugin-template/tests/processors.test.ts
@@ -467,5 +467,47 @@ describe('extract-inline-html', () => {
         });
       });
     });
+
+    describe('messages from inline template HTML', () => {
+      it('should adjust message locations from the inline templates', () => {
+        const fileContent = `
+          @Component({
+            template: '<div ([ngModel])="value"></div>',
+          })
+          export class Component {
+            value = '';
+          }
+        `;
+        processors['extract-inline-html'].preprocess(
+          fileContent,
+          'test.component.ts',
+        );
+        const mockError = {
+          ruleId: 'banana-in-box',
+          severity: 2,
+          message: 'Invalid binding syntax. Use [(expr)] instead',
+          line: 1,
+          column: 28,
+          nodeType: 'Literal',
+          messageId: 'bananaInBox',
+          endLine: 1,
+          endColumn: 39,
+          fix: { range: [5, 16], text: '[(ngModel)]' },
+        };
+        const expectedMessage = {
+          ...mockError,
+          line: 3,
+          endLine: 3,
+          fix: { range: [52, 63], text: '[(ngModel)]' },
+        };
+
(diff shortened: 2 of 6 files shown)


JamesHenry (changes requested):
Thanks @json-derulo!

When you say you weren't sure where to start... Did you consider adding the cases you reported to the relevant rule's test cases?

You just need to look at the existing tests, duplicate one or more and amend.

We will definitely want to capture your findings as tests before proceeding

json-derulo:
@JamesHenry added a test case

JamesHenry:
@json-derulo I meant more capturing the real experience you were having using rules e.g. here https://github.com/angular-eslint/angular-eslint/issues/1464#issuecomment-1673008115

json-derulo:
@JamesHenry I am a bit stuck trying to add tests for the real cases. I have tried to setup test cases with the `RuleTester`. When I use parser `@typescript-eslint/parser`, I'm always getting the following error:

>  Error while loading rule: You have used a rule which requires '@angular-eslint/template-parser' to be used as the 'parser' in your ESLint config.

I have tried to extend from `@angular-eslint/template/process-inline-templates`, use processor `@angular-eslint/template/extract-inline-html`, plugin `@angular-eslint/template` and several combinations of those. The error persists.

When I use parser `@angular-eslint/template-parser` no errors are reported for the invalid cases, it seems like it cannot understand the TypeScript syntax.

Do you have any other idea how I could test this?

JamesHenry:
Yes sorry I think I was skipping some relevant complexity here, it probably isn’t a good fit for the unit tests, it would be better as an integration test.

I’m on my phone and thinking out loud but I think we could add a relevant file to the generated fixture in one of the existing integration tests and snapshot it before and after applying fixing. 

If you have time to play around with this idea in the next few days then that’s great, otherwise I can take a proper look this weekend sometime

json-derulo:
@JamesHenry thanks for your input, I think I figured it out by myself. I've implemented an integration test which generates a fresh Angular Workspace and runs Angular ESLint Add schematics. Afterwards, a new test component file is generated with the test cases. I excluded the `prefer-self-closing-tags` rule there because it's not part of the recommended rule set and would require additional steps. Finally, the lint fixer is executed, and the resulting file content is checked against a snapshot.

Not sure why the CI is failing, on my machine the integration tests succeed. Also the failure is related to a different integration test which was not changed.

json-derulo:
@JamesHenry after rebasing the latest changes from the main branch, CI is green now. Could you have a look again to the changes of this PR?

JamesHenry (approved):
Great work, thank you @json-derulo!

PR #1477, opened 2023-08-17
feat(eslint-plugin-template): [prefer-ngsrc] add rule

### LINKED ISSUE #1265: New rule suggestion: no-restricted-attributes
Hi, while integrating the NgOptimizedImage directive (https://angular.io/guide/image-directive) I searched for a way to restrict the usage of the `src` attribute in the template in favour of the `ngSrc` attribute.
This can be done via the rule "no-restricted-attrs" of another project: "HTML ESlint" (https://yeonjuan.github.io/html-eslint/docs/rules/no-restricted-attrs/), hovever, when already using "Angular ESLint" to lint HTML files, it is not possible to combine the parsers of both packages. So I think it is useful - not only for this use case - to provide such functionality in "Angular ESlint", too.

The provided documentation of the other project (https://yeonjuan.github.io/html-eslint/docs/rules/no-restricted-attrs/) contains an example of how to use the rule (I'll provide another example of how it could look like below) and I think it is a very convenient way to configure which tags are allowed and which are not.

```JSON
 {
  "rules": {
    "@html-angular/no-restricted-attributes": ["warn",  {
      "tagPatterns": ["^img$"],
      "attrPatterns": ["src.*"]
      "message": "consider using ngSrc instead of src"
  }]
  }
}

```

```HTML
// Warning
<img src="imageSource">

// Correct
<img ngSrc="imageSource">
```

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `15.0.0` |
| `@angular-eslint/template-parser`         | `15.0.0` |
| `@typescript-eslint/parser`               | `5.43.0` |
| `ESLint`                                  | `8.27.0` |
| `node`                                    | `16.17.0` |

```sh
Angular CLI: 15.0.0
Node: 16.17.0
Package Manager: npm 8.19.2
OS: linux x64

Angular: 
... 

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1500.0 (cli-only)
@angular-devkit/core         15.0.0 (cli-only)
@angular-devkit/schematics   15.0.0 (cli-only)
@schematics/angular          15.0.0 (cli-only)
```

strawberry-choco:
Adds a rule to use the `ngSrc` attribute over `src`.

Closes #1265


packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
```diff
@@ -0,0 +1,68 @@
...
+  defaultOptions: [],
+  create(context) {
+    const parserServices = getTemplateParserServices(context);
+
+    return {
+      'Element$1[name=img]'(element: TmplAstElement) {
+        const ngSrcAttribute = hasNgSrcAttribute(element);
+        const srcAttribute = hasNormalSrcAttribute(element);
+
+        if (srcAttribute == null) {
```
JamesHenry, line 38: Whilst functional, this is an unintuitive choice of condition for checking a value which will never be `null`.

Let's prefer simple code:

```suggestion
        if (!srcAttribute) {
```

packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        const loc = parserServices.convertNodeSourceSpanToLoc(
+          srcAttribute.sourceSpan,
+        );
+
+        context.report({
+          loc,
+          messageId:
+            ngSrcAttribute == null ? 'missingAttribute' : 'invalidDoubleSource',
```
JamesHenry, line 49: Whilst functional, this is an unintuitive choice of condition for checking a value which will never be `null`.

Let's prefer simple code:

```suggestion
            !ngSrcAttribute ? 'missingAttribute' : 'invalidDoubleSource',
```

JamesHenry (changes requested):
Thank you @strawberry-choco!

Just a couple of instances of the same point and then this is good to go

packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
strawberry-choco: I am using double equals here, which checks for both `null` and `undefined`. Your suggestion converts the value into a boolean. While it works perfectly fine here since the type of `srcAttribute` is an object, I had many troubles with this convention when used on numbers, since `!0` equals `false`. For the sake of consistency and clear intention (I am checking if the variable holds a value and not negating a boolean) I always use `== null`.

Since this change is not something magnificent, I do not mind applying them if you still prefer it. :)

JamesHenry: I understand how it works, that's why I started the comment with "whilst functional". I have been writing JavaScript for 19 years. Again, what I said was that what you wrote was "unintuitive".

I am strongly anti "clever code", particularly in a community project. Code is read far more than it is written.

Replacing your usage of `value == null` with `value === undefined || value === null` would make your code much more intuitive for readers of all levels.

In this case, there would be no additional value in only narrowing to `undefined` and `null` so we may as make the code the simplest it can possibly be and discount all falsey values.

JamesHenry: I also still don't understand where you are anticipating a `null` value to come from. And if the answer is nowhere, and you are trying to be precise with this, then you should have written `srcAttribute  === undefined`. You currently have it in a strange middle ground where you are discounting one extra falsey value which should never occur, instead of all of them.

strawberry-choco: I have a slightly different perspective on this, but that should not stop us from getting this PR done. :)

Regarding your second comment: That is a habit coming from the eslint rule `eqeqeq` with the `null: 'ignore'` option.

strawberry-choco: Suggestions have been implemented.

JamesHenry:
@strawberry-choco You just need to fix up the formatting, you can run `yarn format`

packages/eslint-plugin-template/tests/rules/prefer-ngsrc/cases.ts
```diff
@@ -0,0 +1,86 @@
...
+        char: '#',
+        messageId: missingAttribute,
+      },
+    ],
+  }),
+  convertAnnotatedSourceToFailureCase({
+    description: 'should fail image when using both src and ngsrc',
+    annotatedSource: `
+      <ng-template>
+        <img ngSrc="http://localhost" src="http://localhost">
```
jerone, line 45: I read from the source that the error will always be at the `src`-attribute, and not necessarily the second attribute. 
Might be a good idea to add a test where `src` and `ngSrc` are reversed.

JamesHenry: Agreed, thanks @jerone!

packages/eslint-plugin-template/src/configs/accessibility.json
```diff
@@ -10,6 +10,7 @@
     "@angular-eslint/template/mouse-events-have-key-events": "error",
     "@angular-eslint/template/no-autofocus": "error",
     "@angular-eslint/template/no-distracting-elements": "error",
+    "@angular-eslint/template/prefer-ngsrc": "error",
```
JamesHenry, line 13: Please remove this again, it is what is causing the integration tests to fail

A) you should not manually patch this file, it is generated based on using a prefix in the rule description
B) we will not be modifying the accessibility config in minor versions. We will only do that in major versions
C) I don't believe this even is an accessibility rule? It's a web performance rule?

JamesHenry:
@strawberry-choco just following up on this one, please could you add a test case per jerone's comment, and revert the update to the accessibility config?

packages/eslint-plugin-template/tests/rules/prefer-ngsrc/cases.ts
strawberry-choco: Fixed.

packages/eslint-plugin-template/src/configs/accessibility.json
strawberry-choco: Fixed.

strawberry-choco:
> @strawberry-choco just following up on this one, please could you add a test case per jerone's comment, and revert the update to the accessibility config?

Sorry, got a little busy recently and forgot about this PR. Thanks for the reminder.

JamesHenry (approved):

JamesHenry:
Thanks!

PR #1497, opened 2023-08-22
feat(eslint-plugin-template): allow `alias` option in [use-track-by-function]

### LINKED ISSUE #1323: [template/use-track-by-function] allow custom trackBy-directives

```JSON
{
  "rules": {
    "@angular-eslint/template/use-track-by-function": "warn"
  }
}
```

```HTML
<li *ngFor="let item of items; trackByProperty: 'id'"></li>
```

**Versions**

| `@angular-eslint/eslint-plugin`    | `5.45.1` |
| `@typescript-eslint/parser`        | `5.45.1` |
| `ESLint`                           | `8.29.0` |
| `node`                             | `18.12.1` |
| `ng-for-track-by-property`                             | `15.0.1` |

Is: the *ngFor has the ESLint warning "ESLint: Missing trackBy function in ngFor directive" even though `trackByProperty` was provided, which has the same functionality as providing a `trackBy`.

--- 

Custom directives that provide the same functionality as trackBy, for example `trackByProperty` by dependency "ng-for-track-by-property", aren't recognized by rule "@angular-eslint/template/use-track-by-function" and reported as if a trackBy was missing.

Would it be possible to allow optional configuration of the function `isNgForTrackBy` so that additional directives, not just "ngForTrackBy", are seen as a trackBy-function?

Here's an example of how these custom trackBy-directives are implemented: https://github.com/nigrosimone/ng-for-track-by-property/blob/main/projects/ng-for-track-by-property/src/lib/ng-for-track-by-property.directive.ts

arturovt:
This commit adds an `alias` option to the `use-track-by-function` rule, allowing the specification of a list of properties that should be treated similarly to `ngForTrackBy`.

Closes #1323

### DIFF
--- a/packages/eslint-plugin-template/jest.config.ts
+++ b/packages/eslint-plugin-template/jest.config.ts
@@ -4,13 +4,13 @@
 export default {
   displayName: 'eslint-plugin-template',
   preset: '../../jest.preset.js',
-  globals: {
-    'ts-jest': {
-      tsconfig: '<rootDir>/tsconfig.spec.json',
-    },
-  },
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        tsconfig: '<rootDir>/tsconfig.spec.json',
+      },
+    ],
   },
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx'],
   coverageDirectory: '../../coverage/packages/eslint-plugin-template',
--- a/packages/eslint-plugin-template/tests/rules/use-track-by-function/cases.ts
+++ b/packages/eslint-plugin-template/tests/rules/use-track-by-function/cases.ts
@@ -68,6 +68,14 @@ export const valid = [
       <div *ngFor="let photo of photos; trackById"></div>
     `,
   },
+  {
+    options: [{ alias: ['ngForTrackByProperty'] }],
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin-template/jest.config.ts
JamesHenry, line 1: Why did this require changing a jest config?

packages/eslint-plugin-template/tests/rules/use-track-by-function/cases.ts
JamesHenry, line 1: Please add a full ng-template example of using the custom directive (both a passing and failing case, like we have elsewhere in this file

JamesHenry (changes requested):
Thanks @arturovt! Comments inline

packages/eslint-plugin-template/jest.config.ts
arturovt: This is causing warnings to appear in the console, and these warnings are overlapping everything I logged with `console.log`.

packages/eslint-plugin-template/tests/rules/use-track-by-function/cases.ts
arturovt: Updated.

JamesHenry:
You need to run `yarn update-rule-docs` please

JamesHenry (approved):
Thanks a lot @arturovt!

PR #1520, opened 2023-09-08
fix(eslint-plugin-template): [prefer-self-closing-tags] improve code style of fixer result

json-derulo:
Currently when running the fixer of the `prefer-self-closing-tags` rule, in some scenarios a result is generated, which is considered as bad code style by common sense and formatters like prettier. More specifically, currently the fixer always adds an additional whitespace before the self closing tag. However when the end of the starting tag is indented or on a new line, this additional whitespace produces misalignment.

For example, take the following code:

```html
<my-component
  someAttribute="someValue"
></my-component>
```

Currently the fixer would produce the following reusult:

```html
<my-component
  someAttribute="someValue"
 />
```

With the changes of this PR it would become:

```html
<my-component
  someAttribute="someValue"
/>
```

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/prefer-self-closing-tags.md
+++ b/packages/eslint-plugin-template/docs/rules/prefer-self-closing-tags.md
@@ -118,6 +118,37 @@ The rule does not have any configuration options.
 ~~~~~~~~~~~~~~~
 ```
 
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks @json-derulo!

PR #1540, opened 2023-09-24
chore: fix output path separator in generate-rule-docs.ts on windows

sonallux:
When the `generate-rule-docs.ts` is executed on windows it will output paths with backslashes in the rule doc header. This commit fixes this by replacing the backslashes with normal slashes, so the output of the `generate-rule-docs.ts` is equal across different OS.


JamesHenry:
Thank you @sonallux!

PR #1559, opened 2023-10-18
chore: fix rule generator template defaultOptions

abaran30:
This pull request fixes the template for the new rule generator by removing `defaultOptions` from the `meta` object, as the property does not belong there.


JamesHenry (approved):

JamesHenry:
Thanks! FYI I changed this one to a chore so that it doesn't appear in the changelog

PR #1586, opened 2023-11-07
fix: generate type declarations for published packages

timkendrick:
Some of the `@angular-eslint/*` packages declare a `"types"` field in their published `package.json`, however their respective `tsconfig.build.json` configuration files lack the`"declaration": true` field.

This means that no `.d.ts` files will be created for these packages when building the package for publishing to npm, causing the `"types"` field to reference a missing file.

This PR fixes this, allowing the published npm packages to be imported directly within a TypeScript project.


JamesHenry:
Thank you @timkendrick!

PR #1600, opened 2023-11-08
feat(eslint-plugin-template): [prefer-control-flow] add rule

### LINKED ISSUE #1591: [prefer-control-flow-syntax] Add new rule to support the new control flow syntax (Request)
With the release of Angular 17 and the new Control Flow Syntax, it would be great if there would be a new (recommended) rule, that checks whether the new Control Flow Syntax is used instead of the current/old directives (`NgIf`, `NgFor`, NgSwitch`).

When using Angular 17, this could be part of `recommended` and maybe even have an auto-fix.

json-derulo:
Adds a rule to prefer control flow over the `ngIf`, `ngFor` and `ngSwitch` directives, according to this [rule proposal discussion](https://github.com/angular-eslint/angular-eslint/discussions/1592).

As discussed, no fixer is added, because Angular has their own control flow migration schematics.

Closes #1591


JamesHenry:
Thank you @json-derulo!

PR #1627, opened 2023-11-19
feat(eslint-plugin): add prefer-standalone rule for checking all components, directives and pipes

csvn:
Adds new rules for checking `@Directive` and `@Pipe` for being `standalone` as a companion to `@Component` being checked by the existing rule.

An alternative would be to add options to the current `prefer-standalone-component` rule to allow checking pipes/directives too, but I was afraid it might be confusing with the naming of the existing rule to also apply to directives/pipes. Let me know if think it would be better to merge the code into a single rule (it would likely be less code duplication), or refactor common code into utility functions maybe.


JamesHenry (changes requested):
Hi @csvn I'm sorry for the delay in getting back to you on this one.

I think what might be best overall here is to create a new rule: `prefer-standalone` which applies checks for all the things (as really people should either be doing all or nothing I think).

At the same time we deprecate the `prefer-standalone-component` rule and look to remove it in the next major for clarity

csvn:
I would be fine with that. I'll update this MR soon based on that feedback.

packages/eslint-plugin/src/rules/prefer-standalone.ts
```diff
@@ -0,0 +1,89 @@
...
+    fixable: 'code',
+    schema: [],
+    messages: {
+      preferStandalone: `The {{type}} \`${METADATA_PROPERTY_NAME}\` property should be set to \`${IS_STANDALONE}\``,
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const standaloneRuleFactory =
+      (type: DecoratorTypes) => (node: TSESTree.Decorator) => {
```
csvn, line 33: The contents of this function is the same as for [prefer-standalone-component](https://github.com/angular-eslint/angular-eslint/blob/ee4ebf2ab381164eebf889b131c15f15a93a9693/packages/eslint-plugin/src/rules/prefer-standalone-component.ts#L34), with the addition of turning it into a factory function and passing in the decorator `type` for the messageID data property.

packages/eslint-plugin/tests/rules/prefer-standalone/cases.ts
```diff
@@ -0,0 +1,410 @@
+import { convertAnnotatedSourceToFailureCase } from '@angular-eslint/utils';
+import type { MessageIds } from '../../../src/rules/prefer-standalone';
+
+const messageId: MessageIds = 'preferStandalone';
+
+export const valid = [
```
csvn, line 6: The `valid` & `invalid` test cases here are the same as for `prefer-standalone-component` + new cases for `@Pipe` and `@Directive` (taken from prior commits).

csvn:
@JamesHenry I think I've addressed your feedback now, and that this PR is ready for review. Sorry for it taking a while for me to get back to this one.

JamesHenry:
Thanks so much @csvn!

Issue #1543, opened 2023-09-29
[prefer-ngsrc] Disable the rule if the src is a Base64-encoded string


scurk1415:
**Description and reproduction of the issue**

The `"@angular-eslint/template/prefer-ngsrc": "error"` rule does not check for instances of Base64-encoded string (which are valid src values, but are not supported with the NgOptimizedImage).

When putting the Base64-encoded string into ngSrc we get this runtime error
```
ERROR Error: NG02952: The NgOptimizedImage directive has detected that `ngSrc` is a Base64-encoded string (data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDov...). NgOptimizedImage does not support Base64-encoded strings. To fix this, disable the NgOptimizedImage directive for this element by removing `ngSrc` and using a standard `src` attribute instead.
```

```JSON
{
  "rules": {
    "@angular-eslint/template/prefer-ngsrc": "error"
  }
}
```

I didn't add you library to this reproduction example, as I think it is clear where the issue is
https://stackblitz.com/edit/angular-slnnqs?file=src%2Fmain.ts

**Versions**

| package                                   | version |
| ----------------------------------        | ------- |
| `@angular-eslint/eslint-plugin-template`  | `16.2.0` |
| `@angular-eslint/template-parser`         | `16.2.0` |
| `@typescript-eslint/parser`               | `5.62.0` |
| `ESLint`                                  | `8.49.0` |
| `node`                                    | `18.12.1` |

```sh
Angular CLI: 16.2.3
Node: 18.12.1
Package Manager: npm 7.11.1
OS: win32 x64

Angular: 16.2.6
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1602.3
@angular-devkit/build-angular   16.2.3
@angular-devkit/core            16.2.3
@angular-devkit/schematics      16.2.3
@angular/cli                    16.2.3
@nguniversal/builders           16.2.0
@nguniversal/express-engine     16.2.0
@schematics/angular             16.2.3
rxjs                            7.8.1
typescript                      5.1.6
zone.js                         0.13.3

```

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

JamesHenry:
@scurk1415 would you like to submit a PR for this?

PR #1628, opened 2023-11-20
fix(eslint-plugin-template): [prefer-ngsrc] Do not prefer ngsrc for base64-encoded strings

estevezluis:
Closes https://github.com/angular-eslint/angular-eslint/issues/1543

Rule prefer-ngsrc should not error if image src is a base64-encoded string because NgOptimizedImage does not support Base64-encoded strings.

Expected Behavior:
`` ✅ should not throw an error

` ` ❌ should throw an error because `[src]` is a BoundAttribute

`` ❌ should throw an error because there is two sources.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
+++ b/packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
@@ -1,8 +1,8 @@
 import type {
   TmplAstBoundAttribute,
   TmplAstElement,
-  TmplAstTextAttribute,
 } from '@angular-eslint/bundled-angular-compiler';
+import { TmplAstTextAttribute } from '@angular-eslint/bundled-angular-compiler';
 import { getTemplateParserServices } from '@angular-eslint/utils';
 import { createESLintRule } from '../utils/create-eslint-rule';
 
@@ -34,7 +34,10 @@ export default createESLintRule<Options, MessageIds>({
         const ngSrcAttribute = hasNgSrcAttribute(element);
         const srcAttribute = hasNormalSrcAttribute(element);
 
-        if (!srcAttribute) {
+        if (
+          !srcAttribute ||
+          (!ngSrcAttribute && isSrcBase64Image(srcAttribute))
+        ) {
           return;
         }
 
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
```diff
@@ -66,3 +69,13 @@ function hasNormalSrcAttribute({
 }: TmplAstElement): TmplAstTextAttribute | TmplAstBoundAttribute | undefined {
   return [...inputs, ...attributes].find(({ name }) => name === 'src');
 }
+
+function isSrcBase64Image(
+  attribute: TmplAstTextAttribute | TmplAstBoundAttribute,
+) {
+  if (attribute instanceof TmplAstTextAttribute) {
+    return attribute.value.trim().startsWith('data:');
```
abaran30, line 77: I'm not an expert on Base64, but would it be worth adding a check for the value to also contain `;base64`?

estevezluis: I apologize for the lack of details! This fix follows angular's assertion[^1] that ngSrc does not contain `data:`. So it seems that not only is base64 not allowed but any data URI is not allowed, including svg that are **NOT** base64 encoded.

Example[^2]:
```<circle r='50' cx='50' cy='50' fill='red'/></svg>">```

Reference:
[^1]: https://github.com/angular/angular/blob/17.0.3/packages/common/src/directives/ng_optimized_image/ng_optimized_image.ts#L585

[^2]: https://stackblitz.com/edit/angular-qrkion?file=src%2Fmain.ts

abaran30: Thank you for clarifying. LGTM, though there is a CI failure [here](https://github.com/angular-eslint/angular-eslint/actions/runs/6924924361/job/18854030056?pr=1628#step:10:490), if you could please take a look.

Also, if this PR fixes #1543 , could you please update the PR description to automatically close the issue upon merge? See [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

abaran30 (approved):

abaran30:
@estevezluis I'm unsure why codecov isn't running (maybe CI flake). Would you be able to make one more commit to refresh the CI run? Maybe add a comment to the `isSrcBase64Image` function explaining that the logic is based on Angular's assertion (with a link?) for future reference.

abaran30 (approved):

PR #1630, opened 2023-11-20
fix(eslint-plugin): fix max inline declarations rule for styles string

csvn:
This fix is for `component-max-inline-declarations` rule to also apply for `styles` with a `string`:

```ts
@Component({
  styles: `
    :host { display: block; }
  `
})
class Foo {}
```


JamesHenry:
Thank you @csvn!

PR #1656, opened 2023-12-01
fix(eslint-plugin-template): fix control flow syntax with i18n rule

### LINKED ISSUE #1640: [template/i18n] doesn't work with control flow syntax

```JSON
{
  "rules": {
    "@angular-eslint/template/i18n": ["error", {
      checkId: true,
      checkAttributes: false,
      checkDuplicateId: false,
      ignoreTags: [':svg:title', ':svg:desc', ':svg:tspan'],
    }]
  }
}
```

```HTML
<div
  i18n="@@foo"
>
  Name: @if (name) {
  <span>{{ name }}</span>
  }
</div>
```

The i18n rule should walk up past the control flow syntax and see the i18n tag on the parent div element. Instead it reports on the span element that an `i18n` attribute is required.

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `17.1.0` |
| `@angular-eslint/template-parser`         | `17.1.0` |
| `@typescript-eslint/parser`               | `6.11.0` |
| `ESLint`                                  | `8.46.0` |
| `node`                                    | `18.17.1` |

```sh
Angular CLI: 17.0.3
Node: 18.17.1
Package Manager: pnpm 8.10.5
OS: darwin arm64

Angular: 17.0.4
... animations, common, compiler, compiler-cli, core, elements
... forms, language-service, localize, platform-browser
... platform-browser-dynamic, router, service-worker

@angular-devkit/architect       0.1700.1
@angular-devkit/build-angular   17.0.3
@angular-devkit/core            17.0.3
@angular-devkit/schematics      17.0.3
@angular/cdk                    17.0.1
@angular/cli                    17.0.3
@angular/google-maps            15.2.9
@schematics/angular             17.0.3
ng-packagr                      17.0.2
rxjs                            7.8.0
typescript                      5.2.2
webpack                         5.88.2
zone.js                         0.14.2
```

mattlewis92:
Adds support for control flow block syntax inside the i18n rule. Currently using elements inside control flow blocks will result in a lint error, even if there is a parent element that has the i18n attribute.

Fixes #1640


packages/template-parser/src/index.ts
```diff
@@ -49,10 +49,11 @@ const KEYS: VisitorKeys = {
   PropertyRead: ['receiver'],
   Template: ['templateAttrs', 'children', 'inputs'],
   BindingPipe: ['exp'],
-  DeferredBlock: ['children'],
+  DeferredBlock: ['children', 'placeholder', 'loading', 'error'],
```
mattlewis92, line 52: Without this angular-eslint won't parse `@placeholder`, `@loading` or `@error` nodes

packages/template-parser/src/index.ts
```diff
@@ -99,6 +100,9 @@ function preprocessNode(node: Node) {
 
   for (i = 0; i < keys.length; ++i) {
     const child = node[keys[i]];
+    if (child == null) {
```
mattlewis92, line 103: Without this defer block parsing fails with:
```
● parseForESLint() › @defer › should support the different variants of @defer

    TypeError: Cannot read properties of null (reading 'type')

      105 |     }
      106 |     const isArr = Array.isArray(child);
    > 107 |     if (child.type !== undefined) {
          |               ^
      108 |       // Angular sometimes uses a prop called type already
      109 |       child.__originalType = child.type;
      110 |     }

      at preprocessNode (src/index.ts:107:15)
      at preprocessNode (src/index.ts:129:11)
      at parseForESLint (src/index.ts:272:3)
      at Object.<anonymous> (tests/index.test.ts:559:23)
```

Maybe I got something wrong in the visitor keys?

JamesHenry:
Thanks @mattlewis92!

PR #1257, opened 2022-12-11
fix(eslint-plugin-template): [i18n] handle ng-template properly

mattlewis92:
Rebased version of https://github.com/angular-eslint/angular-eslint/pull/702 with the only difference being that I left off the broken and not working fix for i18n attributes on `<ng-template>` tags (which is really a separate bug from the original issue reported anyway)

Fixes #701

PR #1662, opened 2023-12-06
fix(eslint-plugin-template): handle i18n tags on structural direcives

mattlewis92:
Fixes a regression caused by https://github.com/angular-eslint/angular-eslint/pull/1257 where i18n attributes on structural directives are not propertly detected. The fix is to tighten the `isTemplate` helper to only check for `ng-template` nodes and not all template nodes which are created by using a structural directive.

First commit is from https://github.com/angular-eslint/angular-eslint/pull/1656 which this PR builds on as it requires the fix from that PR as well.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/i18n.ts
+++ b/packages/eslint-plugin-template/src/rules/i18n.ts
@@ -232,7 +232,7 @@ export default createESLintRule<Options, MessageIds>({
       node: StronglyTypedI18n<TmplAstNode> | StronglyTypedBoundTextOrIcuOrText,
     ): AST | undefined {
       const parent = node.parent;
-      if (parent && !isElement(parent) && !isTemplate(parent)) {
+      if (parent && !isElement(parent) && !isNgTemplate(parent)) {
         return getNextElementOrTemplateParent(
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/i18n.ts
```diff
@@ -522,26 +522,26 @@ function isBoundTextAllowed(
   return !PL_PATTERN.test(text) || allowedBoundTextPattern.test(text);
 }
 
-function isTemplate(ast: unknown): ast is TmplAstTemplate {
-  return ast instanceof TmplAstTemplate;
+function isNgTemplate(ast: unknown): ast is TmplAstTemplate {
+  return ast instanceof TmplAstTemplate && ast.tagName === 'ng-template';
```
mattlewis92, line 526: This is the actual fix, the rest is just a refactor to rename the method to something better.

JamesHenry (changes requested):
Thanks! Ready to be reconciled with latest main

mattlewis92:
> Thanks! Ready to be reconciled with latest main

Done!

JamesHenry (approved):

JamesHenry:
THanks @mattlewis92

PR #1691, opened 2024-01-11
feat(eslint-plugin): [use-lifecycle-interface] add fixer for the rule

bulldog98:
I got the fixer to work, but could not fix one of the tests, seems to be a white space issue.


packages/eslint-plugin/tests/rules/use-lifecycle-interface/cases.ts
```diff
@@ -120,6 +138,20 @@ export const invalid = [
...
+          
+          
+          
+          ngOnInit() {}
+          
+          
+          ngOnDestroy() {}
+          
+        }
+      `,
```
bulldog98, line 154: This is the test case I was unable to get to work

abaran30: Hey @bulldog98, could you try the following?
```ts
    annotatedOutput: `
        @Injectable()
        class Test implements DoBootstrap {
          ngDoBootstrap() {}
                       

          ngOnInit() {}
                  

          ngOnDestroy() {}
          
        }
      `,
```

bulldog98: Yes that does the trick. I updated the PR. Thanks @abaran30

packages/eslint-plugin/tests/rules/use-lifecycle-interface/cases.ts
```diff
@@ -120,6 +138,19 @@ export const invalid = [
         },
       },
     ],
+    annotatedOutput: `
+        @Injectable()
+        class Test implements DoBootstrap {
```
JamesHenry, line 143: This didn't fix all the issues?

packages/eslint-plugin/src/rules/use-lifecycle-interface.ts
```diff
@@ -1,4 +1,7 @@
-import { ASTUtils, toPattern } from '@angular-eslint/utils';
+import {
+  ASTUtils, toPattern,
```
JamesHenry, line 2: You will need to run `yarn format` and commit the result to fix the formatting issues (which is why CI is red)

JamesHenry (changes requested):

JamesHenry:
Thank you @bulldog98! Left some comments

bulldog98:
I'll try to fix the issues, not sure if I have time before the weekend though

packages/eslint-plugin/tests/rules/use-lifecycle-interface/cases.ts
bulldog98: If I understood the documentation of eslint correctly the fixes are applied again and again, as long as the lint error is still reported. I'm unsure, how I can use the plugin in a local testing repo to test that out, please advice me on that.

bulldog98: So yes I found a way to try it out. (run e2e tests and then try your changes in tmp/e2e-fixtures/ in any of the apps there. Maybe that should be documented) Indeed eslint runs again on the output of fixes until there are no changes, so it indeed results in a fix that adds all the implements clauses.

bulldog98: Only thing I noticed is that I do not add an import for the Interface, here I would like guidance @JamesHenry

JamesHenry:
I don't seem to be able to push to the PR, I have added the import fix in a separate commit

JamesHenry:
I think it's because you opened the PR from your main branch maybe?

packages/eslint-plugin/tests/rules/use-lifecycle-interface/cases.ts
JamesHenry: @bulldog98 Sorry for the delay. We have a utility for it used in a few other rules, it's very straightforward: https://github.com/angular-eslint/angular-eslint/pull/1829

JamesHenry: I want to get things ready for Angular 18, so I went ahead and applied the change

JamesHenry:
Thanks again @bulldog98!

bulldog98:
Thanks for adding the import part @JamesHenry, I'll try to keep in mind to open from a separate branch in the future.

PR #1724, opened 2024-02-20
fix: output declaration files in all packages

### LINKED ISSUE #1523: Type declaration files are missing from the distributed packages

It appears that type declaration files are missing in the bundled and published packages, including `@angular-eslint/template-parser`. 

For example, with: https://cdn.jsdelivr.net/npm/@angular-eslint/template-parser@16.1.2/ the `package.json` refers to a type declaration file in `dist/index.d.ts` that does not exist in the bundle. 

The package itself extends the root `tsconfig.base.json` which disables generating type declaration files: https://github.com/angular-eslint/angular-eslint/blob/main/tsconfig.base.json#L6

This results in errors when using the package in typescript environments and prevents any type detection on the public APIs.

Is this intended?

**Versions**

| `@angular-eslint/template-parser`  | `16.1.2` |
| `ESLint`                           | `8.0.0` |
| `node`                             | `18.0.0` |

I'm not using this package in an Angular workspace, but a typescript project that needs to parse Angular component templates. 

joeyj:
Resolves #1523

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -3,7 +3,7 @@
     "allowSyntheticDefaultImports": true,
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
-    "declaration": false,
+    "declaration": true,
     "declarationMap": false,
     "esModuleInterop": true,
     "module": "CommonJS",


JamesHenry:
Thanks @joeyj!

PR #1757, opened 2024-04-09
docs: fix incorrect link

undsoft:
Fixes incorrect link at the bottom of this page: https://github.com/angular-eslint/angular-eslint/tree/main/packages/eslint-plugin


JamesHenry:
Thanks for identifying this @undsoft. These are autogenerated, so this isn't the correct fix, but it allowed me to discover that the rule lists are not being verified in CI.

I have fixed the issue and applied the check in CI here, so it shouldn't happen again: #1764

Issue #1759, opened 2024-04-12
[label-has-associated-control] ids in the for attribute are not checked for existence


air2:
The following code should error, but does not:

```
<label for="id">Label</label>
<input id="otherId" />

```
```JSON
{
  "rules": {
    "@angular-eslint/template/label-has-associated-control": ["error"]
  }
}
```

```HTML
<label for="id">Label</label>
<input id="otherId" />
```

**Versions**

| package                                   | version |
| ----------------------------------        | ------- |
| `@angular-eslint/eslint-plugin-template`  | `17.3.0` |
| `@angular-eslint/template-parser`         | `17.3.0` |
| `@typescript-eslint/parser`               | `7.4.0` |
| `ESLint`                                  | `8.57.0` |
| `node`                                    | `18.19.0` |

```sh
# Please run `npx ng version` in your project and paste the full output here:
Angular: 17.3.3
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1703.3
@angular-devkit/build-angular   17.3.3
@angular-devkit/core            17.3.3
@angular-devkit/schematics      17.3.3
@schematics/angular             17.3.3
rxjs                            7.8.1
typescript                      5.4.3
zone.js                         0.14.4

```

- [X ] I have tried restarting my IDE and the issue persists.
- [ X] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

PR #1760, opened 2024-04-12
[accessibility-label-has-associated-control] check id's in the for attribute of a label for existence

air2:
This should fix issue #1759


air2:
I saw my code was bades on a fork of somebody else, really do not know how that happend, I created a new pull request with  just my changes

PR #1761, opened 2024-04-12
fix(eslint-plugin-template): [label-has-associated-control] check id's in the for attribute of a label for existence

air2:
This should fix issue #1759


JamesHenry (changes requested):
Thanks @air2, please run `yarn update-rule-docs` and commit the result

air2:
@JamesHenry   Sorry I had some other projects lately, so I had not time to get back at this, before now, I see it is already, merged, thank you! And sorry for my late reply and not picking your request up sooner

PR #1772, opened 2024-04-16
fix(eslint-plugin): remove `no-host-metadata-property` from the recommended list.

JeanMeche:
As part of the v18 release  and as discussed in angular/angular#53888 & angular/angular#54284, the styling guide will now recommend using the host property instead of the `@HostBinding`/`@HostListener`.

The change reflect in the recommended rule list reflect this change.

### DIFF
--- a/packages/eslint-plugin/src/configs/recommended.json
+++ b/packages/eslint-plugin/src/configs/recommended.json
@@ -1,12 +1,13 @@
 {
   "parser": "@typescript-eslint/parser",
-  "plugins": ["@angular-eslint"],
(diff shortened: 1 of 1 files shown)


JamesHenry:
Thank you @JeanMeche!

We only make changes to recommended in major versions, so I'll be sure to get this in v18

JamesHenry:
Thanks again @JeanMeche, superseded by https://github.com/angular-eslint/angular-eslint/pull/1839 (you updated a generated file here, and we also need to formally deprecate the rule)

PR #1779, opened 2024-04-18
feat(eslint-plugin): [no-duplicates-in-metadata-arrays] new rule

### LINKED ISSUE #1754: Is there a linting role for detecting duplicated imports?
for example:

```
@Compoennt({
   imports: [CommonModule, SomeOtherModule, CommonModule.....]
})
```

abaran30:
This pull request introduces a new rule `no-duplicates-in-metadata-arrays`. The rule will check all key metadata arrays for `NgModule`, `Component`, and `Directive`. 

Fixes #1754


packages/eslint-plugin/src/configs/recommended.json
```diff
@@ -5,6 +5,7 @@
     "@angular-eslint/component-class-suffix": "error",
     "@angular-eslint/contextual-lifecycle": "error",
     "@angular-eslint/directive-class-suffix": "error",
+    "@angular-eslint/no-duplicates-in-metadata-arrays": "error",
```
JamesHenry, line 8: Please revert, we don't make changes to recommended configs outside of major versions.

Related: did you manually add this? Because I don't see any recommended metadata on the rule? If we are generating here by default then that's a bug in the generation script IMO

JamesHenry (changes requested):

packages/eslint-plugin/src/configs/recommended.json
abaran30: Reverted. I must've done this manually forgetting that we don't do this outside of major versions. I ran `update-rule-lists` and it did not update this file.

PR #1781, opened 2024-04-19
chore(eslint-plugin): [sort-ngmodule-metadata-arrays] deprecate rule

### LINKED ISSUE #1232: Improving/replacing `sort-ngmodule-metadata-arrays`

abaran30:
This pull request deprecates rule `sort-ngmodule-metadata-arrays`; see discussions in #1232.

Because there is no replacement for this rule, this pull request also updates the rule doc generator to not include a rule replacement note if one is not available.

Closes #1232


JamesHenry (approved):

JamesHenry:
FYI updated to fix so that it shows up in the release notes

PR #1795, opened 2024-05-01
feat(eslint-plugin-template): [i18n] add allowMarkupInContent option

### LINKED ISSUE #930: [i18n] option to check content of tag with i18n
I would like to propose that you include an option to the i18n rule.

This option should make the rule check the content of a tag that has the i18n attribute set, and report if the content is not plain text (eg. it contains markup).

I recently discovered that a misplaced i18n attribute caused a bug in an application, because the markup content in the tag was replaced with plain text.

I try to enforce the convention that markup (and application logic) should not be mixed with i18n, and I would like this rule to help me with that. Therefore it is also my opinion that this new option should be enabled by default.

```JSON
{
  "rules": {
    "@angular-eslint/template/i18n": ["error"]
  }
}
```
OR
```JSON
{
  "rules": {
    "@angular-eslint/template/i18n": ["error",{"newOptionName": "true"}]
  }
}
```

**Invalid code for this rule**
```HTML
<div i18n="@@mainHeader">
  <h1>Hello i18n!</h1>
</div>
```
OR
```HTML
<!-- this was changed from a text-button to an icon with a tooltip, but we forgot to remove the i18n for the button text -->
<div i18n="@@buttonText" i18n-title="@@buttonTooltip" title="Enable magic">
  <fa-icon [icon]="['fas', 'magic']"></fa-icon>
</div>
```

**Valid code for this rule**
```HTML
<div>
  <h1 i18n="@@mainHeader">Hello i18n!</h1>
</div>
```
OR
```HTML
<div i18n-title="@@buttonTooltip" title="Enable magic">
  <fa-icon [icon]="['fas', 'magic']"></fa-icon>
</div>
```

m-akinc:
Fixes #930

Added `allowMarkupInContent` option for `i18n` rule. Option defaults to `true` to maintain backwards compatibility. Setting to `false` causes rule violation whenever an element with the `i18n` attribute has HTML markup in its content.

Note that due to the way the AST parser handles ICUs (i.e. it escapes any markup in cases so that they are strings), we cannot report violations for markup inside the ICU cases.


m-akinc:
@JamesHenry There haven't been any comments or review since I put this up a couple weeks ago. Just wanted to ping someone to make sure it isn't overlooked.

JamesHenry (changes requested):
Thanks @m-akinc!

I would prefer it if no existing tests were modified by this PR to prove that this change is backwards compatible. Please can you make the necessary updates?

JamesHenry (approved):

JamesHenry:
Thank you @m-akinc!

Issue #1446, opened 2023-07-06
[@angular-eslint/no-input-rename] incorectly triggers on input transforms


kbrilla:
On Angular 16.1.3 
```
@Input({ transform: (val) => val ?? 'default') }) input!: string;
```
string `'default` will cause to trigger @angular-eslint/no-input-rename

It seems that rule only looks for string literals as this does not trigger the rule
```
@Input({ transform: (val) => val) }) input!: string;
```

ptu14:
same problem here prosze to naprawić

orestisioakeimidis:
Same problem for me. The following
```ts
@Input({ transform: (value: any) => value || [10, 25, 50, 100] })
options: number[];
```
fails on the numbers with:
```
Input bindings should not be aliased
```

abaran30:
Confirmed that with the mentioned examples, `'default'` and `[10, 25, 50, 100]` are getting picked up as literals, resulting in an incorrect reporting.

I added the **PRs Welcome** label in case anyone would like to attempt a fix. One idea is to check if there is a metadata object, and if the object does not contain property `alias`, no need to continue with the rest of the logic.

Something like the following:
```ts
        const inputCallExpression = ASTUtils.getNearestNodeFrom(
          node,
          ASTUtils.isCallExpression,
        );

        // Get the Input metadata object if available
        if (
          inputCallExpression &&
          TSESLintASTUtils.isIdentifier(inputCallExpression.callee) &&
          inputCallExpression.callee.name === 'Input' &&
          inputCallExpression.arguments[0] &&
          ASTUtils.isObjectExpression(inputCallExpression.arguments[0])
        ) {
          // Search for `alias` property in the object
          const aliasProperty =
            inputCallExpression.arguments[0].properties.find((property) => {
              if (
                ASTUtils.isProperty(property) &&
                ASTUtils.getRawText(property.key) === 'alias'
              ) {
                return property;
              }

              return undefined;
            });

          // No `alias` property found, no need to continue
          if (!aliasProperty) {
            return;
          }
        }
```
However, in my opinion, the real challenge then is adjusting the fixer so that the `alias` node, as well as any commas that came with it, get removed correctly so that the resulting source is valid. Manipulating objects via fixer can get tricky...

rala72:
same problem for `required`

LayZeeDK:
Also observed when the empty string (`''`) is used in an input transformer.

PR #1809, opened 2024-05-07
fix(eslint-plugin): [no-input-rename] incorrectly triggers on input transforms

ptu14:
closes https://github.com/angular-eslint/angular-eslint/issues/1446


JamesHenry (approved):
Thanks a lot @ptu14!

Issue #1784, opened 2024-04-22
Unexpected 9MB binary produced from `@angular-eslint/schematics`: `.nx/cache/18.3.3-nx.linux-x64-gnu.node`


johnnygerard:
A ~9MB binary is generated from `@angular-eslint/schematics@17.3.0`.
I'm assuming it is a regression bug maybe after your migration to Nx 18.

To reproduce:

```sh
ng new app --defaults
cd app
ng add @angular-eslint/schematics --skip-confirmation
```

You should see this file: `app/.nx/cache/18.3.3-nx.linux-x64-gnu.node`.

**Environment:**

```txt
     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 17.3.5
Node: 20.12.2
Package Manager: npm 10.5.2
OS: linux x64

Angular: 17.3.5
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1703.5
@angular-devkit/build-angular   17.3.5
@angular-devkit/core            17.3.5
@angular-devkit/schematics      17.3.5
@schematics/angular             17.3.5
rxjs                            7.8.1
typescript                      5.4.5
zone.js                         0.14.4
```

**Added dependencies:**
```txt
    "@angular-eslint/builder": "17.3.0",
    "@angular-eslint/eslint-plugin": "17.3.0",
    "@angular-eslint/eslint-plugin-template": "17.3.0",
    "@angular-eslint/schematics": "17.3.0",
    "@angular-eslint/template-parser": "17.3.0",

    "@typescript-eslint/eslint-plugin": "7.2.0",
    "@typescript-eslint/parser": "7.2.0",
    "eslint": "^8.57.0",
```

robertIsaac:
I think eslint should add `.nx` to `.gitignore`

MonsieurBon:
[...]

PR #1815, opened 2024-05-14
refactor: Use `nx` only at build time

ranma42:
and rely on `@angular-devkit` at runtime.

### DIFF
--- a/packages/builder/package.json
+++ b/packages/builder/package.json
@@ -18,10 +18,6 @@
     "LICENSE",
     "builders.json"
   ],
-  "dependencies": {
-    "@nx/devkit": "^17.2.8 || ^18.0.0",
-    "nx": "^17.2.8 || ^18.0.0"
-  },
   "builders": "./builders.json",
   "peerDependencies": {
     "eslint": "^7.20.0 || ^8.0.0",
--- a/packages/builder/src/lint.impl.ts
+++ b/packages/builder/src/lint.impl.ts
@@ -1,25 +1,21 @@
 import type { BuilderOutput } from '@angular-devkit/architect';
-import {
-  convertNxExecutor,
-  joinPathFragments,
-  workspaceRoot,
-} from '@nx/devkit';
+import { createBuilder } from '@angular-devkit/architect';
 import type { ESLint } from 'eslint';
 import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { dirname, join, resolve } from 'path';
 import type { Schema } from './schema';
 import { resolveAndInstantiateESLint } from './utils/eslint-utils';
 
-export default convertNxExecutor(
+export default createBuilder(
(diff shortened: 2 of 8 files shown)


JamesHenry:
Thanks @ranma42 but I am not looking to rip out Nx at runtime, it's still an important use case for the tool.

The current issue of a tracked native file is being addressed in two ways:
- I will be releasing a patch powered by #1816 which will apply the relevant gitignore config automatically on `ng add` for this temporary situation where the native file is placed in the workspace
- Per my discussion with the Nx team, we are updating it so that it no longer puts the native file there https://github.com/nrwl/nx/pull/23375

Many thanks!

ranma42:
> Thanks @ranma42 but I am not looking to rip out Nx at runtime, it's still an important use case for the tool.

This PR should not adversely affect the NX use case (NX projects should be able to use angular-eslint just fine even after this change). I might be misunderstanding... what is the use case that would encounter problems with this change?

> The current issue of a tracked native file is being addressed in two ways:
> 
> * I will be releasing a patch powered by [fix(schematics): ignore nx cache #1816](https://github.com/angular-eslint/angular-eslint/pull/1816) which will apply the relevant gitignore config automatically on `ng add` for this temporary situation where the native file is placed in the workspace
> * Per my discussion with the Nx team, we are updating it so that it no longer puts the native file there [fix(core): copy native files to tmp file location instead of .nx/cache nrwl/nx#23375](https://github.com/nrwl/nx/pull/23375)
> 
> Many thanks!

The `nx` runtime dependency also introduces additional inconveniences, as the consumer project might end up with conflicting nx versions, more dependency bloat and in general additional "risks" (as seen in #1784).

09f242ee1597bcbb04fe998359c8c0660bba4a93 seems particularly trivial (in fact, it simply restores a previous version of the code; I checked it and AFAICT it is doing exactly the same operations, just with a slightly different syntax).
b303c40caecdf27f046f41c95b78b38b92e37e7c is definitely more risky, but does not seem to make the code much more complex.

If you are worried about the additional maintenance burden (or for whatever reason prefer the convenience of using `nx` also for implementing the schematics and builder), how would you see a `angular-eslint-slim` package?
It could be managed as a fork or even a branch within this repo... I would be happy to volunteer the effort to try and maintain that with git-based autosyncing + manual fixes as needed, but I would like to avoid doing so if it is considered "adversarial".

JamesHenry:
Hi @ranma42, just to let you know  I appreciate the feedback here, and I did decide to remove Nx from the runtime of angular-eslint and just leave it to enhancing my development time concerns in the repo, so there is no direct dependency in the `@angular-eslint/builder` and `@angular-eslint/schematics` any longer as of `18.3.0`, and `nx` and `@nx/devkit` will only be pulled into your workspaces if you choose to use it as your own dev tool (which I naturally highly recommend 😄)

Thanks again

PR #1849, opened 2024-05-29
docs: Removed some unnecessary characters from interactive-supports-focus.md

DerZyklop:


JamesHenry (changes requested):
These docs are autogenerated from the unit tests.

These characters are in the unit test cases, so the resolution here would be to update `packages/eslint-plugin-template/tests/rules/interactive-supports-focus/cases.ts` and then run `yarn update-rule-docs` and commit the result.

JamesHenry:
As I haven't heard back I have applied the change per my notes in #1903

Issue #1857, opened 2024-05-30
[prefer-self-closing-tags] <app-root /> conflicts with angular-cli build


joergbaier:
**Description and reproduction of the issue**

This linter rule will change `<app-root></app-root>` in `index.html` to be a self closing tag.

However `@angular/build` v18.0.0 will throw a build error
```
⠋ Generating index html...Invalid self-closing element in index HTML file: '<app-root />'.
✖ Index html generation failed.
```
https://github.com/angular/angular-cli/releases/tag/18.0.0
https://github.com/angular/angular-cli/commit/70dbc7a6e9a7f6d55aeb4e10e8e686b186e6cdf3

joergbaier:
There was this comment in the past, but I imagine more people will face this issue as they upgrade to 18
https://github.com/angular-eslint/angular-eslint/issues/1663#issuecomment-1851517064

JamesHenry:
@joergbaier I would accept a PR to specifically ignore the filename ending in `src/index.html` from consideration within that rule if we know it is never applicable

joergbaier:
https://github.com/angular-eslint/angular-eslint/pull/1865

vitaly-phntm:
@joergbaier @JamesHenry Angular supports having different folder structure, not just `src/index.html` e.g. `frontend/index.html`. This fix doesn't work for it.

JamesHenry:
@vitaly-phntm there isn’t anything wrong with using an eslint-disable comment for that specific rule in that specific location, you could alternatively add an override in config which disables the rule for your relevant file location. This was just meant to be a convenience for the common case

PR #1865, opened 2024-06-06
fix(eslint-plugin-template): [prefer-self-closing-tags] always ignore index.html files

Squixx:
Ignores index.html files for prefer-self-closing-tags  as described in #1857


JamesHenry (changes requested):
You should be able to add a unit test for this by setting `filename` please

Squixx:
added!

JamesHenry:
Test looks good, please run `yarn update-rule-docs` and commit the result, the docs are generated from the unit tests so whenever they change the docs need to too

Squixx:
i've updated the docs. However currently the result doesn't make sense in this context. Is there a ways to simply annotate a rule output? e.g. `src/index.html` is excluded as this would prevent angular from bootstrapping

JamesHenry (approved):

JamesHenry:
Thanks so much @Squixx!

Squixx:
Thanks for guiding me through the repo.

silvanadrian:
@JamesHenry Is there a plan on getting this fix released anytime soon?

JamesHenry:
image

https://github.com/angular-eslint/angular-eslint?tab=readme-ov-file#package-versions

PR #1878, opened 2024-06-11
fix(test-utils): use configured RuleTester.afterAll over global

csvn:
The `RuleTester` class from `@typescript-eslint/rule-tester` can be extended for various testing frameworks by updating `RuleTester.describe`, `RuleTester.afterAll`, etc. But the Angular one is hard-coded to a `globalThis.afterAll`, and ignores `RuleTester.afterAll`. This PR will make `@angular-eslint/test-utils` respect `RuleTester.afterAll`, by using that first if it exists.

image


JamesHenry (approved):
Thank you, @csvn!

Issue #1567, opened 2023-11-02
[@angular-eslint/template/prefer-self-closing-tags] `ng-content` should be marked as self-closing tag


dzhavat:
Hey 👋

Thanks for your great work!

I recently started using the `prefer-self-closing-tags` rule in a project, and noticed that the `ng-content` element was not marked as error when using the "full" syntax.

**Description and reproduction of the issue**

```JSON
{
  "rules": {
    "@angular-eslint/template/prefer-self-closing-tags": ["error"]
  }
}
```

```HTML
// your repro code case

<!-- Should be marked as error -->
<ng-content></ng-content>

<!-- Correctly marked as error -->
<ng-container></ng-container>

<!-- Correctly marked as error -->
<router-outlet></router-outlet>
```

**Versions**

| package                                   | version |
| ----------------------------------        | ------- |
| `@angular-eslint/eslint-plugin-template`  | `16.2.0` |
| `@angular-eslint/template-parser`         | `16.2.0` |
| `@typescript-eslint/parser`               | `5.62.0` |
| `ESLint`                                  | `8.46.0` |
| `node`                                    | `18.16.0` |

```sh
# Please run `npx ng version` in your project and paste the full output here:
Angular CLI: 16.2.9
Node: 18.16.0
Package Manager: npm 8.19.3
OS: win32 x64

Angular: 16.2.11
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1602.9
@angular-devkit/build-angular   16.2.9
@angular-devkit/core            16.2.9
@angular-devkit/schematics      16.2.9
@angular/cli                    16.2.9
@schematics/angular             16.2.9
rxjs                            7.8.1
typescript                      5.1.6
zone.js                         0.13.3
```

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

json-derulo:
The same thing happens with the `<ng-template></ng-template>` element.

JamesHenry:
This is addressed in v17 thanks to @json-derulo!

Sergiobop:
I think this is wrong after Angular 18 since we can have default content now inside the ng-content:

```
<ng-content select="header">
      Hi!
</ng-content>
``` 

Transforms into:
```
    <ng-content select="header"> Hi />
[...]

PR #1880, opened 2024-06-12
fix(eslint-plugin-template): [prefer-self-closing-tags] support ng-content with fallback content

json-derulo:
Since Angular 18, `<ng-content>` can be used with a fallback content. The `prefer-self-closing-tags` rule currently raises an error when using this feature, see https://github.com/angular-eslint/angular-eslint/issues/1567#issuecomment-2156430905.


JamesHenry (approved):
Thank you @json-derulo!

PR #1887, opened 2024-06-14
fix(template-parser): visit receiver of Call expression

### LINKED ISSUE #1885: receiver of Call expression is skipped during AST traversal
Due to how the fallback for visitorKeys is handled, ESLint never enters the `receiver` of a `Call` expression. This can lead to various weird issues, for example the `template/no-call-expression` will skip some invalid function calls in the templates:
```html
<!-- with allowList: ['ok'] the call to notOk won't be reported -->
{{ notOk().ok() }}

<!-- nested calls should report multiple errors but produce only one -->
{{ foo().bar() }} {{ foo()() }}
```

The logic of `getFallbackKeys` appears to be buggy in some way - when viewing the nodes in AST explorer or just by printing to `console.log`, I can see that the receiver doesn't have its `type` set, for example.

**Versions**

| `@angular-eslint/template-parser`  | `18.0.1` |
| `ESLint`                           | `9.4.0` |
| `node`                             | `20.9.0` |

```sh
                |___/
    

Angular CLI: 18.0.2
Node: 20.11.1
Package Manager: npm 10.2.4
OS: linux x64

Angular: 18.0.1
... animations, cdk, common, compiler, compiler-cli, core, forms
... language-service, material, platform-browser
... platform-browser-dynamic, router

@angular-devkit/architect       0.1800.2
@angular-devkit/build-angular   18.0.2
@angular-devkit/core            18.0.2
@angular-devkit/schematics      18.0.2
@angular/cli                    18.0.2
@schematics/angular             18.0.1
ng-packagr                      18.0.0
typescript                      5.4.5
zone.js                         0.14.6    
```

P4:
When traversing a Call AST node, ESLint would never visit its receiver/callee. Since Call didn't have visitor keys listed explicitly, the parser would use `getFallbackKeys`, which seems to be buggy as it only returns `args` and not `receiver`.

Update test cases for no-call-expression to deal with new call expressions being reported.

Since convertAnnotatedSourceToFailureCase doesn't deal well with overlapping errors at the moment, instances of `()()` were removed from existing tests and added as separate cases for nested calls.

Closes #1885

### DIFF
--- a/packages/eslint-plugin-template/tests/rules/no-call-expression/cases.ts
+++ b/packages/eslint-plugin-template/tests/rules/no-call-expression/cases.ts
@@ -71,8 +71,8 @@ export const invalid: readonly InvalidTestCase<MessageIds, Options>[] = [
   convertAnnotatedSourceToFailureCase({
     description: 'should fail for `FunctionCall` within `Interpolation`',
     annotatedSource: `
-        <div>{{ getInfo()() }}</div>
-                ~~~~~~~~~~~
+        <div>{{ getInfo() }}</div>
+                ~~~~~~~~~
       `,
     messageId,
   }),
@@ -96,8 +96,8 @@ export const invalid: readonly InvalidTestCase<MessageIds, Options>[] = [
     description:
       'should fail for `FunctionCall`, `MethodCall` and `SafeMethodCall` within `Binary`',
     annotatedSource: `
-        <a [href]="id && createUrl() && test()($any)">info</a>
-                         ~~~~~~~~~~~    ^^^^^^^^^^^^
+        <a [href]="id && createUrl() && test($any)">info</a>
+                         ~~~~~~~~~~~    ^^^^^^^^^^
         {{ id || obj?.nested1() }}
                  ##############
       `,
@@ -113,8 +113,8 @@ export const invalid: readonly InvalidTestCase<MessageIds, Options>[] = [
     annotatedSource: `
         <a [href]="id ? a?.createUrl() : editUrl(3)">info</a>
                         ~~~~~~~~~~~~~~   ^^^^^^^^^^
-        {{ 1 === 2 ? 3 : obj?.nested1()() }}
-                         ################
+        {{ 1 === 2 ? 3 : obj?.nested1() }}
+                         ##############
       `,
     messages: [
       { char: '~', messageId },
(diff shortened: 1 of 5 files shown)


P4:
Hi, any info on this? I've updated the PR after `nx run eslint-plugin-template:check-rule-docs` failed to include the updated docs by running `yarn update-rule-docs` as instructed, and it's been sitting like this since

JamesHenry:
@P4 rule tests are now more strictly typed after a large scale refactoring was needed to bring us up to the lates typescript-eslint v8 stable in #1956

The latest CI run fails on this one and it looks to be related to the stricter types, please could you take a look?

P4:
I've rebased the changes on top of the most recent `main` and  slightly changed my workaround for overlapping errors to make it pass typechecking

packages/eslint-plugin-template/tests/rules/no-call-expression/cases.ts
```diff
@@ -228,4 +228,61 @@ export const invalid: readonly InvalidTestCase<MessageIds, Options>[] = [
...
+    annotatedSource: `
+    {{ notOk().ok() }}
+       ~~~~~~~
+    `,
+    options: [{ allowList: ['ok'] }],
+    messages: [{ char: '~', messageId }],
+  }),
 ];
+
+// convertAnnotatedSourceToFailureCase unfortunately cannot handle multiple errors starting/ending at the same position;
```
JamesHenry, line 281: It's our own utility, so feels a little weird to add a comment saying "unfortunately cannot handle". Can we make it handle it? 🤔

JamesHenry (changes requested):
Once we address the point around the error utility this is good to go, thanks so much!

packages/eslint-plugin-template/tests/rules/no-call-expression/cases.ts
P4: I looked at the implementation, but only briefly, I didn't really want this PR to turn into a side-quest of re-writing a test utility to handle just these few cases. 

If you feel that it's worthwhile, I can try and implement the necessary changes.

JamesHenry: It's a unit test utility right, so a super fast feedback loop - if the existing tests and the new tests pass then the patch is "correct" :)

If it bogs you down for a long time then let me know, but it's definitely worth giving it a go

P4: Alright, I think I've managed to finally figure it out - this hack is no longer necessary and the other tests didn't break.

The code for processing annotated cases assumed the code would always be at `line-1`, but this is not true if we have multiple lines of annotations, I changed it to instead keep track of the last non-annotation line and use that for setting start and end line positions

JamesHenry (approved):
Thanks a lot, @P4!

JamesHenry:
@P4 I'm afraid I'm not sure what happened with this one, but the tests are _not_ passing. Please can you recreate the PR and we can take a look?

PR #1897, opened 2024-06-22
feat(eslint-plugin): [no-call-expression] add allowPrefix and allowSuffix

MaximilianMain:
Currently the usage of Signals in templates will conflict with the rule noCallExpression. 
If you manage to use a prefix or suffix for your signals in your project, it is possible to define allow definitions for them.


JamesHenry (changes requested):
Thanks @MaximilianMain, this seems reasonable as an opt in.

Our rule docs are generated from implementations and unit tests, so you will need to run

```sh
yarn update-rule-docs
```

And commit the result for this PR to pass CI

MaximilianMain:
Thanks @JamesHenry, I have executed yarn update-rule-docs already 
 NX   Successfully ran target update-rule-docs for 2 projects and 4 tasks they depend on (14s)
as a result the documentation for no-call-expression.md has been changed

JamesHenry:
Thanks @MaximilianMain looks like the formatting check failed - `yarn format` will fix it

JamesHenry:
I'm not a fun of the blockless `if`s that were used in the original implementation so I have update those and fixed the formatting

MaximilianMain:
Thanks for the merge, i am also not a fan of the blockless if conditions, but the first condition was already implemented this way ;)

JamesHenry:
Yeah I know, sorry for not being clearer, that's what I meant by original implementation

PR #1898, opened 2024-06-27
feat(eslint-plugin): new rule runtime-localize

m-akinc:
For applications that rely on runtime translation of strings (via [`loadTranslations`](https://angular.dev/api/localize/loadTranslations)), `$localize`-tagged strings must not be evaluated until after translations have been loaded. For example, if a file contains a declaration like ``const myString = $localize`foo`;`` (at the top-level), it will be evaluated when the module is loaded, i.e. before the app has had a chance to call `loadTranslations`. Safe places to evaluate a  `$localize`-tagged string are:
1. within a function/method body
2. in a function parameter default value
3. in a non-static class member initializer
All of these defer evaluation until a function runs or a class is constructed.

Even if an app calls `loadTranslations` as part of an `APP_INITIALIZER` provider, some modules may be loaded before that (e.g. modules referenced by the `AppModule`'s `@NgModule` decorator). Therefore, the safest practice is to only use `$localize` in functions and non-static class members.

Note that calling a function that includes `$localize` at the top-level of a file is a case that should probably be reported as an error, but it's not technically practical. I.e.
```ts
function getTranslatedFoo() { return $localize`foo`; }
const translatedFoo = getTranslatedFoo();
```


m-akinc:
Looks like the CI is currently broken. 💥🔥

JamesHenry:
@m-akinc CI is appropriately catching that you added a rule but did not include it in the configs exposed by the plugin.

```sh
yarn nx update-rule-configs
```

Is the command you want for this.

I have run out of time today, but I will have a look at your PR soon

m-akinc:
> @m-akinc CI is appropriately catching that you added a rule but did not include it in the configs exposed by the plugin.

Oh, I didn't notice that error. Thanks, I'll get that fixed. The error that made me think the CI was broken, though, was this one that failed the "Primary" job:
```
Error: Command failed: git diff --name-only HEAD~1 HEAD
fatal: ambiguous argument 'HEAD~1': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Let's see if I still run into it on the next run.

m-akinc:
@JamesHenry just a ping, in case this was forgotten. No rush, though.

m-akinc:
@JamesHenry another ping

JamesHenry (changes requested):
I'm sorry for the delay on this one @m-akinc, I've pulled in the latest changes into the branch.

Please can you align the test file to match the conventions now applied to the rest of the repo in terms of exporting Options from the rule implementation and specifying the types of `valid` and `invalid` in the tests.

Once those changes are made we can get this merged!

JamesHenry:
@m-akinc Please also remove the argument to `RuleTester`.

In flat config, parsers are passed by reference, nested within `languageOptions.parser`, not as a string via a top level `parser` property. `@typescript-eslint/parser` is the default for our `RuleTester`, so rather than fixing the passed config you can omit it entirely: `new RuleTester();`

JamesHenry (approved):

JamesHenry:
Thanks again @m-akinc!

PR #1922, opened 2024-07-25
fix(eslint-plugin-template): [interactive-supports-focus] allowList with form as default option to support event bubbling

### LINKED ISSUE #1429: [@angular-eslint/template/interactive-supports-focus] Forms are erroneously marked as unfocusable

The rule erroneously marks the following form element as "Elements with interaction handlers must be focusable."

```HTML
<form (ngSubmit)="submit()" (keydown.control.enter)="submit()">
...
```

This is wrong because the handler is perfectly useful for submitting the form when **focus is in any of its controls**. The form itself is not required to have a tabindex, and it isn't meant to ever be focused itself.

In my opinion, this rule should check not only the element containing the handler, but all child elements as well, and defer if any of those is focusable. Keyboard events are [bubbling](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) by default:

> Keyboard events are only generated by `<input>`, `<textarea>`, `<summary>` and anything with the `contentEditable` or `tabindex` attribute. If not caught, they [bubble](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#event_bubbling) up the [DOM tree](https://developer.mozilla.org/en-US/docs/Web/API/Document_object_model/Using_the_Document_Object_Model#what_is_a_dom_tree) until they reach [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document).

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `16.0.3` |
| `@angular-eslint/template-parser`         | `16.0.3` |
| `@typescript-eslint/parser`               | `5.60.0` |
| `ESLint`                                  | `8.43.0` |
| `node`                                    | `18.16.0` |

```sh
Angular CLI: 16.1.0
Node: 18.16.0
Package Manager: npm 9.5.1
OS: linux x64

Angular: 16.1.2
... animations, common, compiler, core, forms, localize

@angular-devkit/architect       0.1601.1
@angular-devkit/build-angular   16.1.1
@angular-devkit/core            16.1.0
@angular-devkit/schematics      16.1.0
@angular/cli                    16.1.0
@angular/compiler-cli           16.1.0
@schematics/angular             16.1.0
```

sandikbarr:
Resolves #1429 

Adds an `allowList` to the interactive-supports-focus template rule to with `<form>` as a default option to allow for event bubbling and capturing events from `<form>` elements. 
* The `allowList` can be configured as an empty array to more strictly permit click and key events on interactive elements only
* Additional HTML or custom elements can be configured and added to the `allowList` array to permit click and key events on elements that are otherwise not interactive

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/interactive-supports-focus.md
+++ b/packages/eslint-plugin-template/docs/rules/interactive-supports-focus.md
@@ -913,6 +913,32 @@ interface Options {
 #### ✅ Valid Code
 
 ```html
+<test-component (keydown)="onKeyDown()"></test-component>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/interactive-supports-focus": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```html
 <form (keydown)="onKeyDown()"></form>
 ```
 
--- a/packages/eslint-plugin-template/docs/rules/interactive-supports-focus.md
+++ b/packages/eslint-plugin-template/docs/rules/interactive-supports-focus.md
@@ -23,7 +23,17 @@
 
 ## Rule Options
 
-The rule does not have any configuration options.
+The rule accepts an options object with the following properties:
+
+```ts
+interface Options {
+  /**
+   * Default: `["form"]`
+   */
+  allowList?: string[];
+}
+
+```
 
 <br>
 
@@ -356,6 +366,36 @@ The rule does not have any configuration options.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ```
 
+<br>
+
+---
+
+<br>
+
+#### Custom Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/interactive-supports-focus": [
+      "error",
+      {
+        "allowList": []
+      }
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```html
+<form (keydown)="onKeyDown()"></form>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+```
+
 </details>
 
 <br>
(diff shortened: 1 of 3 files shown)


abaran30:
Changes LGTM, just need to get the code coverage addressed.

abaran30:
@sandikbarr that's interesting... Even locally, that line is marked with "if path not taken".
Screenshot 2024-07-31 at 9 58 14 AM
Other accessibility rules that also use an `allowList` do not have this. Could it be some minor detail with how the rule and options are set up?

sandikbarr:
> @sandikbarr that's interesting... Even locally, that line is marked with "if path not taken". Screenshot 2024-07-31 at 9 58 14 AM Other accessibility rules that also use an `allowList` do not have this. Could it be some minor detail with how the rule and options are set up?

@abaran30 Thanks so much for the input. I will look further at other rules with an allowList option and see if I can make any changes that will resolve the code coverage issue.

sandikbarr:
@abaran30 I have set the coverageProvider to 'v8' for the eslint-plugin-template package, and reports are now showing full coverage for this rule and patch.

abaran30:
@sandikbarr I'm still wondering why we need `v8` for this package and not the others, and why the default (I believe `babel`) still does work for other accessibility rules with an `allowList`.

sandikbarr:
> @sandikbarr I'm still wondering why we need `v8` for this package and not the others, and why the default (I believe `babel`) still does work for other accessibility rules with an `allowList`.

I made several attempts at getting code coverage to match the other rules with an allowList using the babel provider. I tried removing the default options matching other rule implementations. I defined the schema and messages separately. Yet the same line in the configuration still showed `if path not taken`. Results were also unchanged with istanbul ignore. The babel provider is not giving accurate results.

abaran30:
I played around with this a little and discovered that the following condition seems to be causing that `if path not taken`:
```ts
        if (!getDomElements().has(elementType)) {
          return;
        }
```
By adding the following valid test case ...
```ts
  {
    code: `
      <test-component (keydown)="onKeyDown()"></test-component>
    `,
    options: [
      {
        allowList: ['form', 'test-component'],
      },
    ],
  },
```
... I see 100% coverage. Admittedly, I don't know how valuable this kind of test case is, but it does remove the need to override the `coverageProvider`.

What do you think of this approach?

sandikbarr:
Thanks for the assist on this @abaran30. That's a good catch. I looked at the other rules that only apply to HTML DOM elements, and they have test cases for custom elements. I added a positive test case without allowList configuration, and it's now fully covered.

JamesHenry:
Thank you both!

JamesHenry (approved):

Issue #1901, opened 2024-06-30
[@angular-eslint/use-lifecycle-interface] --fix adds interface at wrong position 


phihochzwei:
**Description and reproduction of the issue**

When running eslint with the --fix param, he finds some files in our project which extend from a parent class. When trying to add the lifecycle interface, he adds it BEFORE the extension creating something like this

```TS
export class SpacerComponent implements OnInit extends AbstractCMSComponent {
}
```

which then breaks the project.

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@angular-eslint/eslint-plugin`    | `18.0.1` |
| `@typescript-eslint/parser`        | `18.0.1` |
| `ESLint`                           | `8.57.0` |
| `node`                             | `22.2.0` |

```sh
# Please run `npx ng version` in your project and paste the full output here:
Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1800.5
@angular-devkit/build-angular   18.0.5
@angular-devkit/core            18.0.5
@angular-devkit/schematics      18.0.5
@angular/cli                    18.0.3
@angular/google-maps            18.0.3
@angular/ssr                    18.0.5
@schematics/angular             18.0.5
ng-packagr                      18.0.0
rxjs                            7.5.7
typescript                      5.4.3
webpack                         5.91.0
zone.js                         0.14.3


```

- [ X] I have tried restarting my IDE and the issue persists.
- [ X] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

maxhaice2:
@mattlewis92 Hi, thank you for your work, any thoughts?

PR #1965, opened 2024-08-06
fix(eslint-plugin): fix placement of lifecycle interface for subclasses

al-mart:


al-mart:
https://github.com/angular-eslint/angular-eslint/issues/1901

JamesHenry (changes requested):
@al-mart Yes please take a look at existing rules, there are many cases where auto fixing is tested.

`convertAnnotatedSourceToFailureCase()` is out utility to allow for a nice visual way to write assertions about errors.

`annotatedSource` is with the errors, `annotatedOutput` is after the fixes are applied

al-mart:
Thank you. Will add the missing test cases accordingly.

JamesHenry:
@al-mart Following up on this one please

al-mart:
Yeah, sorry was on vacation without the pc. Will add the PR with tests tomorrow.

al-mart:
@JamesHenry added the missing test case, the problem is only occurring when there is a class extension but no interfaces at all.

JamesHenry:
@al-mart I have updated the branch, will you be addressing the CI failures please?

JamesHenry:
@al-mart I fixed things up for you, thanks again for the PR

PR #1967, opened 2024-08-07
fix(prefer-self-closing-tags): handle both forward and backward slash

5im0n:
Updated the condition to check for both `src/index.html` and `src\index.html` paths in order to support Windows file paths.


JamesHenry:
Thank you @5im0n!

PR #1968, opened 2024-08-07
feat: support eslint.config.mjs and eslint.config.cjs

JasonWeinzierl:
ESLint supports `mjs` and `cjs` file extensions for their flat configs: https://eslint.org/docs/latest/use/configure/configuration-files.  It would be nice to support particularly `mjs` configs so users can use `import`/`export` ESM syntax.

This does not change the schematic to generate an ESM config file; the user would need to manually write/convert their own config.

### DIFF
--- a/packages/builder/src/lint.impl.ts
+++ b/packages/builder/src/lint.impl.ts
@@ -41,10 +41,14 @@ export default convertNxExecutor(
     /**
      * Until ESLint v9 is released and the new so called flat config is the default
      * we only want to support it if the user has explicitly opted into it by converting
-     * their root ESLint config to use eslint.config.js
+     * their root ESLint config to use eslint.config.(js|mjs|cjs).
      */
     const useFlatConfig = existsSync(
       joinPathFragments(workspaceRoot, 'eslint.config.js'),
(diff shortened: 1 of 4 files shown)


JamesHenry (changes requested):
@JasonWeinzierl Thank you but this is incomplete, because the logic for adding extends to newly created projects within a workspace would need to handle this. I would also want to see e2e test coverage for something like to cover a realistic user experience. You can take a look at the existing fixtures for inspiration if you are happy to take that on, otherwise this will need to wait until I have some time.

Please also be aware that recently .ts extension support was added behind a flag... So it might be better to tackle all non .js containing commonjs cases at once instead of this workaround.

JamesHenry:
@JasonWeinzierl Following up on this one please

hakimio:
@JasonWeinzierl are you planning to finish this PR?

JasonWeinzierl:
> @JasonWeinzierl are you planning to finish this PR?

My apologies.  This PR came from a patch to get it working for myself.  I have not had time to look into the requested changes, nor do I feel I have a deep enough understanding of this project to do it properly.

PR #1995, opened 2024-08-15
fix(eslint-plugin-template): [label-has-associated-control] allow custom inputs for default labelComponents

forivall:

With [ng-zorro's checkbox](https://ng.ant.design/components/checkbox/en), the label is used with the `nz-checkbox` directive to create a checkbox element. However, `"@angular-eslint/template/label-has-associated-control"` does not recognize the directive even with the following config:

```JSON
{
  "rules": {
    "@angular-eslint/template/label-has-associated-control": [
      "error",
      {
        "labelComponents": [
          {
            "selector": "label",
            "inputs": [
              "nz-checkbox"
            ]
          }
        ]
      }
    ],
  }
}
```

```HTML
<label nz-checkbox [(ngModel)]="checked">Checkbox</label>
```

**Proposed fix**

By spreading the configured `labelComponents` before the default config, we can override the `inputs` value for the given selector.

### DIFF
--- a/packages/eslint-plugin-template/src/rules/label-has-associated-control.ts
+++ b/packages/eslint-plugin-template/src/rules/label-has-associated-control.ts
@@ -88,8 +88,8 @@ export default createESLintRule<Options, MessageIds>({
       ...(controlComponents ?? []),
     ]);
     const allLabelComponents = [
-      ...DEFAULT_LABEL_COMPONENTS,
       ...(labelComponents ?? []),
+      ...DEFAULT_LABEL_COMPONENTS,
     ] as const;
     let inputItems: TmplAstElement[] = [];
     let labelItems: TmplAstElement[] = [];


JamesHenry:
@forivall I'm afraid this contribution cannot be accepted.

- It would never be possible to accept something like this without appropriate test coverage
- Your proposed change does the opposite of what you are saying it does here:
> By spreading the configured labelComponents before the default config, we can override the inputs value for the given selector.
If you spread before default config, the default config will override. That's why it is the way it is, because the opposite would be undesired behavior.
- Frankly the way this `nz-checkbox` directive works is conceptually not aligned with this lint rule. "label has associated control" is the name, but in the way this has been designed, there is never going to be an associated control for the label.

If there API had involved being more explicit with the control (input), something like this:

```html
<input nz-checkbox [(ngModel)]="checked" />
<label>Checkbox</label>
```

Then it would make sense with this rule. As it is in reality this rule is never going to make sense with it by definition, so you will either need to use `eslint-disable` comments, or disabled the rule.

Hope that helps!

PR #2015, opened 2024-08-26
fix(template-parser): visit receiver of Call expression

P4:
Re-created due to https://github.com/angular-eslint/angular-eslint/pull/1887#issuecomment-2308874016

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/no-call-expression.md
+++ b/packages/eslint-plugin-template/docs/rules/no-call-expression.md
@@ -393,6 +393,62 @@ interface Options {
 
 <br>
 
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/no-call-expression": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```html
+{{ foo()() }}
+   ~~~~~~~
+   ~~~~~
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/no-call-expression": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```html
+{{ foo().bar() }}
+   ~~~~~~~~~~~
+   ~~~~~
+```
+
+<br>
+
+---
+
+<br>
+
 #### Custom Config
 
 ```json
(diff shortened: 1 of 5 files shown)


P4:
@JamesHenry I've re created the PR, sorry for the mess.

The only tests failing are the ones I've added, so it looks like my fix for `Call` doesn't work anymore. I'm changing this into a draft until I manage to fix it.

P4:
Looks like the issue is not with the code but with package resolution during testing via `pnpm`?

I've made changes to both `template-parser` and `eslint-plugin-template`, the latter depends on fixes to the former.

But when running tests for `eslint-plugin-template` it's not actually using the template-parser version that's in the workspace, but pulls an already published version from pnpm instead.

At least that's what it looks like when debugging - `require('@angular-eslint/template-parser')` resolves to `node_modules/.pnpm/@angular-eslint+template-parser@18.3.0_eslint@9.9.1_jiti@1.21.6__typescript@5.5.4/node_modules/@angular-eslint/template-parser/dist/index.js` which doesn't include any of my fixes.

JamesHenry:
Ah ok thanks for digging @P4, that would explain why it seemed to work at first, and then not (cause I changed to pnpm recently and applied that missing dependency relationship).

JamesHenry:
@P4 Please can you add a `.npmrc` file at the root of the workspace with the following contents:

```
link-workspace-packages=true
```

And run `pnpm install`. You should see that the packages are now linked in the lock file. Then try your tests again and hopefully they should work

P4:
Looks like that solved the problem! Should I make a commit with this change or wait for it to be fixed on `main` and do a rebase?

JamesHenry:
Yeah I've opened https://github.com/angular-eslint/angular-eslint/pull/2016

JamesHenry:
Merged, thanks! Sorry for the extra work this caused you

JamesHenry:
@P4 Just wondering, with the utility changes you made, do the existing test cases actually need to change?

e.g.

```
<div>{{ getInfo()() }}</div>
```

I thought the original reason for this change was the utility not supporting it?

P4:
I've rebased the changes on top of the fixed `main` - everything seems to work again, thanks for solving the issue so quickly!

The existing cases would still have to change I believe, since they were expecting the wrong number of errors.
```
<div>{{ getInfo()() }}</div>
```
should report two errors, but was annotated for only one.

I could restore all my test cases and update the annotations on existing ones, but I've decided it's probably better to keep the simple and nested/chained cases separate, at least for readability.

JamesHenry (approved):
Fair enough, thanks again!

P4:
By the way, I've been wondering where some of the names in the test cases come from, namely things like `FunctionCall`, `MethodCall` and `SafeMethodCall`, etc. from the descriptions. AFAIK none of them correspond to the current Angular AST names - is this some carryover from TSLint?

JamesHenry:
Yes it will be

PR #2031, opened 2024-09-03
fix(template-parser): traverse ng-content fallback content

mattlewis92:
In angular 18 support was added for a fallback in ng-content: https://medium.com/@davidepassafaro/angular-fallback-content-in-ng-content-5aceb31acc0a

Which results in angular parsing the contents of `ng-content` differently and there is a new `Content` AST node. In order to lint the contents of `ng-content` fallbacks we need to add the visitor keys for that node type.


packages/template-parser/tests/index.test.ts
```diff
@@ -15397,4 +15397,20 @@ describe('parseForESLint()', () => {
...
+        parseForESLint(`
+          <ng-content>
+             @if (foo) {
+               <div>Hello!</div>
+             }
+          </ng-content>
+        `, {
+          filePath: './foo.html',
+        }).ast,
+      ).toMatchSnapshot();
```
mattlewis92, line 15413: When I tried to use an inline snapshot jest was throwing:
>  Jest: Couldn't locate all inline snapshots.

So I went with a file based snapshot instead. I think given how big these ASTs are, it makes sense to not have them inline anyway

JamesHenry (approved):
Thank you @mattlewis92!

packages/template-parser/tests/index.test.ts
JamesHenry: Yeah it's weird how that happens sometimes, it seems to be specific to the machine in question

JamesHenry: I've never been able to resolve it once it starts either...

PR #2088, opened 2024-10-23
fix(eslint-plugin-template): prevent the slot tag from being self-closing

### LINKED ISSUE #2087: [prefer-self-closing-tags] The <slot> tag should not be considered a self-closing tag

```JSON
{
  "rules": {
    "@angular-eslint/template/prefer-self-closing-tags": ["error"]
  }
}
```

```HTML
<div #content class="content">
  <slot name="body-slot"></slot>
  <slot></slot>
</div>
```

After eslint --fix the file looks like this, which raises the following template error:
`Only void, custom and foreign elements can be self closed "slot" ngtsc(-995002)`

```HTML
<div>
  <slot name="body-slot" />
  <slot />
</div>
```

The issue seems to be that [get-dom-elements.ts](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/utils/get-dom-elements.ts#L5) uses the `aria-query` library to obtain a list of all native DOM elements, and `slot` is not on that list because it doesn't support ARIA roles.

**Versions**

| `@angular-eslint/eslint-plugin-template`  | `18.4.0` |
| `@angular-eslint/template-parser`         | `18.4.0` |
| `@typescript-eslint/parser`               | `8.11.0` |
| `ESLint`                                  | `9.13.0` |
| `node`                                    | `20.17.0` |

```sh
Angular CLI: 18.2.8
Node: 20.17.0
Package Manager: npm 10.8.2
OS: win32 x64

Angular: 18.2.8
... animations, cdk, cli, common, compiler, compiler-cli, core
... elements, forms, platform-browser, platform-browser-dynamic
... router

@angular-devkit/architect       0.1802.8
@angular-devkit/build-angular   18.2.8
@angular-devkit/core            18.2.8
@angular-devkit/schematics      18.2.8
@schematics/angular             18.2.8
ng-packagr                      16.2.3
typescript                      5.4.2
zone.js                         0.14.0
```

- [ ] I have tried restarting my IDE and the issue persists.

joanllenas:
The slot tag is not included in the tags exposed by the aria-query library, which is used to extract native DOM elements and determine which tags can be self-closing. This fix adds the slot tag and a few other missing tags

Fixes #2087


packages/eslint-plugin-template/tests/rules/prefer-self-closing-tags/cases.ts
```diff
@@ -20,6 +20,7 @@ export const valid: readonly (string | ValidTestCase<Options>)[] = [
       [items]="items" />
   `,
   '<img />',
+  '<slot></slot>',
```
JamesHenry, line 23:
```suggestion
  // These elements cannot be self-closing
  '<slot></slot><math></math><rb></rb><svg></svg><template></template>',
```

JamesHenry (changes requested):
Thank you! May as well include all the examples to protect against regressions

packages/eslint-plugin-template/tests/rules/prefer-self-closing-tags/cases.ts
```diff
@@ -20,6 +20,8 @@ export const valid: readonly (string | ValidTestCase<Options>)[] = [
       [items]="items" />
   `,
   '<img />',
+  // These elements cannot be self-closing
+  '<slot></slot><math></math><rb></rb><svg></svg><template></template>',
```
JamesHenry, line 24: The test case I updated didn't pass, please can you take a look?

joanllenas: ### Problem
While debugging `prefer-self-closing-tags.ts::createESLintRule()`, I discovered that `svg` and `math` elements are parsed differently when resolving the `name` property on the `node` object provided in the function returned by `create()`. Instead of `svg` or `math`, the `node.name` property contains `:svg:svg` and `:math:math`. This discrepancy causes the test to fail.

### Potential Solution
Include `:svg:svg` and `:math:math` in the `nonAriaElements` array:

```ts
const nonAriaElements = [
  'slot',
  'math',
  'rb',
  'svg',
  'template',
  ':svg:svg',
  ':math:math',
];
```

To be honest, I don't like it, but unless this is caused by a bug in the template parser and we can treat it as a temporary workaround to remove once it’s fixed in the parser repository, I don’t know what else we can do.
Suggestions @JamesHenry ?

JamesHenry: It looks like the Angular compiler has special handling for math and svg as namespaces. Their children have the namespace in the node name too:

image

Defined in here: https://github.com/angular/angular/blob/357166e9a4125069cdf27c8d4d4018ceb73f4f14/packages/compiler/src/schema/dom_element_schema_registry.ts

So let's go ahead with it and add a note about this Angular compiler behaviour inline as a comment

joanllenas: Thanks, @JamesHenry,  
I've made the changes. Let me know if you need anything else. This should be ready to go.

JamesHenry (approved):

JamesHenry (changes requested):
You need to please update the rule docs:

```sh
pnpm update-rule-docs
```

We auto generate examples from the unit tests

JamesHenry:
Thank you!

PR #2096, opened 2024-11-14
feat(eslint-plugin)!: prefer-standalone recognizes that standalone is the default

json-derulo:
In Angular 19, [standalone will be the new default](https://blog.angular.dev/the-future-is-standalone-475d7edbc706). There will also be an automigration which removes the `standalone: true` property from the metadata. This plugin needs to be updated in order to adapt to those changes.

Note: It would be better if the fixer would remove the whole attribute, but with the given tooling this is not so easy to achieve. For now the fixer will transform the `standalone: false` attribute to `standalone: true`.

### DIFF
--- a/packages/eslint-plugin/docs/rules/prefer-standalone.md
+++ b/packages/eslint-plugin/docs/rules/prefer-standalone.md
@@ -56,34 +56,6 @@ The rule does not have any configuration options.
 #### ❌ Invalid Code
 
 ```ts
-@Component({})
-~~~~~~~~~~~~~~
-class Test {}
-```
-
-<br>
-
----
-
-<br>
-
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks for this @json-derulo, I will set up the v19 branch and we can target that with this PR

JamesHenry:
FYI https://github.com/angular-eslint/angular-eslint/pull/2112 there is no need for us to update the prefer-standalone-components rule here

JamesHenry:
Merged. Sorry you'll need to rebase now

json-derulo:
@JamesHenry rebased.

JamesHenry:
Thank you @json-derulo!

PR #2113, opened 2024-11-19
feat(eslint-plugin)!: remove deprecated no-host-metadata-property rule

json-derulo:
Removes the deprecated `no-host-metadata-property` rule, as Angular now recommends using it.

PR #2121, opened 2024-11-20
feat(eslint-plugin)!: remove deprecated no-host-metadata-property rule

leosvelperez:
It was deprecated in v18 and the recommendation from the Angular Team has now inverted.


json-derulo:
duplicate of #2113

PR #2122, opened 2024-11-20
feat(eslint-plugin)!: remove deprecated sort-ngmodule-metadata-arrays rule

leosvelperez:
It has been deprecated for some time and was never reliable because of the way ESLint works. Additionally, particularly in v19 of Angular, NgModules are far less emphasized/encouraged in general.


json-derulo:
duplicate of #2114

PR #2156, opened 2024-12-04
feat(eslint-plugin): introduce sort-signals rule

krasi:
The purpose of this PR is to introduce a new `sort-signals` rule for sorting signals. This one is fairly similar to the lifecycle sorting rule.

```json
{
  "rules": {
    "@angular-eslint/sort-signals": [
      "error"
    ]
  }
}
```

Ensures a default order `"input","model","output","viewChild","viewChildren","contentChild","contentChildren","inject","signal","computed"`

It allows defining custom order, such as:
```json
{
  "rules": {
    "@angular-eslint/sort-signals": [
      "error",
      {
        "order": [
          "output",
          "input"
        ]
      }
    ]
  }
}
```

which will check the order only for the specified signal functions and will require them to be put on top of the component.


krasi:
@JamesHenry I'm sorry for tagging you but I was hoping to get some feedback on this. Thank you! 😊

lucasnbsb:
up

JamesHenry (changes requested):
I'm sorry for the delay, I took time off over the holidays.

I would like to have some more discussion about the default ordering. How did you arrive upon this specific choice?

When I look at components written by the Angular and Angular Material folks, the don't necessarily align with what as been set here. That's not the be all and end all, but it's a good data point.

What about other popular, well-established component libraries in the ecosystem, what ordering do they tend to use (even though likely not enforced)? It would be good to match precedent with a rule such as this, or at least be able to clearly justify why the order is the way it is.

Some concrete examples that stood out:

- I would personally expect inputs and outputs to live right next to each and not be split up like you have here.
- inject seems to be commonly used by Angular folks at the very top, even before inputs and outputs, which kind of makes sense when you think about them as the dependencies for the component to work fundamentally...

krasi:
> I'm sorry for the delay, I took time off over the holidays.
> 
> I would like to have some more discussion about the default ordering. How did you arrive upon this specific choice?
> 
> When I look at components written by the Angular and Angular Material folks, the don't necessarily align with what as been set here. That's not the be all and end all, but it's a good data point.
> 
> What about other popular, well-established component libraries in the ecosystem, what ordering do they tend to use (even though likely not enforced)? It would be good to match precedent with a rule such as this, or at least be able to clearly justify why the order is the way it is.
> 
> Some concrete examples that stood out:
> 
> * I would personally expect inputs and outputs to live right next to each and not be split up like you have here.
> * inject seems to be commonly used by Angular folks at the very top, even before inputs and outputs, which kind of makes sense when you think about them as the dependencies for the component to work fundamentally...

Hi @JamesHenry !

Thank you for your feedback!

About the default order:
- My logic for `input, model, output` was that model is like an extended input and will fit next to it but I get the point that it should be `input, output, model`
- Regarding `inject`, I was wondering what kind of information regarding a component will be most valuable for the developer and at the same time, injects are usually private properties and they would usually go after the public ones. At the same time, I get the argument that inject should be on top as in the previous syntax component flow it goes in the constructor. And especially as the major Angular libraries such as Angular Material are following this pattern, it should be the default one.



I just pushed a commit to update the default order as suggested! 😃

JamesHenry:
Thanks @krasi, I've requested some feedback from the Angular Team in case there is a pattern being formally or informally followed within Google for this

JamesHenry:
They've rightly pointed out that this rule name doesn't really make sense, `output` and `inject` are not based on signals...

Another comment was "Also I wouldn’t recommend enforcing computed/signal/linkedSignal sorting as grouping might be more relevant."

In general I kind of hate sorting rules tbh... They are a nightmare to auto-fix because of code comments, and without auto-fixing they are a nuisance...

I'll see if there is any more feedback on that thread before making a final decision, let me know if you have other suggestions around the name

JamesHenry:
Hi Folks, I've discussed this further with the Angular Team and the view is that it is simply not appropriate to develop a one-size fits all recommendation for sorting these properties.

This is in addition to the rule name not being applicable, given not all of these are signal APIs, and my point around reliable auto-fixing not being possible because of code comments etc.

For these reasons, we won't move forward with adding this rule to `angular-eslint` directly. Naturally, you are more than welcome to use the utilities we provide to still leverage this rule implementation in your own plugin, and whether or not you choose to publish that plugin with the community is totally up to you.

Hope that rationale makes sense, and if you do publish it, feel free to comment the link here for visibility.

robertIsaac:
hi @JamesHenry 
while I agree about the name could be better
I strongly disagree about rejecting to add the rule altogether
and there is no need to add autofix for a rule to be useful
https://typescript-eslint.io/rules/member-ordering/ has no fixes and it's still very useful and many people use it

JamesHenry:
I understand (I am the creator of typescript-eslint), the big difference here is that the Angular Team has explicitly told me that they would not recommend having a fixed order, it is situation and codebase dependent.

Overriding that direct message from the framework authors to invent my own order. that would then definitely be treated as "the order" by the majority of the community who would not do their own research, is not something I am comfortable with.

The only way I can imagine such a rule shipping from angular-eslint therefore is if it shipped with no default order, i.e. the rule would blow up at runtime if no options were set to provide an order for it to enforce. This is a bit of a weird situation in ESLint, and off the top of my head I cannot think of any other rule that behaves like this.

Therefore the three conditions to rekindle this would be:
- different name (easy, just requires some bike-shedding)
- no auto-fix (easy, and precedent from member-ordering)
- no default/recommended order (maybe unprecedented and definitely _weird_)

Does that make sense @robertIsaac?

@reduckted @krasi do you have any additional thoughts on the above?

reduckted:
Some thoughts (may or may not be related to what was mentioned above 😁):

I'll start by saying that I'm not really the target for a rule like this as it stands. Strict member ordering rules have always been something that I've shied away from because I find that they can make the code harder to follow. For example, a rule like "all private methods must be last" can mean you are jumping up and down through the file if you're reading a public method and have to keep scrolling down to the private methods that it calls. A better layout would be to have the private methods close to the public method(s) that use them. That is something that a rule could never accurately enforce as it requires a lot more knowledge than just some attributes of the member.

To this rule, I can see some holes with the logic.

Using these options:
```js
`"order": [ "output", "input" ]
```
This would be considered a valid component layout:

```ts
@Component()
class Test {
    readonly alpha = input();

    readonly beta = input();

    public foo(){}

    readonly gamma = input();

    public bar(){}

    readonly delta = output();

    constructor() {}

    readonly epsilon = output();
}
```

All inputs come before all outputs, therefore there is no problem. I don't think that layout is really what this rule intended though.

There _are_ other rules in the ESLint ecosystem that could be used together with this rule to enforce a better ordering, and perhaps that was the intention. I'm thinking specifically about [`perfectionist/sort-classes`](https://perfectionist.dev/rules/sort-classes#groups) and its `groups` option.

Used together, the two rules would be able to define a better layout. For example, you could use `perfectionist/sort-classes` to say that all properties should appear before methods and constructors. That would require the inputs and outputs to be declared first, and then this rule would be able to enforce that the inputs come before the outputs.

----

Let me throw a different idea out there that might (or might not) still fulfill the intention of this rule:

Instead of requiring a specific order based on assignment, what if the rule enforced _grouping_ of these properties. For example, all `viewChild` and `viewChildren` properties should appear together. It doesn't matter _where_ they appear in the component, nor whether they are above or below other types of properties, but they must appear together. This could even be targeted at specific kinds of properties. For example, _private_ `inject` properties must appear together, but public or protected `inject` properties don't have to be together with those private properties.

Here's a hypothetical example of what the options might look like:
```json
{
  "groups": [
    { "functions": ["viewChild", "viewChildren"] },
    { "functions": ["input", "model"] },
    { "functions": ["output"] },
    { "functions": ["inject"], "kinds": ["private"] }
    { "functions": ["signalLike"], "kinds": ["private"] }
  ]
}
```

And here's what a component might look like that follows this rule:
```ts
@Component()
class Test {
  #alpha = inject(AlphaService);             // |___ These must appear together.
  #beta = inject(AlphaService);              // |

  private readonly counter = signal();       // | 
  private readonly open = computed(          // |__ These signal-like properties
    () => counter() === 0                    // |   Must appear together.
  );                                         // |

  protected readonly items = viewChildren(); // |
                                             // |--- These must appear together.
  protected readonly header = viewChild();   // |

  public readonly isImportant() = signal();  // |___ This doesn't have to be with the other
                                             // |    signal properties because it is not private.

  public readonly title = input();           //|
  public readonly subtitle = input();        //|___ These must appear together. The `input` 
  public readonly value = model();           //|    and `model can appear in any order.
  public readonly footer = input();          //|

  public readonly close = output();          // |-- This cannot be mixed in with the 
                                             //    `input` and `model` properties above.

  public readonly helper = inject(Helper);   // |--- This can appear separate to 
                                             //      the other `inject` properties.
}
```

robertIsaac:
@JamesHenry 
I think what you said make sense and agree to the 3 conditions mentioned

robertIsaac:
@JamesHenry what is the plan now regarding this?
it has been months

maartentibau:
> @JamesHenry what is the plan now regarding this? it has been months

I really, really would love this feature as well.
I have been using https://perfectionist.dev/rules/sort-classes

It works `"OK"`, but I had to define all the `regex` patterns myself. It does not work perfectly... but it does the job more or less.

tinpa-de:
Hey @maartentibau, would you kindly provide your regex expressions here? Would be very helpful for me and maybe also others 😊

maartentibau:
@tinpa-de 

This is the config that I'm using:
The `regex` is in general everywhere the same, besides the lookup it needs to do for the specific keyword.

To get to this `regex` I had to do A LOT trail and error, I used a ton of code samples in regex101 to verify if the regex does the job properly and so far I'm cannot complain.

```
{
  'perfectionist/sort-classes': [
    'error',
    {
      type: 'unsorted',
      groups: [
        'angular-inject',
        'angular-input-required',
        'angular-input',
        'angular-model-required',
        'angular-model',
        'angular-output',
        'private-property',
        'protected-property',
        'static-property',
        'public-property',
        'angular-content-child-required',
        'angular-content-child',
        'angular-content-children-required',
        'angular-content-children',
        'angular-view-child-required',
        'angular-view-child',
        'angular-view-children-required',
        'angular-view-children',
        'angular-signal',
        'angular-computed',
        'angular-linked-signal',
        ['get-method', 'set-method'],
        'constructor',
        'static-method',
        'public-static-method',
        'method',
        'public-method',
        'protected-static-method',
        'protected-method',
        'private-static-method',
        'private-method'
      ],
      customGroups: [
        {
          groupName: 'angular-input',
          selector: 'property',
          elementValuePattern: 's?input(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-input-required',
          selector: 'property',
          elementValuePattern: 's?input\\.required(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-model',
          selector: 'property',
          elementValuePattern: 's?model(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-model-required',
          selector: 'property',
          elementValuePattern: 's?model\\.required(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-output',
          selector: 'property',
          elementValuePattern: 's?output(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-inject',
          selector: 'property',
          elementValuePattern: 's?inject(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-signal',
          selector: 'property',
          elementValuePattern: 's?signal(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-content-child',
          selector: 'property',
          elementValuePattern: 's?viewChild(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-content-child-required',
          selector: 'property',
          elementValuePattern: 's?viewChild\\.required(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-content-children',
          selector: 'property',
          elementValuePattern: 's?viewChildren(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-content-children-required',
          selector: 'property',
          elementValuePattern: 's?viewChildren\\.required(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-view-child',
          selector: 'property',
          elementValuePattern: 's?viewChild(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-view-child-required',
          selector: 'property',
          elementValuePattern: 's?viewChild\\.required(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-view-children',
          selector: 'property',
          elementValuePattern: 's?viewChildren(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-view-children-required',
          selector: 'property',
          elementValuePattern: 's?viewChildren\\.required(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-computed',
          selector: 'property',
          elementValuePattern: 's?computed(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        },
        {
          groupName: 'angular-linked-signal',
          selector: 'property',
          elementValuePattern: 's?linkedSignal(?:<[\\s\\S]*?>)?\\([\\s\\S]*?\\)'
        }
      ]
    }
  ],
  'perfectionist/sort-variable-declarations': 'off',
  'perfectionist/sort-intersection-types': 'off',
  'perfectionist/sort-heritage-clauses': 'off',
  'perfectionist/sort-array-includes': 'off',
  'perfectionist/sort-named-imports': 'off',
  'perfectionist/sort-named-exports': 'off',
  'perfectionist/sort-object-types': 'off',
  'perfectionist/sort-union-types': 'off',
  'perfectionist/sort-switch-case': 'off',
  'perfectionist/sort-interfaces': 'off',
  'perfectionist/sort-decorators': 'off',
  'perfectionist/sort-jsx-props': 'off',
  'perfectionist/sort-modules': 'off',
  'perfectionist/sort-imports': 'off',
  'perfectionist/sort-exports': 'off',
  'perfectionist/sort-objects': 'off',
  'perfectionist/sort-enums': 'off',
  'perfectionist/sort-sets': 'off',
  'perfectionist/sort-maps': 'off'
};
```

PR #2184, opened 2024-12-15
fix(eslint-plugin): [no-input-prefix] false positive on input initializer value

lucasnbsb:
Input properties initialized with literal values containing the prohibited prefix were being incorrectly flagged, causing code like:

```typescript
// the prefix is "on"
@Component()
class Test {
  @Input() notOn: string = 'on';
}
```

To trigger on the initializer 

That happened because the selector matched any `literal` or `identifier` that were children of property definitions.

So the fix could either be in the selector or once the match already happened. I went with the latter.

I added 3 test cases, one valid with the disalowed prefix in the initializer instead of the property name and two invalids with the initializer not being matched,


JamesHenry (changes requested):
Thanks a lot @lucasnbsb, you please need to update the rule docs and commit the results. Our rule docs are generated from our unit tests, so whenever they change the rule docs need to as well.

```sh
pnpm update-rule-docs
```

lucasnbsb:
Hi @JamesHenry. Docs updated 👍

PR #2186, opened 2024-12-18
feat(eslint-plugin): add options for custom ID and default value validation

tcorral:
Added two more options to report errors if:
- Custom ID is not present
- Custom ID is present but it doesn't follow the allowed pattern
- Default value is not provided

Change in details:
- Added new configuration options and error messages for reporting.
- Create and use a `validate` function that receives the input string and a type to validate optionally it can receive a pattern to validate custom id.
- The core of the `validate` function is a regular expression with named group, I have followed how Angular manages metadata when using `$localize`
- Added new valid and invalid tests cases to cover the changes.

New config after changes:
- Added `requireCustomId` - `boolean` - default is `false`
- Added `requireDefaultValue` - `boolean` - default is `false`
- Added `boundTextAllowedPattern` - `string` - No default value. It expects to get a string that can be converted to regular expression to test custom id. I have used the same name that already exists in `@angular-eslint/template-parser`.


tcorral:
Any update on this PR?

packages/eslint-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@angular-eslint/eslint-plugin",
-  "version": "19.0.2",
+  "version": "19.1.0",
```
JamesHenry, line 3: Revert this, versioning is automated and does not happen through imperative updates to version fields, and all packages are released together so changing only one package would never be the right thing to do.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -6,29 +6,68 @@ export type Options = [
...
+      } else {
+        result = !!match.groups[typeToValidate];
+      }
+    }
+  }
+  return result;
+};
 
 const STYLE_GUIDE_LINK = 'https://angular.dev/guide/i18n';
 const STYLE_GUIDE_LINK_COMMON_PREPARE = `${STYLE_GUIDE_LINK}-common-prepare`;
```
JamesHenry, line 62: This link no longer seems to be correct at all (I think the new one is https://angular.dev/guide/i18n/prepare?) so let's please fix it as part of this PR so that all the various errors have reasonable links.

I discovered this when trying to verify the ones you added (which should be here now I believe https://angular.dev/guide/i18n/manage-marked-text)

packages/eslint-plugin/src/rules/require-localize-metadata.ts
```diff
@@ -51,23 +90,51 @@ export default createESLintRule<Options, MessageIds>({
...
           },
+          requireCustomId: {
+            type: 'boolean',
+            default: DEFAULT_OPTIONS.requireCustomId,
+          },
+          requireDefaultValue: {
+            type: 'boolean',
+            default: DEFAULT_OPTIONS.requireDefaultValue,
+          },
+          boundTextAllowedPattern: {
```
JamesHenry, line 101: I don't think this is very clear that it relates to custom ID.

I think something like this would be much clearer:

```json
{
  "@angular-eslint/require-localizae-metadata": [
     "error",
     {
       "requireCustomId": true,
       "requireCustomIdMatchingPattern": "custom_id"
     }
  ]
}
```

We potentially could then take it a step further and dynamically default `requireCustomId` to true when `requireCustomIdMatchingPattern`?

JamesHenry: With that said, I am not super clear on the examples you have of:

```
boundTextAllowedPattern:
          '^(?<lib>.[^\\.]*)\\.(?<component>.[^\\.]*)\\.(?<tag>.[^\\.]*)\\.(?<attribute>.[^\\.]*)?\\.(?<name>.[^\\.]*).$',
```

Are all of the sections potentially applicable to just the custom ID in practice? (I am not somebody who has ever leveraged the i18n feature personally). Maybe you could please give some real world examples?

JamesHenry (changes requested):
Thanks a lot @tcorral, some feedback and open questions

packages/eslint-plugin/src/rules/require-localize-metadata.ts
tcorral: To answer your last question:

This is an example of pattern on which you have a more verbose pattern that can be used for more people to understand where the translation is used and that can help to identify and avoid duplicated ids.

Here you have an example of a custom id that matches this regular expression: 'cards-management.form-details.input.placeholder.title'

This is just an example and can be overriden, but I thought it could be interesting to have an example.

tcorral: > With that said, I am not super clear on the examples you have of:
> 
> ```
> boundTextAllowedPattern:
>           '^(?<lib>.[^\\.]*)\\.(?<component>.[^\\.]*)\\.(?<tag>.[^\\.]*)\\.(?<attribute>.[^\\.]*)?\\.(?<name>.[^\\.]*).$',
> ```
> 
> Are all of the sections potentially applicable to just the custom ID in practice? (I am not somebody who has ever leveraged the i18n feature personally). Maybe you could please give some real world examples?

I thought that it could be interesting to give some example to the developers to avoid duplicated custom ids and also to provide some more information about where the custom id is used.

Here you have an example of a custom id matching the regular expression: "cards-management.card-details-form.input.placeholder.remaining-credit"

tcorral: > I don't think this is very clear that it relates to custom ID.
> 
> I think something like this would be much clearer:
> 
> ```json
> {
>   "@angular-eslint/require-localizae-metadata": [
>      "error",
>      {
>        "requireCustomId": true,
>        "requireCustomIdMatchingPattern": "custom_id"
>      }
>   ]
> }
> ```
> 
> We potentially could then take it a step further and dynamically default `requireCustomId` to true when `requireCustomIdMatchingPattern`?

I have changed the name of the property to the one you have suggested.

How would you like to dynamically default `requireCustomId` to true?

packages/eslint-plugin/src/rules/require-localize-metadata.ts
tcorral: Updated.

packages/eslint-plugin/package.json
tcorral: Reverted.

packages/eslint-plugin/tests/rules/require-localize-metadata/cases.ts
```diff
@@ -168,15 +205,260 @@ export const invalid: readonly InvalidTestCase<MessageIds, Options>[] = [
                                      ~~~~~~~~~~~~~~
     `,
     messages: [
+      {
+        char: '~',
```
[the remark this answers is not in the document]
tcorral: I found it was needed for the tests to pass green.

packages/eslint-plugin/src/rules/require-localize-metadata.ts
JamesHenry: You changed the messageId but not the option which is the more prominent user-facing thing that I was concerned about.

For a dynamic default, you could just not specify it on the schema and apply it within the rule implementation at runtime, right?

packages/eslint-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@angular-eslint/eslint-plugin",
-  "version": "19.1.0",
```
JamesHenry, line 3: This shouldn't be here at all, there should not be any changes to package.json files in this PR

packages/eslint-plugin/tests/rules/require-localize-metadata/cases.ts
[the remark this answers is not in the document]
JamesHenry: I'd need to look at the test and your changes in detail, but I probably suggests a breaking change was introduced with this PR which we would want to avoid. Maybe something needs to be off by default? If you could at least please get to the bottom of why this change was needed we could then discuss if the change of behaviour is minor/objectively desired enough for it to go in as is, otherwise it might need to be worked around and changed in a major

JamesHenry:
@tcorral Would you like some help in finalizing this one?

tcorral:
@JamesHenry yes, please.

JamesHenry:
Ok I should be able to find some time in the next couple of weeks

igord:
@JamesHenry please check this PR adding the same functionality but with slightly different approach

JamesHenry:
Closing in favour of #2430

PR #2187, opened 2024-12-20
feat(eslint-plugin): introduce sort keys in type-decorator rule

### LINKED ISSUE #797: [sort-keys-in-type-decorator] Auto sort keys in angular decorators

```ts
@Component({
    styleUrls: ['./money.style.less'],
    selector: 'money',
    changeDetection: ChangeDetectionStrategy.OnPush,
    templateUrl: './money.template.html',
})
```

It would be really cool if we could sort the keys automatically

`eslint . --fix`

```ts
@Component({
    selector: 'money',
    templateUrl: './money.template.html',
    styleUrls: ['./money.style.less'],
    changeDetection: ChangeDetectionStrategy.OnPush,
})
```

```JSON
{
  "rules": {
    "@angular-eslint/sort-keys-in-type-decorator": [
       {
          "Component": ["selector", "templateUrl", "styleUrls", "changeDetection"]
       }
    ]
  }
}
```

bschaeublin:
Fixes #797 

This PR introduces the ability to sort keys within type decorators (e.g., `@Component`, `@NgModule`, etc.) in the `type-decorator-order` rule.

**Motivation:**
This addresses issue #797, which has been open for a while. I was interested in this functionality and, since it hadn't been implemented yet, I decided to take it on as a way to learn more about ESLint rules, AST parsing, and contribute to the angular-eslint project. This is my first PR in this repository and my first time creating a new ESLint rule, so any guidance and feedback are greatly appreciated!

**Changes:**
- Added a new rule `sort-keys-in-type-decorator` to ensure consistent key ordering in type decorators such as `Component`, `Directive`, `NgModule`, and `Pipe`.
- Users can specify a custom sort order, or rely on the default sort order for certain decorators (the default sort is currently taken from the Angular CLI schematics).

**Areas for Discussion/Improvement:**
I'd love to get feedback on the following points to ensure this PR aligns well with the project's goals and coding style:

1.  **Default Sorts from Angular CLI Schematics:**
    - The current implementation uses default sorts for some decorators based on the order generated by Angular CLI schematics.
    - [x] after discussing we can expect few changes in that matter, but added substitutions (template/templateUrl, etc).

2.  **Using Enums vs. Hardcoded Decorator Sets:**
    -   Currently, the rule uses a finite, hardcoded set of decorators. It might be more maintainable to use existing enums like `AngularClassDecorators`.
    -  Is there a way to use enums that can be reflected effectively in the rule's schema? I wasn't sure what is the right way to go.
    - [x] the current proposal is to hardcode the supported decorators in the configuration, as it would be not much of a benefit. Together with another approach, allowing for an unlimited number of properties, rather than a fixed set.  While this avoids the need to regularly update the known properties for each decorator, it does mean we'll lose better schema support during configuration. This could be discussed or changed if needed.

I'm eager to learn and improve this PR based on your feedback. Thanks in advance for your time and help!


packages/eslint-plugin/src/rules/sort-keys-in-type-decorator.ts
```diff
@@ -0,0 +1,226 @@
...
+        'Keys in @{{decorator}} decorator should be ordered: {{expectedOrder}}',
+    },
+  },
+  defaultOptions: [DEFAULT_ORDER],
+  create(
+    context: Readonly<TSESLint.RuleContext<MessageIds, Options>>,
+    [orderConfig]: Readonly<Options>,
+  ) {
+    return {
+      Decorator(node: TSESTree.Decorator) {
```
reduckted, line 82: I think the best way of selecting the decorators is to use the technique from `contextual-lifecycle`:
https://github.com/angular-eslint/angular-eslint/blob/79f2d5d606fae479d95a0a9a69976fbcbca8ab78/packages/eslint-plugin/src/rules/contextual-lifecycle.ts#L53-L69

This ensures that the decorators are selected the same way as they are for other rules that target the decorators.

bschaeublin: @reduckted Thank you for looking at the pr and your suggestion pointing me to the right direction. I updated the PR reusing the pattern from `contextual-lifecycle`, as well reused some `ASTUtils` Helper methods instead of reinventing the wheel :).

If you got any other things I should take care of, let me know!

reduckted:
It looks like the docs haven't been updated. There are three commands that you can run to update things when a new rule is added (I often forget to do this myself 😆 ):
- `pnpm run update-rule-configs`
- `pnpm run update-rule-lists`
- `pnpm run update-rule-docs`

reduckted:
> The current implementation uses default sorts for some decorators based on the order generated by Angular CLI schematics.

I think this is a good idea. If people are using the schematics to create components, then most of the properties will be in the correct order already.

> it would require us to keep these defaults updated whenever the schematics change.

Looking at the `git blame` for the component schematic, other than a recent change to the `standalone` property, the template hasn't changed in years, so I don't think there's any need to be concerned about the schematic reordering the properties. In any case, if the default order is ever changed (apart from adding new properties that come into existence), then that would be a breaking change, so I don't think the default order would be changed.

The only think I would suggest is including _every_ property in the defaults, rather than only those that are used by the schematic. For example, there is `styleUrl`, `styleUrls` and `styles`. If you enable this rule and are using `styles` instead of `styleUrl`, the that property would be moved to the bottom (I think, I'm not quite sure what happens with properties that aren't listed in the options).

> Is there a way to use enums that can be reflected effectively in the rule's schema? I wasn't sure what is the right way to go.

`AngularClassDecorators` is a non-const enum, so in theory you could dynamically build the schema, but it's a bit ugly.
```ts
schema: [
      {
        type: 'object',
        properties: Object.fromEntries(
          Object.keys(AngularClassDecorators).map((key) => [
            key,
            {
              type: 'array',
              items: {
                type: 'string',
              },
            },
          ]),
        ),
        additionalProperties: false,
      },
    ],
```
That produces an options object that looks like this:
```ts
interface Options {
  Component?: string[];
  Directive?: string[];
  Injectable?: string[];
  NgModule?: string[];
  Pipe?: string[];
}
```
So you'd have to exclude `Injectable` manually. Personally, I don't think it's worth it. What I think _might_ be better is to define the options for each decorator type as an "enum" so that you can only specify known properties. For example:
```ts
Component: {
  type: 'array',
  items: {
    type: 'string',
    enum: [
      'changeDetection',
      'viewProviders',
      'moduleId',
      'templateUrl',
      'template',
      'styleUrl',
      'styleUrls',
      'styles',
      'etc...',
    ],
  },
},
```
The only downside to this is there a higher maintenance burden - any time a new property is added to the decorators, the rule needs to be updated. However, the upside is that by defining the schema like this, and also making the `Options` type have the same shape, you'll get strongly-typed configurations, which will make it easy for consumers to know what property names they can specify for each type of decorator. That's just my personal opinion though, and @JamesHenry may have other suggestions. 😃

bschaeublin:
> It looks like the docs haven't been updated. There are three commands that you can run to update things when a new rule is added (I often forget to do this myself 😆 ):
> 
> * `pnpm run update-rule-configs`
> * `pnpm run update-rule-lists`
> * `pnpm run update-rule-docs`

Thanks! i only saw the update-rule-docs so far :).

> The only think I would suggest is including every property in the defaults, rather than only those that are used by the schematic. For example, there is styleUrl, styleUrls and styles. If you enable this rule and are using styles instead of styleUrl, the that property would be moved to the bottom (I think, I'm not quite sure what happens with properties that aren't listed in the options).

Good catch, makes also sense in my opinion. I took a look at the [metadata](https://github.com/angular/angular/blob/main/packages/core/src/metadata/) to see for substitutions - the only ones were it component (style, styleUrls, styleUIrl and templateUrl, template).

Thank you also for the inputs regarding the config-schema!

> ...The only downside to this is there a higher maintenance burden - any time a new property is added to the decorators, the rule needs to be updated.

I totaly see that - so I would wait for @JamesHenry s opinion to that matter  :).

nickbullock:
hey @bschaeublin, thanks for the PR!
very useful rule, will definitely use it in my team once it's merged

bschaeublin:
@JamesHenry I'm sorry to tag you, as I see that you always got a lot to do, but after hitting the three month mark, I thought I try my luck to get your feedback on this PR (regarding configuration, and general) after I've made some improvements with @reduckted, so that we could complete it :). 

Thank you!

bschaeublin:
Thanks @JamesHenry for your response, I saw (to late) your comment in another PR as well, that you didn't like these kind of rules for these (proabably truly) difficult handling of comments. So if you're still open to this like you mentioned, I'd like to search for a stable and nice solution.

I think the repository [eslint-plugin-perfectionist](https://github.com/azat-io/eslint-plugin-perfectionist) (which kind soley sorting stuff), do consider comments when re-sorting properties. I saw for example that they work with ranges which contains relevant comments (top, inlinebefore and after), ex. see [getNodeRange](https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/utils/get-node-range.ts#L21).

I still have to validate that, but if they solved that problem we could do it similar?
Either way I'd be happy to search for a solution.

JamesHenry:
I don't have strong opinions about how you solve it, but the challenge is there on the above snippet and whether it can be proven to be auto-fixable

packages/eslint-plugin/tests/rules/sort-keys-in-type-decorator/cases.ts
```diff
@@ -0,0 +1,606 @@
...
+        templateUrl: './app.component.html',
+        /* Multi-line comment
+           above styleUrl */
+        styleUrl: './app.component.css',
+        // Comment above encapsulation
+        encapsulation: ViewEncapsulation.None, /* Inline comment for encapsulation */
+        // Comment above changeDetection
+        changeDetection: ChangeDetectionStrategy.OnPush // Inline comment for changeDetection
+      })
+      class Test {}
```
bschaeublin, line 603: This is a more complex combined test based on scenarios from your input:
- It ensures that comments above as well as inline comments get sorted correctly with their associated property (see `changeDetection`)
- It also ensures that a comma gets inserted between the property and the comment when it isn't the last property anymore to prevent syntax errors (see `encapsulation`)

All other tests above this ensure dedicated support for multiline comments, inline multi-line comments, and general autofix capabilities.

@JamesHenry Let me know what you think or if you have any other suggestions.

By resorting the properties, newlines between properties are removed by the fixer if there are any between the properties. Let me know if we should support preserving those newlines as well - but that could be seen as a formatting concern rather than a sorting issue.

JamesHenry: Nice work!

JamesHenry:
@bschaeublin Looks like it's failing CI

bschaeublin:
@JamesHenry sorry, I guess there was a merge conflict in https://github.com/angular-eslint/angular-eslint/pull/2187/commits/e5b64ac0db0311cf022664919896885f6d12e991. Solved it and it seems now everything [successful](https://cloud.nx.app/runs/iKXyNz5p0C).

packages/eslint-plugin/tests/rules/sort-keys-in-type-decorator/cases.ts
bschaeublin: > Sorry for the delay I moved across the country I live in and it's been a very busy time at work

No worries - thank you for maintaining and keeping this up here :)

JamesHenry (approved):

nickbullock:
hey @JamesHenry!
could we please release this rule? 
thank you guys @bschaeublin @JamesHenry

PR #2194, opened 2025-01-09
feat: export base configs

Badisi:
A minimal ruleset that sets only the required **parser** and **plugin** options needed to run `angular-eslint` could be useful when one wants to share his own rules configuration.

Exporting the existing `base` configurations would allow this.


packages/angular-eslint/src/index.ts
```diff
@@ -72,15 +76,17 @@ both CJS and ESM in very natural ways.
 */
 export default {
   configs,
+  tsParser: parser,
```
JamesHenry, line 79: Why did you add this? I don't want to export a parser from another package, that should be imported directly

packages/angular-eslint/src/index.ts
```diff
@@ -72,15 +76,17 @@ both CJS and ESM in very natural ways.
...
   tsPlugin,
   templateParser,
   templatePlugin,
   processInlineTemplates,
 };
 export {
   configs,
+  tsParser: parser,
```
JamesHenry, line 87: Same comment

JamesHenry (changes requested):

packages/angular-eslint/src/index.ts
Badisi: Only for convenience and consistency.

---

Because there were no **base** configs exported I had to rewrite them in my code.

Importing your package gave me some insights about what your package was using, thanks to autocompletion:
```ts
import angularPlugin from 'angular-eslint';

angularPlugin.***
```
For instance, I easily found that there were a `templatePlugin` and a `tsPlugin`.
And also a `templateParser`, but no `tsParser`...
I didn't find any info in the doc about the parser that was used, so I had to dig in the code to find out I was supposed to use the parser from `typescript-eslint`.
So I thought it would be convenient and consistent to also export a `tsParser` property.

packages/angular-eslint/src/index.ts
JamesHenry: This isn't valid syntax, it's causing CI to fail

JamesHenry:
Closing for now based on lack of activity, please feel free to reopen when you have time

Issue #997, opened 2022-05-18
[@angular-eslint/template/i18n] Request to add more default ignoreAttributes and ignoreTags


SamuelT-Beslogic:
The following attributes seem like they should not be translated:
### <head> (not Angular-template translatable)
- meta
- link
- title

### No translatable attributes OR content is unlikely to be translatable
- mat-drawer
- mat-icon (the content is the name of the icon to use)
- ng-container
- :svg:polygon

### HTML
- input[accept]
- img[ngSrc]

### Angular Material
- cdkColumnDef
- mat-accordion[displayMode]
- mat-dialog-actions[align]
- mat-form-field[appearance]
- mat-form-field[floatLabel]
- matTooltipPosition
- panelClass

```JSON
{
  "rules": {
    "@angular-eslint/template/i18n": "error",
  }
}
```

```HTML
<mat-icon role="img"></mat-icon>
```

**Versions**

| package                                   | version |
| ----------------------------------        | ------- |
| `@angular-eslint/eslint-plugin-template`  | `13.2.1` |
| `@angular-eslint/template-parser`         | `13.2.1` |
| `@typescript-eslint/parser`               | `5.25.0` |
| `ESLint`                                  | `8.15.0` |
| `node`                                    | `16.13.2` |

```sh
# Please run `npx ng version` in your project and paste the full output here:
 >  NX   Cannot find target 'version' for project 
```

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

Samuel-Therrien-Beslogic:
I see more defaults have been added, some of which were in my list. I've updated the issue description.

JamesHenry:
Hi Folks, I am going to close this one in favour of the new RFC for revamped i18n support: https://github.com/angular-eslint/angular-eslint/issues/2489

Please come and join the discussion and let's ensure that this issue is adequately addressed by the new effort.

Many thanks!

PR #2201, opened 2025-01-14
feat(eslint-plugin-template): [i18n] add additional default ignores

rgant:
Fix for #997

Found some issue with the current issue request:

1. I don't believe `displayMode is an `@Input` for `mat-accordion` in Angular Material 19
2. Adding non-void tags to ignore tags ignores all the attributes and text contents of those tags. Which is not always ideal.
   1. `mat-drawer` can just have text content. https://material.angular.io/components/sidenav/examples#sidenav-backdrop
   2. `mat-icon` can have `aria-label`s. https://material.angular.io/components/icon/examples#icon-overview
   3. `ng-container` is specifically recommended by Angular documentation for i18n. https://angular.dev/guide/i18n/prepare#translate-inline-text-without-html-element
   4. `title` is used in SVG and can be translated using i18n.
   5. `:svg:polygon` I am unaware of what is special about the `<polygon>` element that needs to be ignored.


rgant:
Also, the documentation links for the i18n docs on angular.dev are wrong. Easiest for me is to just correct that in the PR. However if you prefer separate requests then I will try to find time to do that.

rgant:
Also, I couldn't find developer setup documentation, nor contributing documentation. So I ran the `update-rule-docs` manually. Not sure if that was desired.

rgant:
All I care about is `NgOptimizedImage` but I found this (too?) large [issue](#997) and decided to try to resolve as much of it as I could. If there are any concerns then I can simplify this down to only the image changes easily.

JamesHenry:
Hi @rgant, I'm really sorry for the delay on this one. I really want to holistically reconsider the `i18n` rule and get the community's help to arrive at the best solution.

Therefore I created a new RFC for revamped i18n support: https://github.com/angular-eslint/angular-eslint/issues/2489

Please come and join the discussion and let's ensure that this issue is adequately addressed by the new effort.

Many thanks!

Issue #2265, opened 2025-02-26
[@angular-eslint/template/prefer-static-string-properties] The rule replaces where it is not needed


splincode:
```
$ git clone https://github.com/taiga-family/taiga-ui.git && cd taiga-ui
$ npm ci
// ... add to eslint.config.ts that rule for *.html 
```

```
$ npx eslint . --fix
```

Before

```html
 <code [style.color]="'var(--tui-background-accent-2-pressed)'"></code>
```

After

```html
 <code color="var(--tui-background-accent-2-pressed)"></code>
```

Before

```html
<input *ngSwitchCase="'boolean'" />
```

After

```html
<input *ngSwitchCase="boolean" />
```

Before

```
<td *tuiCell="'name'" />
```

After

```
<td *tuiCell="name" />
```

**Versions**

| package                                  | version |
| ---------------------------------------- | ------- |
| `@angular-eslint/eslint-plugin-template` | `19.x` |
| `@angular-eslint/template-parser`        | `19.x` |
| `@typescript-eslint/parser`              | `19.x` |
| `ESLint`                                 | `9.x` |
| `node`                                   | `22.x` |

```sh
Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1602.16
@angular-devkit/build-angular   16.2.16
@angular-devkit/core            16.2.16
@angular-devkit/schematics      16.2.16
@angular/cdk                    16.2.14
@angular/cli                    16.2.16
@schematics/angular             16.2.16
ng-packagr                      16.2.3
rxjs                            7.8.2
typescript                      5.0.4
webpack                         5.98.0
```

splincode:
@JamesHenry @mbriand-lucca Thank you about your fix. However what about that code

```
 <code [style.color]="'var(--tui-background-accent-2-pressed)'"></code>
```

I think is also valid and it cannot be changed to `color`

mbriand-lucca:
> [@JamesHenry](https://github.com/JamesHenry) [@mbriand-lucca](https://github.com/mbriand-lucca) Thank you about your fix. However what about that code
> 
> ```
>  <code [style.color]="'var(--tui-background-accent-2-pressed)'"></code>
> ```
> 
> I think is also valid and it cannot be changed to `color`

I haven’t fixed this problem yet. I have an idea: checking if the attribute name contains a '.', but I’m not sure if it’s reliable...

splincode:
@JamesHenry could you reopen please?

[...]

PR #2271, opened 2025-02-27
fix(prefer-static-string-properties): resolve bug with directives

mbriand-lucca:
We don't want to use this rule for directives like `*ngSwitchCase` or `*ngIf`

fixes https://github.com/angular-eslint/angular-eslint/issues/2265


mbriand-lucca:
I'm not sure if the code is great, feel free to let me know if there’s anything to improve :)

JamesHenry (approved):
Thank you!

PR #2273, opened 2025-02-28
fix(prefer-static-string-properties): exclude special attributes

mbriand-lucca:
Exclude some attributes : 

```
<my-component *ngSwitchCase="'foo'" />
```

```
<my-component [style.color]="'foo'" />
```

```
<my-component [attr.aria-label]="'foo'" />
```

```
<my-component [@fade]="'foo'" />
```

```
<my-component [class.foo]="'foo'" />
```

fixes: #2265

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/prefer-static-string-properties.md
+++ b/packages/eslint-plugin-template/docs/rules/prefer-static-string-properties.md
@@ -430,6 +430,32 @@ The rule does not have any configuration options.
 #### ✅ Valid Code
 
 ```html
+<my-component [class.foo]="'foo'" />
+```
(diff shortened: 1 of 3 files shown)


mbriand-lucca:
If anyone has a better solution, feel free to share... I haven't found anything better.

JamesHenry:
Thanks again @mbriand-lucca, unfortunately I didn't spot the commit/PR scope until too late, but this should have been `fix(eslint-plugin-template): [prefix-static-string-properties] exclude special attributes` based on our existing conventions and CHANGELOGs.

I'll try and get around to adding some tooling to prevent this in future

PR #2308, opened 2025-03-09
feat(eslint-plugin-template): Add [prefer-template-literal] rule

g-drouard:
Closes https://github.com/angular-eslint/angular-eslint/discussions/2297

Some questions/suggestions :

- I named the rule `prefer-template-literal` to avoid ambiguity instead of `prefer-template`. I think the suggested name `prefer-template-in-template` is so cumbersome. Do you have any other suggestion ?
- Should we report/fix string concatenation that doesn't result in a template literal? I tried to report/fix them.
- Should we add it to the recommended rules? I think we can wait and see if the rule doesn't report any problems.
- While trying to test the `@let` I saw that it was not handled. Is there an issue in progress?


reduckted:
> While trying to test the `@let` I noticed that was not handled. Is there an issue in progress?

I noticed the same thing while writing #2311 and added support for it to the template parser. I've pulled that commit out into a separate PR (#2312) so that it can be merged sooner and you can use it.

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
```diff
@@ -0,0 +1,129 @@
...
+
+            return fixes;
+          },
+        });
+      },
+    };
+  },
+});
+
+function getLeftSideFixs(fixer: RuleFixer, left: AST): readonly RuleFix[] {
```
reduckted, line 95:
```suggestion
function getLeftSideFixes(fixer: RuleFixer, left: AST): readonly RuleFix[] {
```

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
```diff
@@ -0,0 +1,129 @@
...
+  } else {
+    // Transform left side to template literal
+    return [
+      fixer.insertTextBeforeRange([start, end], '`${'),
+      fixer.insertTextAfterRange([start, end], '}'),
+    ];
+  }
+}
+
+function getRightSideFixs(fixer: RuleFixer, right: AST): readonly RuleFix[] {
```
reduckted, line 113:
```suggestion
function getRightSideFixes(fixer: RuleFixer, right: AST): readonly RuleFix[] {
```

reduckted (changes requested):
Very cool! I'm looking forward to using this.

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
```diff
@@ -0,0 +1,129 @@
...
+          messageId,
+          fix: (fixer) => {
+            // If both sides are literals, we can just remove the `+` sign and concatenate them
+            if (
+              left instanceof LiteralPrimitive &&
+              right instanceof LiteralPrimitive
+            ) {
+              return fixer.replaceTextRange(
+                [start, end],
+                `'${left.value}${right.value}'`,
```
reduckted, line 70: A few things about this:

1. I think the original type of quote should be kept.
```
{{ "a" + "b" }}
```
```diff
- Expected
+ Received
-     {{ "ab" }}
+     {{ 'ab' }}
```

2. If either string contains a single quote, it breaks.
```
{{ "doesn't " + "work" }}
```
```diff
- Expected
+ Received
-     {{ "doesn't work" }}
+     {{ 'doesn't work' }}
```

3. If either string escapes a backslash, the escaping is removed.
```
{{ 'doesn\'t ' + 'work' }}
```
```diff
- Expected
+ Received
-     {{ 'doesn\'t work' }}
+     {{ 'doesn't work' }}
```

When there's a mix of quote types (for example, the left uses `"` and the right uses `'`), it's probably simplest to use the quote type from the left string and make sure to escape any quotes in the right string that end up needing to be escaped. Alternatively, borrow some logic from the [`quotes`](https://eslint.org/docs/latest/rules/quotes#rule-details) rule's `avoidEscape` option and pick the quote type that won't need escaping. It depends how complicated you want to get with it. :)

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function getLeftSideFixes(fixer: RuleFixer, left: AST): readonly RuleFix[] {
+  const { start, end } = left.sourceSpan;
+
+  if (left instanceof TemplateLiteral) {
+    // Remove the end ` sign from the left side
+    return [fixer.removeRange([end - 1, end])];
+  } else if (isStringLiteralPrimitive(left)) {
+    // Transform left side to template literal
+    return [fixer.replaceTextRange([start, end], `\`${left.value}`)];
```
reduckted, line 103: This needs to escape backticks in the left value.
```
{{ 'left ` ' + `right` }}
```
```diff
- Expected
+ Received
-     {{ `left \` right` }}
+     {{ `left ` right` }}
```

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function getRightSideFixes(fixer: RuleFixer, right: AST): readonly RuleFix[] {
+  const { start, end } = right.sourceSpan;
+
+  if (right instanceof TemplateLiteral) {
+    // Remove the start ` sign from the right side
+    return [fixer.removeRange([start, start + 1])];
+  } else if (isStringLiteralPrimitive(right)) {
+    // Transform right side to template literal
+    return [fixer.replaceTextRange([start, end], `${right.value}\``)];
```
reduckted, line 121: This needs to escape backticks in the right value.
```
{{ `left` + ' ` right' }}
```
```diff
- Expected
+ Received
-     {{ `left \` right` }}
+     {{ `left ` right` }}
```

reduckted (changes requested):

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
g-drouard: After careful consideration and investigation, it appears very challenging to determine the quote type used by LiteralPrimitive.

I believe there are two solutions:

1. Add a `quote` option, which will be used to specify the type of quote to use when the fix doesn't result in a template literal.
2. Adopt the same behavior as the [`prefer-template rule`](https://eslint.org/docs/latest/rules/prefer-template#examples), and therefore only report/fix cases that result in a template literal to avoid this quote issue. As outlined in the documentation of the prefer-template rule, the [`no-useless-concat`](https://eslint.org/docs/latest/rules/no-useless-concat) rule handles static string concatenation and doesn't have an autofix. We could consider making another equivalent rule for templates.

What do you think?

reduckted: You should be able to get the quote by looking at the character in the source code at the start of the literal's source span:

```ts
const leftQuote = sourceCode.text.at(left.sourceSpan.start)!;assertion
const rightQuote = sourceCode.text.at(right.sourceSpan.start)!;assertion
let quote: string;
if (leftQuote === rightQuote) {
  quote = leftQuote;
} else {
  quote = ...; // Decide which quote type to use.
}
```

JamesHenry (changes requested):
Thanks @g-drouard! You should be able to rebase and fix up CI now

JamesHenry:
It might be fixed if I merge `main` in actually, I've done that now

JamesHenry:
So as you said I think the only thing remaining is `@let` coverage, which is parsed correctly now, thanks again!

g-drouard:
There is a bug in Angular with the `@let` and `track`.
I think we should wait for the fix before deploying the rule.

Issue : https://github.com/angular/angular/issues/60528

JamesHenry:
Cool thanks @g-drouard

reduckted:
Looks like the fix landed in 19.2.6.
https://github.com/angular/angular/releases/tag/19.2.6

packages/eslint-plugin-template/src/rules/prefer-template-literal.ts
```diff
@@ -0,0 +1,156 @@
...
+export type MessageIds = typeof messageId;
+export const RULE_NAME = 'prefer-template-literal';
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Ensure that template literals are used instead of concatenating strings or expressions',
```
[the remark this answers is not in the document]
g-drouard: Done, thx !

g-drouard:
I confirm that the bug has been fixed in `19.2.6`.

Should we wait for `19.3` before publishing this rule or should we add a dependency in `19.2.6`?

reduckted:
If I'm not mistaken, the bundled angular compiler is already at 19.2.7, so it will include the fix.
https://github.com/angular-eslint/angular-eslint/pull/2377

JamesHenry:
@reduckted Exactly, we don't need to worry about user setups down to the minor/patch because of that.

P.S. Sorry for my generally low activity levels on here recently, I've just completed an inter-state home move and it's been very chaotic. I'll be catching up on everything by the end of the week

JamesHenry (approved):
Thanks a lot!

PR #2327, opened 2025-03-18
fix(eslint-plugin-template): [i18n] Avoid exception in i18n rule with allowMarkupInContent=false

m-akinc:
The `i18n` rule in `eslint-plugin-template` results in an exception when the option `allowMarkupInContent` is set to `false` and a template includes an element with no `i18n` attribute. There was previously no test case without an `i18n` attribute. Fixed the issue and added a test case.


m-akinc:
Adding @JamesHenry as reviewer. If this is incorrect, please add the correct reviewer instead.

JamesHenry (approved):
Thank you!

m-akinc:
> Thank you!

My pleasure. @JamesHenry any chance we could get this cherry-picked into a new version 18 release?

m-akinc:
@JamesHenry I don't mean to be a pest, but in case you missed it or forgot, I wanted to ping you again about getting a Angular-18-compatible release with this fix?

PR #2335, opened 2025-03-22
fix(eslint-plugin-template): [no-any] handle key reads (#1701)

### LINKED ISSUE #1701: [@angular-eslint/template/no-any] does not throw error for property binding with square bracket notation
**Issue Description**
The angular-eslint/template/no-any rule does not throw an error when used in conjunction with property binding using square bracket notation. The issue arises when using the $any expression, as demonstrated in the following example:

```
<div
    class="impressions__content__card trq-shadow-bottom-sm trq-p-3 trq-bg-white trq-flex-fill trq-mr-1"
    *ngIf="receivedSelected && $any(attributeList)['NPSScore'] !== null">
```
In this case, the rule does not detect and report the potential problem with property binding using square brackets. However, it works as expected when using direct property access, such as [name]= $any(name).

**Expected Behavior**
The angular-eslint/template/no-any rule should identify and report property binding with square bracket notation, especially when used in conjunction with the $any expression.

undsoft:
Fixes #1701

Now properly handles cases like:

```
{{ $any(attributeList)['NPSScore'] }}
```


JamesHenry (approved):
Thank you!

Issue #2359, opened 2025-04-09
[label-has-associated-control] Allow overriding of default label inputs


cprestegard:
**Description and reproduction of the issue**

The [label-has-associated-control] rule hard codes the default inputs to check for on label elements.  It would be great to allow users to override the defaults.  This would allow projects to lint labels that rely on custom directives to hook into custom controls.

Suggestion to have the ability to declare the rule like this:

```json
{
  "rules": {
    "@angular-eslint/template/label-has-associated-control": [
      "error",
      {
        "controlComponents": [
          "my-custom-control"
        ],
        "labelComponents": [
          {
            "inputs": [
              "for",
              "htmlFor",
              "myCustomFor"
            ],
            "selector": "label"
          }
        ]
      }
    ]
  }
}
```

so that the following would be valid:

```html
<label [myCustomFor]="customId">Label</label>
<my-custom-control [id]="customId"></my-custom-control>
```

**Versions**

| package                                  | version |
| ---------------------------------------- | ------- |
| `@angular-eslint/eslint-plugin-template` | `19.3.0` |

PR #2360, opened 2025-04-09
fix(eslint-plugin-template): [label-has-associated-control] labelComponents should override default label inputs

cprestegard:
Example PR to address #2359


JamesHenry (changes requested):
See comment

PR #2383, opened 2025-04-20
feat(no-misused-signals): check use of signals in logical expressions

scj7t4:
Checks that signals used in logical expressions are invoked. For #2302

Thanks for your work on the project! This is my first attempt at a typescript eslint plugin, so feedback is very appreciated on what I messed up :)


packages/angular-eslint/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "angular-eslint",
-  "version": "19.3.0",
```
JamesHenry, line 3: All of these need to be reverted please

JamesHenry (changes requested):
TypeScript has always referred to this as uncalled function checks: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#uncalled-function-checks

So maybe `no-uncalled-signals` is a better name (given it doesn't check any other kind of "misuse")?

JamesHenry:
WDYT about the name @reduckted?

scj7t4:
Thanks for the feedback! I've reverted the package.json files and changed the name of the rule to no-uncalled-signals

JamesHenry:
@scj7t4 Thanks please see the CI failures they tell you what to do to fix them

scj7t4:
> @scj7t4 Thanks please see the CI failures they tell you what to do to fix them

Should be fixed

reduckted:
@scj7t4 You'll need to run these two commands to update some generated files:
```
pnpm update-rule-lists
pnpm update-rule-configs
pnpm update-rule-docs
```

packages/eslint-plugin/src/rules/no-uncalled-signals.ts
```diff
@@ -0,0 +1,58 @@
...
+        'Doing logic operations on signals will give unexpected results, you probably want to invoke the signal to get its value',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const services: ParserServicesWithTypeInformation =
+      ESLintUtils.getParserServices(context);
+
+    return {
+      '*.test[type=Identifier], *.test Identifier'(node: TSESTree.Identifier) {
```
reduckted, line 41: Nice use of `*.test`. 👍 

There are some cases that it doesn't find though:

```ts
const v = aSignal || true;
```
```ts
const v = aSignal ?? true;
```

reduckted (changes requested):

packages/eslint-plugin/src/rules/no-uncalled-signals.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (node.parent.type === AST_NODE_TYPES.CallExpression) {
+          return;
+        }
+
+        const type = services.getTypeAtLocation(node);
+        const identifierType = type.getSymbol()?.name;
+
+        if (identifierType && KNOWN_SIGNAL_TYPES.has(identifierType)) {
+          context.report({
```
reduckted, line 50: Since the most likely change to fix this problem will be to call the signal, it might be worth including a suggestion that adds `()` after the identifier. It won't auto-fix it, but it will provide a quick way for users to apply the code change to get rid of this problem.

JamesHenry: Agreed, it's the exact thing graphite suggested above https://github.com/angular-eslint/angular-eslint/pull/2383#discussion_r2083525002

packages/eslint-plugin/src/rules/no-uncalled-signals.ts
scj7t4: I'm struggling a bit with this one. If I add the matcher:

```
[type=LogicalExpression] Identifier
```

I can find these cases. However, then I can't find a variation of the first matcher (or my new matcher) that will not match for both in some cases. Is it okay to have a rule report the same finding 2 times?

I've pushed the new version with the extra matcher.

packages/eslint-plugin/src/rules/no-uncalled-signals.ts
scj7t4: Added a suggestion and adjusted the tests to verify the suggestion.

JamesHenry (approved):

JamesHenry:
Thanks so much for this!

danielvieira1:
can we use this rule in the template for control flow for example?

invalid
`@if(signal) {`

valid
`@if(signal()) {`

JamesHenry:
@danielvieira1 No, template rules work totally differently in a different plugin, that would require a separate rule

PR #2398, opened 2025-04-28
feat(eslint-plugin-template): [no-nested-tags] add rule

sod:
## What

Shows this lint error on nested a or p tags:
CleanShot 2025-04-28 at 16 13 59@2x

## Why

If a browser encounters nested a or p tags, it'll de-nest them. It's possible via `appendChild` though - so if you use angular without Ssr, this won't apply to you.

With the addition of https://angular.dev/guide/incremental-hydration for Ssr this is a problem though. If you have a component that creates:

`[<a>{<a>()</a>}</a>]`

Then Ssr will send:

`[<a>{</a><a>()</a>}]`

Which is a DOM mismatch and incremental-hydration will throw an error:

> ERROR RuntimeError: NG0500: During hydration Angular expected <a> but the node was not found.

and not successfully bootstrap.

This lint rule tries to prevent this. Of course not across components, but at least if it happens inside a single template.

related: https://github.com/angular/angular/issues/51399

You can observe this behavior also if you try to throw this into a page via the dev toolbar:

https://github.com/user-attachments/assets/e266ff83-b0a0-4316-95b6-b5c92f4a775e


sod:
Here a reproduction for what I mean https://github.com/sod/issue-ssr-reproduction-2398 

CleanShot 2025-04-29 at 13 42 08@2x

JamesHenry:
Thanks @sod I updated the PR title to match the conventions, but I recommend running the commands from the `ci.yml` on your machine because you don't match the required formatting and you haven't generated the required docs yet

JamesHenry:
@sod It's because you are a first-time contributor, I have to grant permission on the first PR for every run

Issue #2347, opened 2025-03-27
(eslint-plugin-template) suggestion : template-parser as regular dependency


mho22:
I was trying to use the `eslint-plugin-template` isolated from the other packages to make my flat config work for the templates only.

I had to manually add the `template-parser` since it is a `devDependency` rather than a `dependency` of `eslint-plugin-template`.

Shouldn't `@angular-eslint/template-parser` be a direct dependency in order for it to work properly ?

PR #2417, opened 2025-05-11
fix: set `template-parser` dependency as dependency instead of dev dependency in `eslint-plugin-template`

mho22:
Based on issue #2347

When installing `eslint-plugin-template` separately, it currently requires manually installing `@angular-eslint/template-parser`. However, according to the `README`:

> [@angular-eslint/eslint-plugin-template](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/README.md) - An ESLint-specific plugin which, when used in conjunction with @angular-eslint/template-parser, allows for Angular template-specific linting rules to run.

Given this dependency relationship, it might make sense to list `template-parser` under `dependencies` instead of `devDependencies`.

### DIFF
--- a/packages/eslint-plugin-template/package.json
+++ b/packages/eslint-plugin-template/package.json
@@ -19,12 +19,12 @@
   ],
   "dependencies": {
     "@angular-eslint/bundled-angular-compiler": "workspace:*",
+    "@angular-eslint/template-parser": "workspace:*",
     "@angular-eslint/utils": "workspace:*",
     "aria-query": "5.3.2",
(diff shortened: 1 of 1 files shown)


JamesHenry:
Thanks @mho22 but it isn't a dependency, it's a peer dependency. There is no direct invocation of template-parser anywhere in the logic of eslint-plugin-template, and I've double checked around the ecosystem, and the correct way to express this is indeed via a peer dep.

I've opened that PR here: https://github.com/angular-eslint/angular-eslint/pull/2487

Many thanks

PR #2430, opened 2025-05-16
Feature/refactor localize validation

igord:


JamesHenry:
@tcorral how do you feel about this approach as an alternative?

tcorral:
> @tcorral how do you feel about this approach as an alternative?

@JamesHenry It looks good to me.

JamesHenry (changes requested):
Please run the CI commands locally before pushing, there are still multiple failures

PR #2448, opened 2025-05-24
feat!: angular-eslint v20

### LINKED ISSUE #2206: [@angular-eslint/prefer-standalone] The rule `prefer-standalone` should not be "fixable"

Our team manages hundreds of Angular applications; for each new version of Angular, we run Angular CLI's `ng update`. One of the v19 migration schematics from Angular sets `standalone: false` for any non-standalone components in the project. This works fine.

After the update, we run `eslint --fix` for all of the applications, and because `prefer-standalone` was recently added to the "recommended" ESLint config, it's "fix" script essentially undoes the update schematic and removes the `standalone: false` from the components, putting every application into a failure state (the applications will not build because their components are now considered standalone, but they aren't setup to work that way yet).

As I understand it, a "fixable" rule should make changes to the code that won't put it into a broken state? If it was truly fixable, wouldn't it also add the `imports` array, and remove the component from any modules that declare it?

I'm suggesting that`angular-eslint` removes the ["fix" script from the rule](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/src/rules/prefer-standalone.ts#L48) so that projects running `--fix` won't end up in a broken state.

**Versions**

| package                         | version |
| ------------------------------- | ------- |
| `@angular-eslint/eslint-plugin` | `19.0.2` |
| `@typescript-eslint/parser`     | `8.20.0` |
| `ESLint`                        | `9.18.0` |
| `node`                          | `20.12.2` |

```sh
Angular CLI: 19.0.7
Node: 20.12.2
Package Manager: npm 10.5.0
OS: darwin arm64

@angular-devkit/architect       0.1900.7
@angular-devkit/build-angular   19.0.7
@angular-devkit/core            19.0.7
@angular-devkit/schematics      19.0.7
@angular/cdk                    19.0.5
@angular/cli                    19.0.7
@schematics/angular             19.0.7
typescript                      5.7.3
zone.js                         0.15.0

```

JamesHenry:
TODOs

- [x] Drop support for EOL Node 18, align Node support with Angular v20 => `^20.19.0 || ^22.12.0 || ^24.0.0`
- [x] Drop support for TypeScript prior to `5.8` to align with Angular v20
- [ ] Switch to Angular v20 (https://github.com/angular-eslint/angular-eslint/pull/2449)
- [x] Throw parse errors by default (#2255)
- [x] Do not auto-fix `prefer-standalone`
- [x] Remove `@angular-eslint/component-class-suffix` from recommended to align with v20 style guide and CLI code generation
- [x] Remove `@angular-eslint/directive-class-suffix` from recommended to align with v20 style guide and CLI code generation
- [ ] ... more ...

Fixes #2206


JamesHenry:
@reduckted Please see https://github.com/angular-eslint/angular-eslint/pull/2448/commits/51691cfc18c41b89d511bb79f4644bbe4cea53e5 an existing test case of your causes a parse error now after your PR to switch them on by default

reduckted:
> @reduckted Please see [51691cf](https://github.com/angular-eslint/angular-eslint/commit/51691cfc18c41b89d511bb79f4644bbe4cea53e5) an existing test case of yours causes a parse error now after your PR to switch them on by default

Fixed in #2450.

JamesHenry:
@claude that makes sense, I had messed up the previous rebase and so the relevant changes were not present. They now are. Let's do it again. Run `pnpm update-e2e-snapshots` with a large timeout and commit the results to the .snap files ONLY. Do not commit any modified package.json files.

JamesHenry:
I have run out of time before I am on vacation for a week without a laptop so this needs to go out now :)

PR #2460, opened 2025-05-26
test(eslint-plugin-template): add test cases for prefer-template-literal reported issues

JamesHenry:


reduckted:
This hasn't added a test case where multiple concatenations that include template literals are used. It's added a test case were multiple string literals are joined together (`'first' + 'second' + 'third'`), but not something like `{{ a + ' ' + b + ' ' + c }}`.

JamesHenry:
@reduckted Yes I know

PR #2464, opened 2025-05-26
feat!: switch to angular v20

JamesHenry:


JamesHenry:
After a quick check it looks like the Angular compiler now strips parens in many cases (https://github.com/angular/angular/pull/60169/files) which seems to be causing issues with multiple rules. Needs more investigation.

JamesHenry:
@claude it is because I did not allow you a deep enough git checkout. Let's try again. Run `pnpm update-e2e-snapshots` and commit the results to the branch

PR #2487, opened 2025-05-31
fix(eslint-plugin-template): set template-parser as peer dependency

JamesHenry:
Closes https://github.com/angular-eslint/angular-eslint/issues/2347


mho22:
@JamesHenry Thank you.

PR #2488, opened 2025-05-31
fix(eslint-plugin-template): any valid DOM element with role button is interactive

JamesHenry:


packages/eslint-plugin-template/docs/rules/click-events-have-key-events.md
```diff
@@ -547,6 +547,9 @@ interface Options {
 <a href="#" (click)="onClick()"></a>
 <a [attr.href]="href" class="anchor" (click)="onClick()"></a>
 <a [routerLink]="'route'" (click)="onClick()"></a>
+<div role="button" (click)="doSomething()"></div>
+<span role="button" (click)="doSomething()"></span>
+<p role="button" (click)="doSomething()"></p>
```
JamesHenry, line 552: FYI @sandikbarr I am rushing this one a little bit to get the fix out in 19.x, I wasn't 100% if existing utils allowed me to figure this out so I put in a hardcoded check for role of button

sandikbarr: @JamesHenry those elements with role button won’t behave like a button that would have built in keyboard support.

JamesHenry: For sure, but for this rule, I think it makes sense not to report? It will be caught by other rules for not having a tabindex etc

JamesHenry: To fix this lint issue you would have to add those event handlers and then you would process the event multiple times

sandikbarr: That’s a fair point. Interactive-supports-focus will catch it for not being focusable, but interactive-supports-focus won’t enforce keyboard support.

PR #2501, opened 2025-06-06
feat(eslint-plugin-template): [no-interpolation-in-attributes] add fixer

niwolf:
Adds a fixer to "eslint-plugin-template: no-interpolation-in-attributes". 
The fixer works for cases with full interpolation.


JamesHenry:
Thanks @niwolf you'll need to please regen the docs for this change, our docs are generated from our tests

```
pnpm update-rule-docs
```

JamesHenry:
Oh actually I might be wrong because you didn't actually change any sources, only descriptions and outputs

niwolf:
Thank you! Just pushed them, since autofix should work now, shouldn't it? 😃

JamesHenry:
Oh yeah we add that, also to rule lists so you need to regen those too

niwolf:
Should be done. If anything is missing, please let me know.

JamesHenry:
Thanks!

niwolf:
Thank you too for your fast reponse! Appreciate it!

PR #2514, opened 2025-06-11
fix(eslint-plugin): [no-output-on-prefix] update style guide link

vsantele:
There was a broken link for the rule @angular-eslint/no-output-on-prefix.


JamesHenry:
@vsantele Documentation is auto-generated from the rules, so the rule source needs to be updated and then the rule documentation regenerated with `pnpm update-rule-docs`.

@claude please update this PR to address that

JamesHenry:
Oh Claude can't operate on your fork, please kindly do it yourself @vsantele

vsantele:
Oh I wasn't aware of that! It's good to know, I'll fix that later today (since Claude doesn't work for me ^^)

JamesHenry (changes requested):
Per above

vsantele:
I updated the link in the source code, then ran `pnpm update-rule-docs`. The result is the same as my previous commit. So the source code and documentation are synchronised.

I chose to make a direct link to the page where the angular doc talks about the event name, and not the part in the [style guide](https://angular.dev/style-guide#naming-component-and-directive-members) which only redirects to the component/outputs docs.

JamesHenry:
Sorry forgot you need to update the rule lists for this kind of change too

JamesHenry (approved):
Thanks!

PR #2531, opened 2025-06-18
fix(eslint-plugin-template): [no-interpolation-in-attributes] use keySpan as attr. range and sourceSpan as replacement range

niwolf:
Fix for released version 19.8: https://github.com/angular-eslint/angular-eslint/pull/2501

This PR fixes an issue in the no-interpolation-in-attributes rule where attributes using attr. (e.g. `attr.data-*`) were incorrectly autofixed by removing the attr. prefix. The fixer now correctly preserves the full attribute name, ensuring proper binding syntax like `[attr.data-example]="..."`.

If this fix is valid, I'd appreciate it if it could still be included in a v19 release, if possible.

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/no-interpolation-in-attributes.md
+++ b/packages/eslint-plugin-template/docs/rules/no-interpolation-in-attributes.md
@@ -112,6 +112,33 @@ interface Options {
 
 <br>
 
+#### Default Config
+
+```json
+{
+  "rules": {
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks @niwolf as a special exception to my usual policy, I backported this one as `19.8.1`

niwolf:
Thank you very much! @JamesHenry

PR #2586, opened 2025-07-16
fix(eslint-plugin-template): [no-interpolation-in-attributes] disallow extra properties in options

andreww2012:
[`prefer-contextual-for-variables`](https://github.com/angular-eslint/angular-eslint/blob/e7e2e2876e6454e83e8bc9115fe1c856403c2b13/packages/eslint-plugin-template/src/rules/prefer-contextual-for-variables.ts#L68) rule currently allows extra properties to be passed in options object, which should not be allowed. This makes it easier for typos in rule options to go unnoticed.

This PR simply disallows extra properties in this rule's schema.


JamesHenry:
Thanks!

PR #2605, opened 2025-08-04
feat: add no-case-transformations rule for Angular templates

andreasnicolaou:


JamesHenry:
Thank you @andreasnicolaou!

PR #2712, opened 2025-10-26
fix(eslint-plugin-template): support uppercase tags

undsoft:
Probably no one will do it, but technically uppercase HTML tags is a valid syntax and rules shouldn't break if they encounter them.
This PR updates rules to support uppercase tags.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-uncalled-signals.md
+++ b/packages/eslint-plugin/docs/rules/no-uncalled-signals.md
@@ -57,12 +57,9 @@ The rule does not have any configuration options.
 #### ❌ Invalid Code
 
 ```ts
-const aSignal = createSignal();
-if (aSignal) {
-    ~~~~~~~
-}
-declare function createSignal(): Signal<boolean>;
-interface Signal<T> {}
+let a: Signal<boolean>;
+let b = a ? 1 : 2;
+        ~
 ```
 
 <br>
@@ -88,12 +85,9 @@ interface Signal<T> {}
 #### ❌ Invalid Code
 
 ```ts
-const aSignal = createSignal(false);
-if (aSignal || true) {
-    ~~~~~~~
-}
-declare function createSignal(): Signal<boolean>;
-interface Signal<T> {}
+let a: Signal<boolean>;
+do {} while (a);
+             ~
 ```
 
 <br>
@@ -119,12 +113,9 @@ interface Signal<T> {}
 #### ❌ Invalid Code
 
 ```ts
-const aSignal = createSignal("hello");
-if (aSignal == "hello") {
-    ~~~~~~~
-}
-declare function createSignal(): Signal<boolean>;
-interface Signal<T> {}
+let a: Signal<boolean>;
+for (let i = 0; a; i++) { }
+                ~
 ```
 
 <br>
@@ -150,12 +141,9 @@ interface Signal<T> {}
 #### ❌ Invalid Code
 
 ```ts
-const aSignal = createSignal();
-if (false || (aSignal ?? true)) {
-              ~~~~~~~
-}
-declare function createSignal(): Signal<boolean>;
-interface Signal<T> {}
+let a: Signal<boolean>;
+if (a) { }
+    ~
 ```
 
 <br>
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/alt-text.ts
```diff
@@ -28,7 +28,7 @@ export default createESLintRule<Options, MessageIds>({
     const parserServices = getTemplateParserServices(context);
 
     return {
-      'Element[name=/^(img|area|object|input)$/]'(node: TmplAstElement) {
+      'Element[name=/^(img|area|object|input|IMG|AREA|OBJECT|INPUT)$/]'(node: TmplAstElement) {
```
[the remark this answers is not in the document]
undsoft: Makes sense. Fixed.

JamesHenry:
Thanks, I've rebased your branch so that you get the latest changes.

I've added a new `hideFromDocs` setting that can be applied to test cases after I saw this PR. I really don't want all these uppercase tags going into the docs, we can support them, sure, but we don't need to publicize their usage.

In case a test is already using the object notation it's easy, but for those that are currently just strings, here's the conversion you need to do:

image

(And don't forget to run formatting every time you make changes)

undsoft:
@JamesHenry Can you also update `convertAnnotatedSourceToFailureCase` so that it accepts `settings` and `hideFromDocs`?

JamesHenry:
@undsoft Please just apply it here, it doesn't need to land in advance, sorry I thought they shared what I had quickly changed

JamesHenry:
@undsoft Merged your other one, thanks again, caused inevitable conflicts with this one

undsoft:
@JamesHenry 
Sorry, I don't quite understand what you mean.
I've updated `valid` cases to use `hideFromDocs`, but if you want to hide `invalid` cases too, either `convertAnnotatedSourceToFailureCase` needs to be updated to support `settings` or I need to do something like:
```
{
  ...convertAnnotatedSourceToFailureCase({
    ...
  }),
  settings: {
    hideFromDocs: true,
  }
}
```

JamesHenry:
Yes, _you_ should update it 😅 (that's what I meant by "Please just apply it here") I don't need to do it in a separate PR, just make the change you need and I will review in one go

undsoft:
@JamesHenry Okay, done.

packages/eslint-plugin-template/src/rules/prefer-ngsrc.ts
```diff
@@ -30,7 +30,7 @@ export default createESLintRule<Options, MessageIds>({
     const parserServices = getTemplateParserServices(context);
 
     return {
-      'Element[name=img]'(element: TmplAstElement) {
+      'Element[name=/^(img|IMG)$/]'(element: TmplAstElement) {
```
JamesHenry:
```suggestion
      'Element[name=/^img$/i]'(element: TmplAstElement) {
```

packages/eslint-plugin-template/src/rules/no-nested-tags.ts
```diff
@@ -28,7 +28,7 @@ export default createESLintRule<Options, MessageIds>({
     const parserServices = getTemplateParserServices(context);
 
     return {
-      'Element[name=/^(p|a)$/]'(node: TmplAstElementWithAncestor) {
+      'Element[name=/^(p|a|P|A)$/]'(node: TmplAstElementWithAncestor) {
```
JamesHenry:
```suggestion
      'Element[name=/^(p|a)$/i]'(node: TmplAstElementWithAncestor) {
```

packages/eslint-plugin-template/src/rules/no-distracting-elements.ts
```diff
@@ -26,7 +26,7 @@ export default createESLintRule<Options, MessageIds>({
     const parserServices = getTemplateParserServices(context);
 
     return {
-      'Element[name=/^(blink|marquee)$/]'({
+      'Element[name=/^(blink|marquee|BLINK|MARQUEE)$/]'({
```
JamesHenry:
```suggestion
      'Element[name=/^(blink|marquee)$/i]'({
```

JamesHenry:
Thanks @undsoft, it's looking good, just the last few regex flags

JamesHenry (approved):

PR #2769, opened 2025-11-11
fix(eslint-plugin): false negatives in pipe-prefix, component-selector and directive-selector when linting selectors that only contain a prefix

### LINKED ISSUE #2768: False negatives in multiple rules when linting selectors that only contain a prefix
The following tests will not report an error, as they should:

**pipe-prefix**
```ts
  description: 'should fail when Pipe has no selector after the prefix',
  annotatedSource: `
      @Pipe({
        name: 'ng'
              ~~~~
      })
      class Test {}
    `,
  messageId,
  options: [{ prefixes: ['ng'] }],
  data: { prefixes: '"ng"' },
}),
```

**component-selector** and **directive-selector**
```ts
  description: `should fail if a selector has nothing, but the prefix in camelCase`,
  annotatedSource: `
    @Component({
      selector: 'app'
                ~~~~~
    })
    class Test {}
    `,
  messageId: messageIdStyleFailure,
  options: [{ type: 'element', prefix: 'app', style: 'camelCase' }],
  data: { style: 'camelCase' },
}),
```

undsoft:
Fixes #2768.

Will report a separate error for clarity.


JamesHenry:
@undsoft Please fix the PR title to be clear enough to be a merge candidate

undsoft:
@JamesHenry Done.

PR #2776, opened 2025-11-13
test(eslint-plugin): add failed tests for eslint extends type compatibility

pumano:


pumano:
@JamesHenry I found that extends not supported properly, tests for template config and ts config provided.

JamesHenry:
Thank @pumano are you going to try and address it?

pumano:
> Thank @pumano are you going to try and address it?

No, I really don't know how it should be (but should support defineConfig properly), but I just provide repro for bug in types

JamesHenry:
Addressed in #2793

PR #2827, opened 2025-12-08
feat(eslint-plugin-template): add prefer-class-binding rule to enforce class binding over ngClass

mistrykaran91:
Closes #2824, #2221, #2509


JamesHenry (changes requested):
This PR does not align with the position of the Angular documentation, it's pretty telling that there is no link to official messaging here to back up the rule description.

> NOTE: Class bindings do not support space-separated class names in a single key. They also don't support mutations on objects as the reference of the binding remains the same. If you need one or the other, use the [ngClass](https://angular.dev/api/common/NgClass) directive.

https://angular.dev/guide/templates/binding#css-class-and-style-property-bindings

Any rule disallowing the use of NgClass where class bindings could be used would have to take this important message from the docs into account and always allow such usage.

"Under soft deprecation" and "remains supported for backward compatibility" is not appropriate language to use when the documentation actively tells you to use the feature in specific cases.

`prefer-class-binding` _is_ an appropriate rule name, but right now the implementation and documentation need to change

packages/eslint-plugin-template/src/rules/prefer-class-binding.ts
```diff
@@ -0,0 +1,42 @@
...
+export type MessageIds = 'preferClassBinding';
+export const RULE_NAME = 'prefer-class-binding';
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Disallows the use of ngClass in HTML templates',
+      recommended: 'recommended',
```
JamesHenry: We do not add to the recommended ruleset between majors, and this rule is certainly not ready for that anyway

mistrykaran91:
> This PR does not align with the position of the Angular documentation, it's pretty telling that there is no link to official messaging here to back up the rule description.
> 
> > NOTE: Class bindings do not support space-separated class names in a single key. They also don't support mutations on objects as the reference of the binding remains the same. If you need one or the other, use the [ngClass](https://angular.dev/api/common/NgClass) directive.
> 
> https://angular.dev/guide/templates/binding#css-class-and-style-property-bindings
> 
> Any rule disallowing the use of NgClass where class bindings could be used would have to take this important message from the docs into account and always allow such usage.
> 
> "Under soft deprecation" and "remains supported for backward compatibility" is not appropriate language to use when the documentation actively tells you to use the feature in specific cases.
> 
> `prefer-class-binding` _is_ an appropriate rule name, but right now the implementation and documentation need to change

Then do you need to remove this rule and close the PR ?

JamesHenry:
@mistrykaran91 It can be merged if the implementation and description are changed to align with the documentation. It should not flag on cases (and have unit tests to prove it) that referenced by the documentation as use-cases for ngClass, it should only flag on cases explicitly where regular class bindings could have been used instead

packages/eslint-plugin-template/src/rules/prefer-class-binding.ts
mistrykaran91: Removed

mistrykaran91:
> @mistrykaran91 It can be merged if the implementation and description are changed to align with the documentation. It should not flag on cases (and have unit tests to prove it) that referenced by the documentation as use-cases for ngClass, it should only flag on cases explicitly where regular class bindings could have been used instead

@JamesHenry : I have updated the implementation and added the test cases for the same.

Edit: @JamesHenry is there anything else also needs to updated in this one :)

JamesHenry (approved):
Thank you!

PR #2879, opened 2026-01-30
fix(eslint-plugin): [no-uncalled-signals] add check `InputSignalWithTransform`

g-drouard:
`no-uncalled-signals` doesn't report missing signal calls when `input` use `transform`.

### Works

```typescript
  required = input(true);
  
  test(): void {
    // no-uncalled-signals: Error reported
    if (this.required) {
    ...
    }
  }
```

### Doesn't work

```typescript
  required = input(true, { transform: booleanAttribute });
  
  test(): void {
    // no-uncalled-signals: No error reported
    if (this.required) {
    ...
    }
  }
```


JamesHenry:
Thank you!

Issue #2881, opened 2026-02-01
New rule: ensure that computed returns value


undsoft:
I have recently forgotten to add a return statement to one of my `computed` functions and nothing complained about it – it just didn't work correctly.
I spent some time trying debugging the issue.

Example code (no compile or runtime errors):
```
user = signal({
  firstName: 'Lightning',
  lastName: 'McQueen',
});

fullName = computed(() => {
  // Oops, I forgot to add return
  `${this.user().firstName} ${this.user().lastName}`;
});

Hi, {{ fullName() }},
```

I'd say that this is something that should be prevented on Typescript level, but Angular team disagrees: https://github.com/angular/angular/issues/66720

`@typescript-eslint/no-unused-expressions` catches some cases of this, but only very basic, the moment another signal is called, it cannot help.

PR #2882, opened 2026-02-01
feat(eslint-plugin): add computed-must-return rule

undsoft:
Fixes: #2881

### DIFF
--- a/packages/angular-eslint/src/configs/ts-recommended.ts
+++ b/packages/angular-eslint/src/configs/ts-recommended.ts
@@ -16,7 +16,6 @@ export default (
   {
     name: 'angular-eslint/ts-recommended',
     rules: {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/computed-must-return.ts
```diff
@@ -0,0 +1,83 @@
...
+export type Options = [];
+export type MessageIds = 'computedMissingReturn';
+export const RULE_NAME = 'computed-must-return';
+
+export default createESLintRule<[], MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
+    docs: {
+      recommended: 'recommended',
```
JamesHenry: Please remove, we do not make changes to recommended rules outside of major versions

PR #2885, opened 2026-02-02
feat(eslint-plugin-template): add `no-non-null-assertion` rule

g-drouard:
Add equivalent of https://typescript-eslint.io/rules/no-non-null-assertion/ for Angular templates, without suggestions.


packages/eslint-plugin-template/src/rules/no-non-null-assertion.ts
```diff
@@ -0,0 +1,98 @@
...
+      context.report({
+        messageId: 'noNonNullAssertion',
+        loc: {
+          start: sourceCode.getLocFromIndex(start),
+          end: sourceCode.getLocFromIndex(end),
+        },
+      });
+    };
+
+    /**
```
JamesHenry: This feels like it might be a visitor keys issue if the expected nodes aren't being visited, please can you try adjusting those in the template-parser to achieve the same results without imperative traversal in the rule first?

Issue #1875, opened 2024-06-08
[eqeqeq] error after update to v18


DerHerrGammler:
**Description and reproduction of the issue**

Since the update to v18 I got a error from the `eqeqeq` rule. I am not sure if thats a problem of the rule or of the rule or the template parser itself. But here is the verbose stacktrace:

```sh
Cannot destructure property 'source' of '(0 , get_nearest_node_from_1.getNearestNodeFrom)(...)' as it is null.
Occurred while linting C:\workspace\[...].component.html:NaN
Rule: "@angular-eslint/template/eqeqeq"
TypeError: Cannot destructure property 'source' of '(0 , get_nearest_node_from_1.getNearestNodeFrom)(...)' as it is null.
Occurred while linting C:\workspace\[...].component.html:NaN
Rule: "@angular-eslint/template/eqeqeq"
    at getFix (C:\workspace\node_modules\@angular-eslint\eslint-plugin-template\dist\rules\eqeqeq.js:80:13)
    at Object.fix (C:\workspace\node_modules\@angular-eslint\eslint-plugin-template\dist\rules\eqeqeq.js:66:53)
    at normalizeFixes (C:\workspace\node_modules\eslint\lib\linter\report-translator.js:194:28)
    at C:\workspace\node_modules\eslint\lib\linter\report-translator.js:224:22
    at Array.map (<anonymous>)
    at mapSuggestions (C:\workspace\node_modules\eslint\lib\linter\report-translator.js:218:10)
    at C:\workspace\node_modules\eslint\lib\linter\report-translator.js:366:55
    at Object.report (C:\workspace\node_modules\eslint\lib\linter\linter.js:1047:41)
    at Binary[operation=/^(==|!=)$/] (C:\workspace\node_modules\@angular-eslint\eslint-plugin-template\dist\rules\eqeqeq.js:51:25)
    at ruleErrorHandler (C:\workspace\node_modules\eslint\lib\linter\linter.js:1076:28)
```

```JSON
{
  "rules": {
      "@angular-eslint/template/eqeqeq": [
          "error",
          {
              "allowNullOrUndefined": false,
          },
      ],
  }
}
```

I actually got two components where this error occures:  
First:
```HTML
<ng-container *transloco="let t">
    <mat-card>
        <mat-card-header>
            <mat-card-title>
                <span>{{ t("dash.welcome", { email: displayName() }) }}</span>
            </mat-card-title>
        </mat-card-header>
        <mat-card-content>
[...]

PR #2922, opened 2026-03-06
fix(eslint-plugin-template): [eqeqeq]  could not destructure null

githubuser578:
`getNearestNodeFrom` can return `null` and thus I currently get the following error for nodes like:

```html
<div [class.foo]="items?.length == 1"></div>
```
```
Error when running ESLint: Cannot destructure property 'source' of '(0 , get_nearest_node_from_1.getNearestNodeFrom)(...)' as it is null.
Occurred while linting some.html:30
Rule: "@angular-eslint/template/eqeqeq"
```

### DIFF
--- a/packages/eslint-plugin-template/src/rules/eqeqeq.ts
+++ b/packages/eslint-plugin-template/src/rules/eqeqeq.ts
@@ -113,12 +113,13 @@ const getFix = ({
   sourceCode: Readonly<TSESLint.SourceCode>;
   fixer: TSESLint.RuleFixer;
 }): TSESLint.RuleFix | null => {
-  const { source, ast } = getNearestNodeFrom(
-    node,
-    isASTWithSource,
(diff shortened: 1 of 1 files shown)


githubuser578:
This fix should also solve https://github.com/angular-eslint/angular-eslint/issues/1875

JamesHenry (changes requested):
@githubuser578 Hopefully you saw the CI failure

githubuser578:
@JamesHenry I've removed the trailing whitespace character

JamesHenry:
@githubuser578 As before, the Nx Cloud comment is showing you why the formatting is failing and how to fix it https://github.com/angular-eslint/angular-eslint/pull/2922#pullrequestreview-4067865750

githubuser578:
@JamesHenry Thanks for the hint. I've applied the formatting. Don't know why only trailing whitespaces were flagged the first time. But the formatting should be correct now.

PR #2926, opened 2026-03-07
fix(eslint-plugin): [no-uncalled-signals] properly handle statements with unary operators

BEGEMOT9I:
Previously, not all operations with unary operators were handled correctly. For example, it was possible to return an incorrectly handled signal from a method in a component:
```ts
type User = {
  id: string;
}

@Component({})
class TestComponent {
  user = signal<User | undefined>();

  hasNoUser = computed<boolean>(() => !this.user);
}
```
The same was for the assignment and other cases. This pull request fixes this behavior.

**Note:** the `delete` operator was omitted intentionally because signal properties can be optional

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-uncalled-signals.md
+++ b/packages/eslint-plugin/docs/rules/no-uncalled-signals.md
@@ -147,6 +147,65 @@ for (let i = 0; a; i++) { }
 #### ❌ Invalid Code
 
 ```ts
+function test(): boolean {
+  let a: Signal<boolean>;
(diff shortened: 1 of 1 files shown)


BEGEMOT9I:
@JamesHenry updated docs accordingly, thanks for pointing me out. Btw, why there is no `CONTRIBUTING.md` file?

JamesHenry:
@BEGEMOT9I they go stale and people ignore them anyway 😄 

Thanks a lot for this contribution!

PR #2969, opened 2026-03-23
fix(eslint-plugin): append fixes in prefer-on-push-change-detection, use-injectable-provided-in

undsoft:
Fixes #2968.

Also affects `use-injectable-provided-in`, where it will append `providedIn` to the end of properties in `@Injectable`.
This is not great, but most injectables don't have other properties.


JamesHenry:
Thank you, sorry for the delay

PR #2974, opened 2026-03-30
feat(builder): add apply-suppressions option

chronoB:
Fixes #2973


JamesHenry (changes requested):
Thanks for this! CI is failing please see the feedback

chronoB:
Ok the ci problems should be fixed. Sorry, I missed the required checks before submitting the pr. I have run the required checks mentioned in the CLAUDE.md now and they all pass.

JamesHenry:
@chronoB Looks like we still have some issues I'm afraid

chronoB:
Additional note: This PR does not support the suppressionLocation flag right now. So when applySuppressions is set it will always use the eslint.suppressions.json from the workspace root even if run in a project.

I just noticed this while working on the PR and didn't want to extend the scope without communication. I am fine with extending the scope or adding another PR that extends the functionality. Your call :)

JamesHenry:
Thank you @chronoB, sorry for the delay, feel free to open a dedicated PR for other options

PR #3009, opened 2026-04-22
feat(eslint-plugin): [inject-at-top] add rule requiring inject() at the top of the class

### LINKED ISSUE #3008: [new rule] enforce inject() at the top of the class to prevent runtime bugs
### Summary

I'd like to propose a new rule, `inject-at-top`, that enforces every `inject()` call to be declared at the very top of the class, before any other member.

This is not a stylistic rule. Its purpose is to prevent a real runtime bug that is easy to introduce and silently bypasses the TypeScript compiler.

### The bug

Class fields are initialized in the order they are written. That means `this.someService` only exists from the moment the line `private readonly someService = inject(SomeService);` has executed. Anything declared above that line which tries to read from `this.someService` will see `undefined` at construction time.

TypeScript does catch this when an earlier field references a later field *directly* (TS2729: "used before its initialization"). What it does not catch is the same bug behind one level of indirection, when the earlier initializer calls a getter or a method that itself reaches for the injected service. TS2729 is a local, shallow check: it does not walk into a getter body or a method body, so the moment the read of `this.someService` happens behind that indirection the compiler lets it through and the failure only shows up at runtime. Moving the getter or the method above or below the field that uses it does not help, the compiler does not trace through them either way.

Consider this component:

```ts
@Component({ /* ... */ })
class MyComponent {
  protected readonly pageTitle = `Dashboard ${this.fullName}`;

  private get fullName() {
    return `${this.userService.firstName} ${this.userService.lastName}`.trim();
  }

  private readonly userService = inject(UserService);
}
```

This compiles. The IDE shows no errors. TypeScript is happy. But at runtime, the `pageTitle` initializer runs first, calls the `fullName` getter, which dereferences `this.userService`, and `userService` has not been assigned yet. The class blows up with `Cannot read properties of undefined (reading 'firstName')`. The error points at the getter, not at the real cause (the `inject()` call sitting below it), and there is nothing about the code that looks obviously wrong.

Inlining the getter isn't really an option either: `fullName` composes two fields and trims the result, which is the kind of thing you actually want behind a name. The moment any non-trivial formatting, null-handling or permission check lives behind a getter or a helper method, you are one innocent reorder away from the same crash.

### The proposal

A new rule `@angular-eslint/inject-at-top` that:

- Applies to classes decorated with `@Component`, `@Directive`, `@Injectable` or `@Pipe`.
- Walks the class body in source order and reports any property initialized with `inject(...)` that appears after another (non-static) member.
- Static members are ignored because they are initialized at class-definition time, not per instance, so their position relative to `inject()` is irrelevant.
- No autofix: the correct position for each `inject()` call depends on the grouping the author wants at the top of the class, and auto-moving them risks shuffling unrelated members or comments. A clear error message is enough.

With the rule in place, the example above fails lint on the `inject(UserService)` line, with a message telling the author to move it to the top of the class, which by definition removes any chance of another member being initialized before it.

### Why it should be in `recommended`

The bug it catches is a true runtime bug, not a convention. The TypeScript compiler cannot see it once it is behind a getter or a method, and there is no reason in idiomatic Angular code to have an `inject()` call sit anywhere other than the top of the class. `prefer-inject` is already in `recommended` and nudges people toward the field-level `inject()` pattern, this rule closes the ordering footgun that comes with it.

### Scope

- In scope: field-level `inject()` calls inside `@Component` / `@Directive` / `@Injectable` / `@Pipe` classes.
- Out of scope (at least for the initial version): `inject()` calls inside functions or factories, non-Angular classes, and detecting which specific consumer would fail, the rule keeps it simple and just enforces position.

AdiMarianMutu:
> [!NOTE]  
> This fixes #3008 

Class fields are initialized in the order they are written, so any field, getter, or method declared above an inject() call cannot safely use the injected service: it will still be undefined at the moment its initializer runs.

TypeScript catches the obvious shape of this bug (TS2729) when an earlier field reads a later field directly, but it does not trace through a getter body or a method call, so any indirection hides the problem from the compiler and the failure only shows up at runtime, with a "Cannot read properties of undefined" error that points at the consumer rather than at the misordered declaration.

This rule flags every inject() call in a class decorated with @Component, @Directive, @Injectable or @Pipe that is declared below any other (non-static) member, forcing all injections to sit at the top of the class and eliminating the ordering hazard.

Enabled in the recommended config.


packages/eslint-plugin/src/rules/inject-at-top.ts
```diff
@@ -0,0 +1,74 @@
...
+export const RULE_NAME = 'inject-at-top';
+
+export default createESLintRule<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Requires inject() calls to be declared at the top of the class, before any other member',
+      recommended: 'recommended',
```
JamesHenry: Please remove and regenerate configs, we do not change recommended configs outside of major versions. Please open an issue if you wish for the rule to be considered for recommended inclusion in the next major

JamesHenry (changes requested):
Thanks for this @AdiMarianMutu, there's a lot of syntactic constructs that this currently does not cover:

```ts
foo1 = cond ? inject(A) : inject(B);                  // ConditionalExpression
foo2 = inject(A, {optional: true}) ?? fallback;       // LogicalExpression
foo3 = inject(A, {optional: true})!;                  // TSNonNullExpression
foo4 = inject(A) as Foo;                              // TSAsExpression
foo5 = wrap(inject(A));                               // wrapper CallExpression
foo6 = new Wrapper(inject(A));                        // NewExpression
foo7 = [inject(A), inject(B)];                        // ArrayExpression
foo8 = {svc: inject(A)};                              // ObjectExpression
foo9 = (sideEffect(), inject(A));                     // SequenceExpression
foo10 = `${inject(A).id}`;                            // TemplateLiteral
foo11 = (() => inject(A))();                          // IIFE — still runs at field init
```

See https://stackblitz.com/edit/4vsytrav?file=src%2Fmain.ts for a variant of the angular.dev playground that compiles and runs with these applied

packages/eslint-plugin/src/rules/inject-at-top.ts
AdiMarianMutu: Hi @JamesHenry!

Can [the main](https://github.com/angular-eslint/angular-eslint/issues/3008) issue be considered for recommended inclusion or I should create a new one?

JamesHenry:
@AdiMarianMutu Thanks for continuing your work on this.

Please rebase on the v22 changes and be sure to add `Service` support in here now that that is supported in v22, then we'll look to get this merged. We can bed this in for a while and potentially consider `recommended` status for v23

AdiMarianMutu:
> @AdiMarianMutu Thanks for continuing your work on this.
> 
> Please rebase on the v22 changes and be sure to add `Service` support in here now that that is supported in v22, then we'll look to get this merged. We can bed this in for a while and potentially consider `recommended` status for v23

Done, let me know if I've missed something.

JamesHenry (approved):

JamesHenry:
Thank you @AdiMarianMutu

Issue #2489, opened 2025-06-02
RFC: Updated i18n rule(s)


JamesHenry:
The i18n functionality of Angular itself is very comprehensive and can be complex, particularly in large codebases where people's conventions, and exceptions to those conventions, need to be accounted for.

We have an existing `i18n` rule in `@angular-eslint/eslint-plugin-template` which is already quite heavy, but there have been a number of requested additions which have not yet been addressed.

Not all of these proposals are complementary (as well as some requiring breaking changes to existing features) and so it is worth stepping back and reimagining our `i18n` support as a whole.

One open question is: Do we want to split `i18n` into multiple rules instead of one all encompassing one?

- Pro: Matches other things conceptually, where we don't try and capture every possible feature related to e.g. component metadata in one rule
- Con: May require duplication of code and effort

One thing that makes this trickier is that I have never personally leveraged the `i18n` functionality in Angular, so I don't have a perfect intuition for the most effective lint feedback.

This issue is open for community discussion and design and I am looking for quality input from experienced users of both the `i18n` feature set in Angular and the existing `i18n` rule and its current shortcomings.

In terms of roll out of changes:

- New rule(s) will be added in v20 minor and the existing one will be deprecated
- Old rule will be removed in v21 and an automated migration to the new one(s) provided wherever possible

Thanks all!

aleesaan:
Thanks for looking into this @JamesHenry! I've been using the i18n rules for a while so I thought to share my opinion.  

I do like that I can just enable it and use its defaults for most things, which are very useful. I only override these:

- `checkId: false`: I think this could actually be the default - users could set it to `true` if they want to have full control on the IDs, otherwise they could let angular handles the IDs. In fact the suggestion [in their docs](https://angular.dev/guide/i18n/manage-marked-text) is to add custom IDs only when needed (we actually also need to update the links to the docs [in the rule](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/rules/i18n.ts#L117) as they don't work). But I'm curious to see how the community uses it / if most people actually do provide custom IDs!
- `ignoreAttributes`: this is one I think most people do need to customize based on their apps, the libraries they use etc.; but it's great that we have built-in defaults there too!

So my vote would be to keep them all in one rule as it's quite nice DX to just enable it and get all the defaults automatically.  

[...]

PR #3012, opened 2026-04-24
feat(eslint-plugin-template): [i18n] add includeAttributes option

### LINKED ISSUE #929: [i18n] new option includeAttributes
I would like to propose a new option _includeAttributes_ to the i18n rule.
This option should be a complement to the _ignoreAttributes_ option, and specify attributes that should be checked even if the _checkAttributes_ option is false.

This way when using 3rd party components that have lots of string attributes of which only a handful are texts that should be translated, we could list only the relevant attributes instead of all the irrelevant ones.

I would also like to propose that the enhancement requested in #839 is included in this option as well, for the same reasons.

```JSON
{
  "rules": {
    "@angular-eslint/template/i18n": ["error", {
      "checkAttributes": false,
      "includeAttributes": ["title","caption","name"]
    }]
  }
}
```

```HTML
// your repro code case
<dx-tree-list
    id="employees"
    [dataSource]="employees"
    keyExpr="ID"
    parentIdExpr="Head_ID"
    [showRowLines]="true"
    [showBorders]="true"
    [columnAutoWidth]="true"
    [expandedRowKeys]="[1, 2, 3, 4, 5]"
    (onEditorPreparing)="editorPreparing($event)"
    (onInitNewRow)="initNewRow($event)"
  >
    <dxo-editing
      mode="row"
      [allowUpdating]="true"
      [allowDeleting]="true"
      [allowAdding]="true"
    >
    </dxo-editing>

    <dxi-column dataField="Full_Name">
      <dxi-validation-rule type="required"></dxi-validation-rule>
    </dxi-column>
    <dxi-column dataField="Head_ID" caption="Head" i18n-caption="@@columnCaptionHead">
      <dxo-lookup
        [dataSource]="lookupData"
        valueExpr="ID"
        displayExpr="Full_Name"
      >
      </dxo-lookup>
      <dxi-validation-rule type="required"></dxi-validation-rule>
    </dxi-column>
    <dxi-column dataField="Title" caption="Position" i18n-caption="@@columnCaptionTitle">
      <dxi-validation-rule type="required"></dxi-validation-rule>
    </dxi-column>
    <dxi-column dataField="Hire_Date" dataType="date" [width]="120">
      <dxi-validation-rule type="required"></dxi-validation-rule>
    </dxi-column>
    <dxi-column type="buttons">
      <dxi-button name="edit" i18n-name="@@buttonTextEdit"></dxi-button>
      <dxi-button name="delete" i18n-name="@@buttonTextDelete"></dxi-button>
    </dxi-column>
  </dx-tree-list>
  <dx-accordion
    [collapsible]="true"
    [multiple]="false"
    [animationDuration]="300">
    <dxi-item i18n-title="Accordion title@@options" title="Options">
      <my-content-component></my-content-component>
    </dxi-item>
  </dx-accordion>
```

pierluigilenoci:
## Summary

Adds a new `includeAttributes` option to the `i18n` rule, addressing the long-standing request from [#929](https://github.com/angular-eslint/angular-eslint/issues/929) and contributing to the RFC discussion in [#2489](https://github.com/angular-eslint/angular-eslint/issues/2489).

### Problem

When using third-party component libraries (e.g., DevExtreme, Angular Material), components often have many string attributes, but only a handful are user-facing text that needs translation. The current approach requires either:
1. Setting `checkAttributes: true` and listing **all** non-translatable attributes in `ignoreAttributes` (tedious, fragile)
2. Setting `checkAttributes: false` and losing all attribute checking (too permissive)

### Solution

The new `includeAttributes` option provides a third approach: set `checkAttributes: false` to disable blanket attribute checking, then use `includeAttributes` to explicitly list only the attributes that **must** have i18n markers.

```json
{
  "@angular-eslint/template/i18n": ["error", {
    "checkAttributes": false,
    "includeAttributes": ["title", "caption", "placeholder", "dxi-column[caption]"]
  }]
}
```

### Behavior

- `includeAttributes` accepts plain attribute names (`"caption"`) and element-scoped names (`"dxi-column[caption]"`), matching the existing `ignoreAttributes` syntax
- When an attribute is in `includeAttributes`, it is always checked — even if `checkAttributes` is `false`
- `includeAttributes` takes precedence over `ignoreAttributes` — if an attribute appears in both, it **will** be checked
- The existing autofix (inserting `i18n-<attr>`) works for included attributes

### Test Plan

- [x] Valid case: included attribute with `i18n-*` present does not report
- [x] Valid case: non-included attributes are not checked when `checkAttributes: false`
- [x] Valid case: element-scoped `includeAttributes` only matches the specified element
- [x] Invalid case: included attribute without `i18n-*` reports even with `checkAttributes: false`
- [x] Invalid case: element-scoped included attribute reports correctly
- [x] Invalid case: `includeAttributes` overrides `ignoreAttributes`
- [x] Invalid case: multiple included attributes report independently with autofix
- [x] TypeScript compiles cleanly (`tsc --noEmit` on both lib and spec configs)
- [x] Prettier formatting verified

Closes #929
Ref #2489

---

*I generated this PR description with the help of AI tools but have reviewed and validated the implementation.*


pierluigilenoci:
Hi — friendly follow-up. CI is green and all checks pass. Would you be able to review when you get a chance? Thank you!

JamesHenry:
I'm sorry that it's stalled a bit, but the RFC consolidated things intentionally, so we wouldn't proceed with a small piece of it. If you are available to contribute to the future of i18n linting then please continue on that thread

PR #3050, opened 2026-06-02
fix(eslint-plugin): [prefer-signal-model] skip inputs with transform option

### LINKED ISSUE #2884: [@angular-eslint/prefer-signal-model] Inputs with transform

The rule updates inputs that are using the `transform` option to `model`. As `model` does not have a `transform` option, we end up with broken code. InputSignalWithTransform should therefore be ignored.

https://stackblitz.com/edit/angular-eslint-repros-317evvwt?file=src%2Fmain.ts

**Versions**

Angular CLI       : 21.1.2
Angular           : 21.1.2
Node.js           : 24.5.0
Package Manager   : pnpm 10.18.3
Operating System  : darwin arm64

| Package  | Installed Version | Requested Version |
|--------|--------|--------|
| @angular/animations | 21.1.2 | ^21.1.2 |
| @angular/build                    | 21.1.2            | ^21.1.2           |
| @angular/cdk                      | 21.1.2            | ~21.1.2           |
| @angular/cli                      | 21.1.2            | ~21.1.2           |
| @angular/common                   | 21.1.2            | ^21.1.2           |
| @angular/compiler                 | 21.1.2            | ^21.1.2           |
| @angular/compiler-cli             | 21.1.2            | ^21.1.2           |
| @angular/core                     | 21.1.2            | ^21.1.2           |
| @angular/forms                    | 21.1.2            | ^21.1.2           |
| @angular/material                 | 21.1.2            | ~21.1.2           |
| @angular/platform-browser         | 21.1.2            | ^21.1.2           |
| @angular/platform-browser-dynamic | 21.1.2            | ^21.1.2           |
| @angular/router                   | 21.1.2            | ^21.1.2           |
| @angular/service-worker           | 21.1.2            | ^21.1.2           |
| ng-packagr                        | 21.1.0            | ^21.1.0           |
| rxjs                              | 7.8.2             | ~7.8.2            |
| typescript                        | 5.9.3             | ~5.9.3            |

xonaib:
Fixes #2884

The `prefer-signal-model` rule was converting `input()` calls that include a `transform`
option to `model()`, producing broken code because `model()` does not support the
`transform` option.

Root cause: the rule collected all `input()` calls without checking for a `transform`
option in the arguments.

Fix: added an early return in the `input` handler when the call expression contains an
`ObjectExpression` argument with a `transform` property. These inputs are now silently
skipped by the rule.

Also added two valid test cases covering `booleanAttribute` and `numberAttribute` transforms.

**Why `model()` won't support `transform`:**
This is [by design](https://github.com/angular/angular/issues/55166#issuecomment-2032150999) — `model()` is built for bi-directional synchronization. Supporting `transform` during the data→model phase would cause the internal value to diverge from the external value, potentially leading to `ExpressionChanged` errors or infinite change-detection loops. The Angular team closed the feature request as won't-fix.

Verified locally:
- `.\node_modules\.bin\nx.cmd test eslint-plugin tests/rules/prefer-signal-model/spec.ts`


JamesHenry:
Thanks @xonaib please could you rebase your PR on the latest, I just shipped v22, and please also link to the authoritative comment which explains why `model()` won't support transform:

https://github.com/angular/angular/issues/55166#issuecomment-2032150999

JamesHenry (changes requested):
See comment

xonaib:
hey @JamesHenry, I rebased on main and updated the description with the link to the Angular team's comment. Thanks for the feedback

JamesHenry:
Sorry I should have been clearer - please add the URL of the GitHub comment to the code comment in the implementation so that anyone finding it in the future knows why we did it

xonaib:
Added the URL to the code comment as well. Thanks for clarifying!

AI-JamesHenry (changes requested):
Thanks @xonaib, the fix is correct in concept and skipping these inputs entirely (rather than reporting without an autofix) is the right design choice: per the linked Angular issue there is no valid conversion for a transform input, so there is nothing actionable to report. Cross-checking against the angular.dev API docs, `transform` is precisely the divergence between the two option types (`InputOptions` is `{ alias?, transform?, debugName? }`, `ModelOptions` is `{ alias?, debugName? }`), so guarding on `transform` alone covers exactly what `model()` cannot express, and `alias`/`debugName` inputs remain convertible.

I applied the patch locally against latest `main` and verified the behavior, including that a class mixing a transform input with a plain `input()` + `xChange` pair still reports and fixes the plain one correctly (converts it to `model()`, removes the `Change` output, adds the `model` import, and leaves the transform pair untouched).

Two changes needed before this can land:

### 1. Regenerate the rule docs (currently fails CI)

The rule docs embed the test cases, so adding the two valid cases requires regenerating `docs/rules/prefer-signal-model.md`. I reproduced the `check-rule-docs` failure locally with this PR applied. Please run:

```bash
pnpm update-rule-docs
```

and commit the resulting changes (about 60 added lines in the doc).

### 2. Check `node.arguments[1]` instead of scanning every argument

All documented overloads of `input()` take the initial value first and the options object second; there is no one-argument options form. Scanning every argument means an object *initial value* that happens to contain a `transform` key falsely suppresses the rule. For example, this is skipped even though it converts to `model()` perfectly:

```ts
readonly style = input({ transform: 'scale(2)' }); // CSS transform config as initial value
readonly styleChange = output<object>();
```

That is a plausible shape in real code (style/config objects), so please restrict the guard to `node.arguments[1]`. It makes the check both simpler and precise.

### Test suggestions

- An invalid case for the mixed-class scenario above, since it is the case most likely to regress if the guard placement ever changes.
- Once the guard is restricted to `arguments[1]`, an invalid case locking in that an object initial value containing a `transform` key is still converted.

### Notes, no action needed

- Options passed indirectly (`input(0, SHARED_OPTS)` or `input(0, { ...base })`) with a transform inside will still be converted and break. That is undetectable without type information and fine to live with, though it could be worth a line in the rule docs.
- Handling the string-literal key form `{ 'transform': ... }` was good attention to detail.
- House style for single-rule commit/PR titles is `fix(eslint-plugin): [prefer-signal-model] skip inputs with transform option`. Cosmetic, can be handled at squash time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JamesHenry (approved):
Re-reviewed this from scratch on the latest head (`1498a6a`). Thanks @xonaib, this is in good shape and both items from the previous round are addressed.

### What I verified

- **Design:** silently skipping `input()` calls that carry a `transform` option is the right call. `transform` is the one thing `InputOptions` has that `ModelOptions` does not (the linked Angular comment confirms it will stay that way), so there is no valid conversion and nothing actionable to report. `alias` / `debugName` inputs remain convertible.
- **Guard placement:** restricting the check to `node.arguments[1]` is correct. Every `input()` overload takes the initial value first and options second, so an object *initial value* containing a `transform` key is no longer a false negative. The new invalid case locks that in.
- **Mixed class:** the new invalid case covers a transform pair sitting next to a plain `input()` + `xChange` pair, and the plain one is still reported and fixed while the transform pair is left alone.
- **Docs:** `prefer-signal-model.md` is regenerated and `check-rule-docs` passes.

Ran locally on the PR head, and again after merging the branch onto current `main` (48 commits ahead, merges cleanly):

- `nx test eslint-plugin tests/rules/prefer-signal-model/spec.ts` — 15/15 passing
- `nx run eslint-plugin:check-rule-docs` — passes
- `pnpm format-check`, `nx run-many -t lint typecheck -p eslint-plugin` — pass

I also probed a few option shapes against the built plugin: `{ 'transform': fn }`, `{ ['transform']: fn }`, shorthand `{ transform }`, and `input<T, U>(undefined, { transform })` are all skipped; `{ alias: 'x' }` and `input({ transform: 'scale(2)' })` are still reported. All as expected.

### Non-blocking nits

- The comment on the second valid case says "custom transform function" but the case uses the built-in `numberAttribute`. Either adjust the comment or use an actual custom function (e.g. `(value: string) => value.trim()`), which would also broaden the coverage slightly.
- Optional: a sentence in `RULE_DOCS_EXTENSION` noting that inputs with a `transform` are intentionally skipped, and that options passed indirectly (`input(0, SHARED_OPTS)` or `{ ...base }`) cannot be detected without type information. Fine to leave for a follow-up.

### Notes for the maintainer, no action for the author

- The main CI workflow has not reported on the current head; only "Validate PR title" ran on `1498a6a`. The last Nx Cloud pipeline (on `cb7d53a`) failed only in the two e2e tasks because the Verdaccio uplink returned HTTP 500, which is unrelated to this change. A fresh CI run on the head should be green based on the local results above.
- `input.required(...)` takes options as its first argument, but this rule's selector only matches bare `input(...)` calls, so `input.required` never reaches this code path. Pre-existing and out of scope here.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3054, opened 2026-06-05
flat config file detection relies on file names


chrisguttandin:
**Description and reproduction of the issue**

As far as I can tell wether or not a flat config file is used is currently determined by checking the name of the config file.

https://github.com/angular-eslint/angular-eslint/blob/6b2628a67442014969350e6be0a13c0f147a5665/packages/builder/src/lint.impl.ts#L56-L63

https://github.com/angular-eslint/angular-eslint/blob/6b2628a67442014969350e6be0a13c0f147a5665/packages/builder/src/utils/eslint-utils.ts#L78-L86

https://github.com/angular-eslint/angular-eslint/blob/6b2628a67442014969350e6be0a13c0f147a5665/packages/builder/src/utils/eslint-utils.ts#L4-L10

This means when using flat config files it's currently required to have a file named `eslint.config.*` in the root directory. The actual config file can have any prefix but also needs to end with `eslint.config.*`.

It looks like ESLint itself doesn't require a specific file name to be used when specifying the config file manually.

I was wondering if there is a better way to detect flat config files? Maybe it's possible to get the version of the currently used eslint package from within `angular-eslint`. If possible it could be used to do something like this:

```ts
const useFlatConfig === (esLintVersion === 8 && env.ESLINT_USE_FLAT_CONFIG === 'true')
    || (esLintVersion === 9 && env.ESLINT_USE_FLAT_CONFIG !== 'false')
    || esLintVersion === 10
```

Another option might be to add new property to configure flat config file usage in `angular.json`.

I guess the simplest option is to just wait until dropping support for ESLint v8 and v9. :-)
 

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@angular-eslint/builder`          | `21.4.0` |
| `ESLint`                           |  `9.39.4` |
| `node`                             | `22.22.2` |

```sh
# Please run `npx ng version` in your project and paste the full output here:

Angular CLI       : 21.2.13
Angular           : 21.2.15
Node.js           : 22.22.2
Package Manager   : npm 11.16.0
Operating System  : darwin arm64

┌───────────────────────────────────┬───────────────────┬───────────────────┐
│ Package                           │ Installed Version │ Requested Version │
├───────────────────────────────────┼───────────────────┼───────────────────┤
│ @angular/animations               │ 21.2.15           │ ^21.2.15          │
│ @angular/build                    │ 21.2.13           │ ^21.2.13          │
[...]

Issue #3020, opened 2026-05-02
Add support for Angular `@Service` decorator (only Angular 22+)


splincode:
## Description

Angular is introducing a new `@Service()` decorator for dependency injection services.

It would be useful for `angular-eslint` to support this decorator as a first-class Angular decorator and optionally provide a rule that prefers `@Service()` over `@Injectable({providedIn: 'root'})` when the migration is safe.

Related Angular PR: https://github.com/angular/angular/pull/68506

## Motivation

Today, root-provided Angular services are commonly declared with `@Injectable({providedIn: 'root'})`:

```ts
import {Injectable} from '@angular/core';

@Injectable({providedIn: 'root'})
export class UserService {}
```

With the new `@Service()` decorator, the same root-provided service can be expressed more directly:

```ts
import {Service} from '@angular/core';

@Service()
export class UserService {}
```

Angular itself has started using `@Service()` in places where `@Injectable({providedIn: 'root'})` was previously used.

For factory-based services, the Angular PR also shows this migration pattern:

```ts
@Injectable({
  providedIn: 'root',
  useFactory: () => new NgLocaleLocalization(inject(LOCALE_ID)),
})
```

to:

```ts
@Service({
  factory: () => new NgLocaleLocalization(inject(LOCALE_ID)),
})
```

## Proposed rule

Add a new rule that reports `@Injectable({providedIn: 'root'})` when it can be safely replaced with `@Service()`.

Possible rule names:

- `prefer-service-decorator`
- `prefer-angular-service`
- `prefer-service-over-injectable`

I personally prefer `prefer-service-decorator`, because it is short and directly describes the desired Angular decorator.

## Invalid examples

```ts
import {Injectable} from '@angular/core';

@Injectable({providedIn: 'root'})
export class UserService {}
```

```ts
import {Injectable} from '@angular/core';

@Injectable({
  providedIn: 'root',
})
export class UserService {}
```

```ts
import {Injectable, inject} from '@angular/core';

@Injectable({
  providedIn: 'root',
  useFactory: () => inject(UserServiceImpl),
})
export abstract class UserService {}
```

## Valid examples

```ts
import {Service} from '@angular/core';

@Service()
export class UserService {}
```

```ts
import {Service, inject} from '@angular/core';

@Service({
  factory: () => inject(UserServiceImpl),
})
export abstract class UserService {}
```

```ts
import {Injectable} from '@angular/core';

@Injectable()
export class LocallyProvidedService {}
```

[...]

Issue #502, opened 2021-05-24
[no-conflicting-lifecycle] false positives


rafaelss95:
**Description and reproduction of the issue**
As mentioned in https://github.com/mgechev/codelyzer/issues/560#issuecomment-621313192 and https://github.com/mgechev/codelyzer/issues/588:

> no-conflicting-life-cycle-hooks should prevents this:

> Note that a directive typically should not use both DoCheck and OnChanges to respond to changes **on the same input**, as ngOnChanges will continue to be called when the default change detector detects changes.

> Currently, it doesn't seem to be the case. We just check if we use the both hooks in same time. 

This rule should report **only** if a variable is being checked in both `ngDoCheck` and` ngOnChanges` methods, not as it currently does that reports if you implement both `DoCheck` and `OnChanges` or have corresponding methods.

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@angular-eslint       `        | `12.0.0` |
| `TypeScript`                       | `4.3.0` |
| `ESLint`                           | `7.27.0` |

```sh
# Please run `npx ng version` in your project and paste the full output here:
Angular CLI: 12.0.0
Node: 14.16.0
Package Manager: npm 6.14.11
OS: darwin x64

Angular: 12.0.0
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, localize, material
... platform-browser, platform-browser-dynamic, platform-server
... router, service-worker

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1200.0
@angular-devkit/build-angular   12.0.0
@angular-devkit/core            11.2.11
@angular-devkit/schematics      11.2.11
@nguniversal/builders           12.0.0
@nguniversal/common             12.0.0
@nguniversal/express-engine     12.0.0
@schematics/angular             12.0.0
ng-packagr                      12.0.0
rxjs                            6.5.5
typescript                      4.2.4

```

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest supported version of the packages and checked my `ng version` output per the instructions given here.

rafaelss95:
Being quite honest: this is a very difficult (not to say impossible) rule to implement correctly, due to the many possibilities.

What actually defines whether an input is being checked in both methods at all? Should the code below trigger the rule? If so, why?

```ts
@Component()
class ChildComponent implements DoCheck {
  @Input() person!: Person;
  readonly differ = this.keyValueDiffers.find({}).create();
[...]

PR #3055, opened 2026-06-06
feat!: angular-eslint v22

### LINKED ISSUE #3052: Support for Angular v22
Angular 22 has [released today](github.com/angular/angular/releases/tag/v22.0.0) - when can we expect a new version of angular-eslint with support?

Thanks!

JamesHenry:
- [x] Angular 22
- [x] TypeScript 6
- [x] Switch to supporting only node: `^22.22.3 || ^24.15.0 || ^26.0.0` to match Angular 22
- [x] Drop support for ESLint v8
- [x] Drop support for legacy eslintrc config format
- [x] Support `@Service` across existing rules
- [x] Add coverage for `//` comment format and update doc references
  - ⚠️ NOTE: `eslint-disable` comments will NOT work in these because the Angular compiler does not take them into account as part of `collectCommentNodes` right now. I have reported it to the Angular Team privately.
- [x] Promote `@angular-eslint/prefer-on-push-component-change-detection` to recommended and handle changed default Angular behaviour
- [x] Update AST traversal for new template syntax for rest/spread and arrow functions
- [x] Remove deprecated `no-conflicting-lifecycle`
- [x] Remove restriction on ESLint config names from builder

Fixes https://github.com/angular-eslint/angular-eslint/issues/3052 https://github.com/angular-eslint/angular-eslint/issues/3054 https://github.com/angular-eslint/angular-eslint/issues/3020 https://github.com/angular-eslint/angular-eslint/issues/502


medbenmakhlouf:
@JamesHenry , will you also include the suppression of this rule: '**@angular-eslint/prefer-on-push-component-change-detection'**

Since ChangeDetectionStrategy.OnPush becomes the default in v22, this rule may no longer be necessary. Perhaps instead of suppressing it, the rule could be updated to recognize and respect the framework default. I may be missing something, though. Thanks!

JamesHenry:
@medbenmakhlouf I'm not following you - the rule has just as much merit now as it did before. Users can still choose not to use OnPush and the rule will inform them that they should prefer to use it. I will tweak the rule's docs entry but this has no impact on the rule's behaviour. I will also add it to the recommended config at this point.

UPDATE: Sorry I think I now see part of what you mean - you mean the fact that the rule previously would require explicit annotation, whereas now because it is the default, it can be left off, I will update the rule to take the scenarios into account

medbenmakhlouf:
@JamesHenry yes exactly

PR #3067, opened 2026-06-08
no-implicit-take-until-destroyed introduced false-negatives

PowerKiKi:
@pierluigilenoci, @JamesHenry, the very recently merged PR #3011 fixed the false-positive from no-implicit-take-until-destroyed, that [I originally reported](https://github.com/angular-eslint/angular-eslint/pull/3011). Unfortunately it also introduces false-negatives when a `takeUntilDestroyed()` is part of a method that is called _both_ by the constructor and another (public) method.

This PR adds a failing test cases that prove the introduction of false-negatives.

I would argue that false-negative is worse than false-positive, because false-positives might be workaround-ed, but not false-negatives, since we are not even aware of them. IMHO the PR #3011 should either be entirely reverted, or thoroughly completed.

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-implicit-take-until-destroyed/cases.ts
+++ b/packages/eslint-plugin/tests/rules/no-implicit-take-until-destroyed/cases.ts
@@ -324,4 +324,26 @@ export const invalid: readonly InvalidTestCase<MessageIds, Options>[] = [
     `,
     messageId,
   }),
+  convertAnnotatedSourceToFailureCase({
+    description:
+      'should fail in method called both from constructor and from not constructor',
+    annotatedSource: `
+      @Component()
+      class Test {
+        constructor() {
+          this.loadData();
+        }
+
+        ngOnInit() {
+          this.loadData();
+        }
+
+        loadData() {
+          this.data$.pipe(takeUntilDestroyed()).subscribe();
+                          ~~~~~~~~~~~~~~~~~~~~
+        }
+      }
+    `,
+    messageId,
+  }),
 ];


JamesHenry:
@pierluigilenoci I'll give you a chance to respond

pierluigilenoci:
@PowerKiKi thanks for the quick follow-up. Upfront caveat: my expertise on this rule isn't as deep as yours or @JamesHenry's, so take what follows as my reading of the code rather than a definitive answer — happy to be corrected. From what I can see, you're right: this is a real false-negative and worth fixing.

To make sure I understand the case correctly:

```ts
@Component()
class Test {
  constructor() { this.loadData(); }   // injection context ✅
  ngOnInit()    { this.loadData(); }   // NO injection context ❌
  loadData()    {
    this.data$.pipe(takeUntilDestroyed()).subscribe();
  }
}
```

At runtime the `ngOnInit` call site will throw `NG0203` because Angular invokes lifecycle hooks after construction, outside the injection context. The current heuristic in `isMethodCalledFromInjectionContext` returns `true` as soon as it finds *any* injection-context caller — it doesn't check whether the method is *also* invoked from a non-injection caller.

I'd push back gently on a full revert though: that would re-introduce the false-positives from #2929, which are arguably the more painful failure mode in practice (they fire on every clean refactor where the constructor delegates to a private helper, with no actionable workaround short of disabling the rule per-line).

If I'm reading the rule correctly, the fix is to make the suppression bidirectional: only suppress when the method is reachable from injection contexts **and never from a non-injection caller**. Rough sketch:

```ts
let calledFromInjection = false;
let calledFromNonInjection = false;

for (const member of classDeclaration.body.body) {
  if (member === method) continue;
  if (!containsCallToMethod(member, methodName, isPrivateField)) continue;

  if (isConstructorMethod(member) || ASTUtils.isPropertyDefinition(member)) {
    calledFromInjection = true;
  } else {
    calledFromNonInjection = true;
  }
}

return calledFromInjection && !calledFromNonInjection;
```

That would make your fail-on-main case correctly flagged, keep every valid case from #3011 passing, and close the gap symmetrically — but I'd appreciate a sanity check from either of you before going further, in case I'm missing a wrinkle (transitive calls through a third method, methods called from both a constructor and another helper that's only called from the constructor, etc. — the simple two-flag approach above doesn't recurse).

@JamesHenry happy to open a follow-up PR with the fix plus the test cases from this PR (constructor + non-constructor, lifecycle-only, plus a recursive-self-call edge case), if that direction works for you. PowerKiKi, would you prefer to extend this PR with the fix yourself, or shall I take it from here and credit you?

(Side note: the PR title is failing the semantic-PR-title check — needs a conventional-commits prefix, something like `test(eslint-plugin): [no-implicit-take-until-destroyed] add failing case for false-negative`, depending on which direction we go.)

PowerKiKi:
@pierluigilenoci I think your approach is right. It should get rid of false-negative.

One more delicate cases is if it is called by a public, non-constructor, method and it happens that that external caller is in injection context (eg: `ServiceB.constructor()` calls `ServiceA.loadData()`). I suppose those cases would still be false-positives, even after the correction you suggest. And that is entirely OK for me. As long as we don't have false-negatives.

pierluigilenoci:
@PowerKiKi good point on the cross-class case — and thanks for stating the trade-off so clearly, it really helps frame the response.

My honest take: I don't think we can resolve that last case **robustly** without paying a significant cost, and I'm not sure the cost is worth it. The fundamental issue is that ESLint, in single-file mode, sees `ServiceA.loadData()` in isolation — it has no visibility into whether a `ServiceB.constructor` somewhere else in the codebase is the only caller. Any heuristic based on method name, visibility, or argument shape would be fragile and likely re-introduce a different class of false-positives.

There are, as far as I can see, three paths — with increasing cost:

1. **Accept the cross-class false-positive** (the position you yourself suggested). The rule guarantees zero false-negatives, and the user reaches for `// eslint-disable-next-line` on the rare cross-class case — which is arguably the legitimate use of the escape hatch. Lowest implementation cost, lowest maintenance burden.

2. **Explicit opt-in marker.** A JSDoc tag like `/** @injectionContext */` on the method, or a rule option (e.g. `allowedMethodNames`), would let the user *declare* "trust me, this is always called in injection context." Cheap to implement (one check before the bidirectional one) and keeps the responsibility explicit and reviewable. A reasonable middle ground if the `eslint-disable-next-line` ergonomics turn out to be painful in practice.

3. **Type-aware cross-file analysis.** Use `parserServices` + the TypeScript checker to resolve the method symbol, collect all references project-wide, and walk each call site upward to classify the caller (constructor / property initializer / lifecycle hook / public method). Suppress only if **every** reference is in an injection context. Technically feasible, but: (a) it would require flipping the rule to `requiresTypeChecking: true`, which is a category change with real perf implications on large projects; (b) it still wouldn't help when the class is part of a published library consumed by external apps the checker can't see; (c) it adds a non-trivial surface area of bugs for a case you've explicitly classified as acceptable.

If I had to pick, I'd lean toward **option 2**: it gives users an explicit, reviewable way to opt out of the warning when they know the call is safe, without the perf and scope costs of option 3, and without leaving them stuck with a per-line `eslint-disable` as the only workaround. Implementation would be a small addition on top of the bidirectional fix — happy to include it in the same follow-up PR if you both agree it's the right direction.

That said, I'd genuinely value a sanity check from either of you before committing to a path — especially @JamesHenry, in case there's a precedent in other angular-eslint rules I should align with (or avoid).

PowerKiKi:
Am I talking to a LLM ?

Option 1 have my preference, by far. Keep analysis cost low, do not introduce new concept `/** @injectionContext */`

JamesHenry:
@PowerKiKi I think you probably were 😅 that user has disappeared.

Sadly this means we have no record of the Option 1 you were referencing here... would you mind updating the thread with the latest please?

JamesHenry (changes requested):
Just marking it as not ready

PowerKiKi:
Copy-pasting the (LLM generated ?) comment from deleted user @pierluigilenoci:

> [@PowerKiKi](https://github.com/PowerKiKi) good point on the cross-class case — and thanks for stating the trade-off so clearly, it really helps frame the response.
> 
> My honest take: I don't think we can resolve that last case robustly without paying a significant cost, and I'm not sure the cost is worth it. The fundamental issue is that ESLint, in single-file mode, sees ServiceA.loadData() in isolation — it has no visibility into whether a ServiceB.constructor somewhere else in the codebase is the only caller. Any heuristic based on method name, visibility, or argument shape would be fragile and likely re-introduce a different class of false-positives.
> 
> There are, as far as I can see, three paths — with increasing cost:
> 
> 1. **Accept the cross-class false-positive** (the position you yourself suggested). The rule guarantees zero false-negatives, and the user reaches for // eslint-disable-next-line on the rare cross-class case — which is arguably the legitimate use of the escape hatch. Lowest implementation cost, lowest maintenance burden.
> 
> 2. **Explicit opt-in marker.** A JSDoc tag like /** @injectionContext */ on the method, or a rule option (e.g. allowedMethodNames), would let the user declare "trust me, this is always called in injection context." Cheap to implement (one check before the bidirectional one) and keeps the responsibility explicit and reviewable. A reasonable middle ground if the eslint-disable-next-line ergonomics turn out to be painful in practice.
> 
> 3. **Type-aware cross-file analysis**. Use parserServices + the TypeScript checker to resolve the method symbol, collect all references project-wide, and walk each call site upward to classify the caller (constructor / property initializer / lifecycle hook / public method). Suppress only if every reference is in an injection context. Technically feasible, but: (a) it would require flipping the rule to requiresTypeChecking: true, which is a category change with real perf implications on large projects; (b) it still wouldn't help when the class is part of a published library consumed by external apps the checker can't see; (c) it adds a non-trivial surface area of bugs for a case you've explicitly classified as acceptable.
> 
> If I had to pick, I'd lean toward **option 2**: it gives users an explicit, reviewable way to opt out of the warning when they know the call is safe, without the perf and scope costs of option 3, and without leaving them stuck with a per-line eslint-disable as the only workaround. Implementation would be a small addition on top of the bidirectional fix — happy to include it in the same follow-up PR if you both agree it's the right direction.
> 
> That said, I'd genuinely value a sanity check from either of you before committing to a path — especially [@JamesHenry](https://github.com/JamesHenry), in case there's a precedent in other angular-eslint rules I should align with (or avoid).

I'm still in favor of option 1. Because option 2 feels like duplicating the already existing `eslint-disable-next-line`, and option 3 seems unreasonably complex to implement and maintain.

JamesHenry:
Resolved via https://github.com/angular-eslint/angular-eslint/pull/3164 thanks folks

PR #3073, opened 2026-06-10
feat(eslint-plugin): add `prefer-service-decorator` rule

g-drouard:
## Overview

Adds a new fixable rule, `@angular-eslint/prefer-service-decorator`, that migrates `@Injectable({ providedIn: 'root' })` to the new Angular 22 [`@Service()`](https://angular.dev/api/core/Service) decorator — a concise shorthand for the most common service declaration.

The rule was proposed [here](https://github.com/angular-eslint/angular-eslint/issues/3020)

## What the rule does

- Reports and auto-fixes only when the metadata is exactly `providedIn: 'root'` **optionally** combined with `useFactory`.
- Drops `providedIn: 'root'` (implied by `@Service()`) and renames `useFactory` → `factory`.
- **Migrates the import**: `Injectable` → `Service` from `@angular/core`.
- **Leaves untouched** (no report, no fix): `@Injectable()` with no metadata, other `providedIn` values (`null`, `'platform'`, `'any'`, a module reference), any unsupported provider metadata (`useClass`, `useExisting`, `useValue`, `deps`).

### Examples

```ts
// ❌ before
import { Injectable } from '@angular/core';

@Injectable({ providedIn: 'root' })
class MyService {}

// ✅ after
import { Service } from '@angular/core';

@Service()
class MyService {}
```

```ts
// factory is preserved
@Injectable({ providedIn: 'root', useFactory: () => new MyService() })
// →
@Service({ factory: () => new MyService() })
```

## Shared `rule-fixes` improvements

While building the auto-fixer, two shared helpers in `@angular-eslint/utils` were improved/added (benefiting other rules):

- **`getNodeToCommaRemoveFix`** — now removes a comma-separated element cleanly (no leftover blank lines or double spaces, handles first/middle/last and trailing-comma cases) while **keeping the preceding comma** so an existing trailing-comma style is preserved. This makes the existing autofix output of `no-pipe-impure`, `prefer-on-push-component-change-detection`, and `use-component-view-encapsulation` cleaner (their test expectations were updated accordingly).
- **`getImportReplaceFix`** — new helper that migrates a named import `from` → `to`: renames the specifier in place when possible, otherwise composes the existing `getImportAddFix` + `getImportRemoveFix` (the rename-in-place avoids an overlapping-fix conflict when the old symbol is the sole import).


JamesHenry (approved):
Thank you!

PR #3077, opened 2026-06-11
chore: add schematics package bugs metadata

mawxcodehub:
## Summary
- add `bugs.url` metadata for the published `@angular-eslint/schematics` package

## Validation
- `node -e "const p=require('./packages/schematics/package.json'); if(p.name!=='@angular-eslint/schematics'||p.bugs.url!=='https://github.com/angular-eslint/angular-eslint/issues') process.exit(1); console.log(p.name, p.version, p.bugs.url)"`
- `git diff --check`

### DIFF
--- a/packages/schematics/package.json
+++ b/packages/schematics/package.json
@@ -10,6 +10,9 @@
     "url": "https://github.com/angular-eslint/angular-eslint.git",
     "directory": "packages/schematics"
   },
+  "bugs": {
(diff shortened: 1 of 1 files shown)


JamesHenry:
Addressed holistically in https://github.com/angular-eslint/angular-eslint/pull/3108

I do find it suspicious that multiple of these piecemeal PRs were opened very close together.

PR #3080, opened 2026-06-13
fix: add missing metadata fields to package.json

BWM0223:
Adds missing `bugs` metadata to `packages/schematics/package.json`.

Fixes npm tooling unable to resolve package metadata.


JamesHenry:
Addressed holistically in https://github.com/angular-eslint/angular-eslint/pull/3108

I do find it suspicious that multiple of these piecemeal PRs were opened very close together.

PR #3089, opened 2026-06-23
fix(eslint-plugin): [prefer-signal-model] don't report when input and output types differ

g-drouard:
## Overview

`prefer-signal-model` flags an `input`/`output` pair (e.g. `value` + `valueChange`) and suggests merging them into a single `model()`. But `model<T>()` exposes **one** type for both directions, so the merge is only valid when the input and output share the same type. Previously the rule reported (and auto-fixed) regardless of the types, which could produce incorrect code.

## Behaviour

By default the rule stays purely syntactic and compares the written type arguments as text:

- ✅ Still reports when the type arguments match — `` input<number>() `` + `` output<number>() ``
- ✅ Still reports when either side has no type argument — `` input() `` + `` output() ``
- ❌ No longer reports when the type arguments differ — `` input<string>() `` + `` output<number>() ``, or `` input<string | null>() `` + `` output<string>() ``

### New `useTypeChecking` option (default `false`)

Following the `prefer-signals` house pattern, an opt-in `useTypeChecking` option compares the two types semantically through the TypeScript type checker (mutual assignability, since `isTypeIdenticalTo` is internal API):

- `` input<string | null>() `` + `` output<null | string>() `` → reported (union order does not matter)
- `` input<{ id: string }>() `` + `` output<{ id: string }>() `` → reported (structurally identical types written at two locations)
- the input's type is inferred from its initial value, so `` input('') `` + `` output<number>() `` is skipped, while `` input('') `` + `` output<string>() `` is still migrated
- a type argument that does not resolve (typo, missing declarations, stale import) becomes the error type, which is assignable in both directions — so those pairs fall back to the text comparison instead, and `` input<Foo>() `` + `` output<Bar>() `` is skipped in this mode too

`ESLintUtils.getParserServices` is called lazily and only when the option is enabled, so the rule has no hard dependency on typed linting. With the option off, a pair where either side has no written type argument is assumed compatible — the rule has no way to know better without type information, and this is the documented default.

### `input.required()` support

`` input.required<T>() `` is now recognized and migrated to `` model.required<T>() `` — it was previously ignored because the selector only matched the bare `input` identifier. Note that `input.required()` (like `output()`) takes an options object as its first argument, so its type only ever comes from the explicit `<T>` and is never inferred from that argument. The `transform` guard added in #3050 now applies to `input.required()` as well, since `model.required()` has no transform support either.

### The fixer now targets the callee node

The fix replaces the `input` callee identifier instead of regex-replacing the first `input` word of the property text, which used to rename a property literally named `input`:

```ts
readonly input = input<string>();
readonly inputChange = output<string>();

// before: readonly model = input<string>();   ← property renamed, call untouched
// after:  readonly input = model<string>();
```

This also makes `input.required()` → `model.required()` safe.

## Notes

- Rebased onto current `main`, so the `transform` skip from #3050 is integrated into the restructured collector rather than merged around it.
- The typed test cases opt in per-case via `options: [{ useTypeChecking: true }]`, so the default (syntactic) path is exercised too; the rule file is at 100% statement/branch/function coverage.
- Not addressed here: both maps are still cleared on every `ClassDeclaration:exit`, so a nested class declaration's exit wipes pairs already collected for the enclosing class. Also, a multi-type-argument input such as `` input<number, string>(0, { transform: numberAttribute }) `` is now skipped by the `transform` guard, but a bare `` input<number, string>() `` would still be reported. Both pre-existing, and left as follow-ups as suggested in review.


JamesHenry (changes requested):
> [!NOTE]
> 🤖 This is an **AI-generated review** (Claude), prepared and posted at the maintainer's request. The maintainer has read and endorsed its content.

Thanks for this PR! The core insight is absolutely right — `model()` exposes a single type for both directions, so merging a type-mismatched `input`/`output` pair produces incorrect code, and the rule must not report/autofix those. We want to land this. However, the current revision isn't merge-ready yet. Details below.

## 1. The PR description claims features the diff doesn't implement

The description says the input's type is "inferred from its initial value" and that `input.required()` is "now recognized and migrated to `model.required()`". Neither is in the diff:

- The selectors are still `PropertyDefinition > CallExpression[callee.name='input']` / `[callee.name='output']`. `input.required()` has a `MemberExpression` callee, so `callee.name` never matches and those properties are never collected.
- `areTypesCompatible` only ever reads the explicit `typeArguments` — there is no initializer inference anywhere.

Please either implement those cases or trim the description to match the diff. (Side note: GitHub's markdown ate the `<string>`/`<number>` generics written outside code spans in the description, so most of its examples render as bare `input()` + `output()` — worth fixing with backticks when you update it.)

## 2. A missing type argument is treated as a match — which still produces the bad fix

`areTypesCompatible` early-returns `true` when either side lacks a type argument, before the checker is even consulted. So `input('')` + `output<number>()` is still reported and autofixed to `model('')` paired with removal of the number-emitting output — the exact case the description claims is "correctly skipped". This needs to be handled (via inference when type information is available), or at minimum the behavior needs to be an explicit, documented decision rather than an accidental fall-through.

## 3. Type comparison relies on `ts.Type` object identity

```ts
typeServices.getTypeAtLocation(input.typeArgument) ===
  typeServices.getTypeAtLocation(output.typeArgument)
```

Identity happens to work for the union test case (the checker interns unions), but structurally identical types written at two locations are often distinct objects — e.g. two inline `{ a: string }` annotations — so this silently skips valid merges. Use the type checker instead. Note `isTypeIdenticalTo` is internal API; the public route is mutual assignability:

```ts
checker.isTypeAssignableTo(a, b) && checker.isTypeAssignableTo(b, a)
```

## 4. Follow the `prefer-signals` pattern: add a `useTypeChecking` option (default `false`)

As written, the rule silently changes behavior depending on whether the consumer's parser setup provides a Program. House style (see `prefer-signals`) is an explicit `useTypeChecking` option defaulting to `false`, with `ESLintUtils.getParserServices(context)` called lazily only when the option is enabled. Syntactic comparison when off; checker when on; never a hard requirement on typed linting.

This also fixes the test/coverage problem: keeping `projectService: true` on the single `RuleTester` is fine (it matches `prefer-signals`' spec), but typed cases should opt in per-case via `options: [{ useTypeChecking: true }]` so the default cases exercise the untyped path. Right now every test runs the typed path, which is why Codecov flags 1 uncovered line — the `sourceCode.getText(...)` syntactic fallback, i.e. the entire untyped code path, currently ships untested.

Adding the option also means regenerating the generated files (`pnpm nx run-many -t update-rule-docs / update-rule-lists / update-rule-configs`) so the schema/docs checks stay green.

## 5. The fix should replace the callee node, not regex the property text

`inputText.replace(/\binput\b/, 'model')` replaces the *first* word-boundary match in the whole property text. A property literally named `input` breaks today:

```ts
readonly input = input<string>();
readonly inputChange = output<string>();
// fix produces: readonly model = input<string>();  ← property renamed, call untouched
```

This flaw predates this PR, but since the fixer is being rewritten here anyway, please target the callee identifier node with `fixer.replaceText(...)` instead. That also becomes a hard requirement if `input.required()` support is added.

## 6. Optional: pre-existing nested-class bug

Both maps are cleared on every `ClassDeclaration:exit`, so a nested class declaration's exit wipes pairs already collected for the enclosing class. Pre-existing and not a blocker — fine as a follow-up issue, but worth noting since this PR restructures exactly that code.

## 7. Rebase

CI is green but last ran on 2026-07-12; the base has moved since. Please rebase onto current `main` so a fresh run validates the final revision.

---
_Generated by [Claude Code](https://claude.ai/code)_

JamesHenry (changes requested):
> [!NOTE]
> 🤖 This is an **AI-generated review** (Claude), prepared and posted at the maintainer's request. The maintainer has read and endorsed its content.

Thanks for the thorough update — every point from the previous review has been addressed properly, and the new tests cover exactly the cases we called out (structural identity, initial-value inference, `input.required()`, the property named `input`, and the default vs. typed paths). Fresh CI is green and the rule sits at 100% coverage. One small correctness gap in the new typed path to close before merging, then this is good to go.

## 1. Unresolvable types are treated as a match under `useTypeChecking`

When a written type argument doesn't resolve (a typo, a package without type declarations, a stale import), the checker returns its error type. That type carries `TypeFlags.Any` and is assignable in both directions, so this pair is reported and autofixed:

```ts
readonly value = input<Foo>();
readonly valueChange = output<Bar>();
```

The syntactic default correctly skips it, so the opt-in mode ends up less safe than the default for this case. Suggested fix — fall back to the text comparison whenever either side resolves to an `any`-like type:

```ts
if ((inputType.flags | outputType.flags) & ts.TypeFlags.Any) {
  return haveSyntacticallyEqualTypes(input, output);
}
```

(where `haveSyntacticallyEqualTypes` is the existing `!useTypeChecking` branch extracted into a helper — `src/utils/jsdoc.ts` already imports `typescript` directly, so that's fine here too). I verified locally that this passes the existing spec and the `Foo`/`Bar` case. A valid test case for it would be great.

## 2. Non-blocking

- **Transform inputs.** `input<number, string>(0, { transform: numberAttribute })` paired with `output<number>()` is still reported, and the fix keeps the second type argument and the `transform` option on `model()`, which Angular rejects. Pre-existing, but it's the same class of bug this PR targets. A guard on `typeArguments.params.length > 1` or a `transform` key in the options object would cover it — happy for that to be a follow-up together with the nested-class issue.
- **PR description.** The Behaviour section still has the generics eaten by GitHub markdown, so most examples render as bare `input()` + `output()`. Worth wrapping them in backticks.

---
_Generated by [Claude Code](https://claude.ai/code)_

JamesHenry (approved):

PR #3090, opened 2026-06-28
feat(eslint-plugin): [reactive-context-must-read-signal] add new rule

anousss007:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added
- [x] Docs have been added / updated

## PR Type

- [x] Feature

## What is the new behavior?

Adds a new type-aware rule, `@angular-eslint/reactive-context-must-read-signal`, that reports a reactive context whose tracked function never reads a signal. Such a context runs once and never re-runs, which is almost always a mistake: either a signal was not called (`this.name` instead of `this.name()`), or the value is static and should not be reactive at all.

```ts
const value = 42;
const c = computed(() => value * 2); // ❌ reads no signal, never recomputes

const count = signal(0);
const c2 = computed(() => count() * 2); // ✅
```

Covered reactive contexts:

| API | Tracked function | Default |
| --- | --- | --- |
| `computed(fn)` | `fn` | on |
| `effect(fn)` | `fn` | on |
| `linkedSignal(fn)` / `linkedSignal({ source, computation })` | `fn`, or `source` and `computation` | on |
| `afterRenderEffect(fn)` / `afterRenderEffect({ earlyRead, write, mixedReadWrite, read })` | `fn`, or each phase | on |
| `resource({ params })` / `rxResource({ params })` | `params` (or legacy `request`) | opt-in via `checkResources` |

Detection is type-aware and reuses `KNOWN_SIGNAL_TYPES` from `no-uncalled-signals`, so the rule requires type information.

The rule only reports when everything the tracked function does is known not to read a signal. Calls to the TypeScript standard library (arrays, strings, `Math`, `JSON`, `console`, DOM APIs) are known safe unless they receive a function value the rule cannot see inside. Anything else keeps the rule silent: helper functions, service methods, getters, `new` on user classes, tagged templates, unresolved symbols, and every Angular API including `untracked()`.

Added to the `ts-all` config only.

## Does this PR introduce a breaking change?

- [x] No

## Other information

Originates from https://github.com/angular-eslint/angular-eslint/discussions/2970.


anousss007:
Any news on this?

JamesHenry (changes requested):
Thanks for this @anousss007, and I'm very sorry for the delay. This one required a bit more thinking.

The idea is good and the implementation is well structured. I want to land it, but there are some blockers first, plus a rename.

## Rename

Please rename the rule to `reactive-context-must-read-signal`. The current name says `computed`, but the rule covers `computed`, `effect`, `linkedSignal` and `resource`. "Reactive context" is the term the Angular docs use for all of these, and the "must" form matches the existing `computed-must-return`. Update the rule file, the tests directory and the config entry, then regenerate docs and READMEs with the repo scripts.

Change the message to match: `{{primitive}}() must read at least one signal, otherwise it will never re-run when state changes.`

## Blockers

1. **Typecheck fails.** In `cases.ts` line 120, the `.map()` after the array literal widens `options` from a tuple to an array, so the result is not assignable to `InvalidTestCase<MessageIds, Options>`. Give the array a type before mapping:

   ```ts
   const invalidBase: readonly InvalidTestCase<MessageIds, Options>[] = [ /* cases */ ];
   export const invalid: readonly InvalidTestCase<MessageIds, Options>[] =
     invalidBase.map((test) => ({ ...test, code: appendTypes(test.code) }));
   ```

   Confirm with `pnpm nx run eslint-plugin:typecheck`. CI has not run on this PR yet, which is why this was not caught.

2. **False positive on the standard `linkedSignal` usage.** The Angular docs pass a signal directly as `source`. The rule skips that value because it is not a function literal, analyses only `computation`, and reports:

   ```ts
   selected = linkedSignal({ source: this.options, computation: (opts) => opts[0] });
   ```

   Fix in `registerTrackedFunctions`: if a tracked slot (positional arg or option key) holds anything other than an inline function, set `hasUnknownCall = true` so the rule stays silent. This also covers `source: this.someFn`.

3. **False positive on getters.** `computed(() => this.name)` is reported when `name` is a getter that reads a signal. For a `MemberExpression`, get the symbol via `services.getSymbolAtLocation(node)` and treat the read as unknown when `symbol.flags & ts.SymbolFlags.GetAccessor` is set.

4. **`new` and tagged templates are not treated as calls.** A constructor or a template tag can read a signal. Handle `NewExpression` and `TaggedTemplateExpression` the same way as an unknown `CallExpression`.

5. **Any call at all silences the rule, which leaves it with very little real-world coverage.** Today the rule reports only when the callback contains no calls other than signal reads. `computed(() => this.firstName + ' ' + this.lastName)` is caught, but `computed(() => this.items.filter(...))` or `computed(() => this.name.toUpperCase())` with the same missing parentheses is not. For `effect`, coverage is close to zero, because an effect body nearly always calls something.

   Narrow the bail-out: when the callee's symbol resolves to a declaration in a TypeScript default library file (`program.isSourceFileDefaultLibrary(decl.getSourceFile())`), treat the call as safe rather than unknown. That covers array, string, `Math`, `JSON`, `Object`, `console` and DOM APIs, none of which can read an Angular signal. Inline callbacks are already analysed by the frame stack. If any argument is a function-typed value that is not an inline function literal (for example `this.items.map(this.transform)`), keep bailing. Anything without a resolvable symbol, and every Angular API including `untracked`, stays unknown.

6. **Add `afterRenderEffect`.** It is a reactive context like `effect`, with the same one-shot failure mode, and `afterNextRender` is the correct alternative. Support the single-callback form and the phases object (`earlyRead`, `write`, `mixedReadWrite`, `read`) using the existing `args`/`optionKeys` config. Add it to the imports in `appendTypes`.

7. **Merge conflict** in `packages/eslint-plugin/src/index.ts` from a rule that landed on `main`. Please merge `main` in.

Please add valid and invalid tests for each of items 2 to 6.

Once the above is in, I will approve the CI run and we can take it from there.

anousss007:
Thanks for your feedback.

Let me know if I can do anything else.

JamesHenry (approved):

JamesHenry:
Thank you!

PR #3091, opened 2026-06-29
feat(eslint-plugin-template): add `require-switch-default` rule

g-drouard:
## Overview

Adds a new template rule, `@angular-eslint/template/require-switch-default`, which requires every `@switch` block to declare a `@default`. It addresses the same need as [`@typescript-eslint/switch-exhaustiveness-check`](https://typescript-eslint.io/rules/switch-exhaustiveness-check/) — making sure no value falls through a switch unhandled — but it is **not a strict equivalent** (see below).

## Motivation

A `@switch` with no `@default` silently renders nothing when the value matches none of the `@case` branches, which is rarely intended and easy to miss during review.

Unlike the TypeScript rule, true type-based exhaustiveness checking is **not possible** in Angular templates: template rules have no access to the TypeScript type checker, so the linter cannot know whether the listed `@case`s are exhaustive over a union or enum.

The realistic, robust equivalent is therefore to require an explicit `@default`, which guarantees unhandled values are always accounted for. For genuine compile-time exhaustiveness, Angular's own `@default never` form can be used — the rule accepts it.

## Behaviour

The rule reports a `@switch` unless it contains one of:

- a plain `@default { ... }` block, or
- the exhaustive `@default never;` form, or
- the exhaustive `@default never(value);` form (passing the switch value so the template type-checker asserts it is `never`).

### Examples

❌ Invalid:

```html
    @switch (value) {
      @case (1) { one }
    }
```

✅ Valid:

```html
    @switch (value) {
      @case (1) { one }
      @default { other }
    }
```

```html
    @switch (value) {
      @case (1) { one }
      @default never;
    }
```

```html
    @switch (item.kind) {
      @case ('a') { a }
      @default never(item.kind);
    }
```

## Notes

- No configuration options (option-less, like the related `no-empty-control-flow`).
- Not added to the `recommended` config — it is opt-in via `all` only.
- Docs, rule lists, and configs were regenerated with the standard `update-*` scripts.

## Checklist

- [x] Rule implementation + tests (`pnpm nx test eslint-plugin-template`)
- [x] `pnpm format-check`
- [x] `pnpm nx sync:check`
- [x] `pnpm nx run-many -t check-rule-docs check-rule-lists check-rule-configs`
- [x] Lint + typecheck pass


JamesHenry (approved):
Awesome, thanks!

PR #3092, opened 2026-07-02
feat(eslint-plugin-template): add `no-outerhtml` rule

g-drouard:
## Overview

Adds a new template rule, `@angular-eslint/template/no-outerhtml`, which disallows binding to the `[outerHTML]` DOM property in Angular templates.

## The problem

Binding to `[outerHTML]` looks harmless and even renders correctly on first paint, but it is **broken by design** and throws as soon as the bound value updates:

    <p [outerHTML]="html">lib works!</p>

The first render works. On the next change-detection cycle (e.g. the property changes, or a test calls `fixture.detectChanges()`), Angular throws:

    NoModificationAllowedError: Failed to set the 'outerHTML' property on 'Element': This element has no parent node

### Root cause

Setting `outerHTML` replaces the **element itself** in the DOM — including the exact node Angular is bound to. Angular keeps an internal reference to that node so it can apply future updates. Once the node has been swapped out it no longer has a parent in the document, so the next write to `outerHTML` fails.

This is inherent to how the DOM `outerHTML` setter works, not a fixable framework bug. Both upstream Angular issues were closed without a fix:

- angular/angular#41131
- angular/angular#41576

An Angular team member confirmed this is not supported ([angular/angular#41576 (comment)](https://github.com/angular/angular/issues/41576#issuecomment-818242377)):

> Angular doesn't support binding to `outerHTML` like this because `outerHTML` replaces the element upon which it is set.

As a result there is effectively **no correct way** to use `[outerHTML]` in an Angular template, which makes it a good fit for a lint rule (near-zero false positives).

## What the rule does

`outerHTML` is a DOM property, so the rule targets **native elements only** and reports every form of `outerHTML` on them — none of which is ever valid:

- property bindings: `[outerHTML]="..."` and the `bind-outerHTML="..."` canonical form;
- the attribute binding `[attr.outerHTML]="..."`;
- the static attribute `outerHTML="..."` (not a real DOM attribute — it has no effect).

Not reported:

- A binding on a **component** whose input happens to be named `outerHTML` (e.g. `<my-component [outerHTML]="html">`). There it is an ordinary input, not the DOM property, so it is left alone.

### Suggested fix (in the message)

In almost all cases the intent is to render dynamic markup _inside_ the element, which is what `[innerHTML]` does safely — it replaces only the element's content and preserves the host node:

    <div [innerHTML]="html"></div>

The message intentionally does **not** present this as a mechanical 1:1 replacement:

- `[innerHTML]` replaces content only, not the element (tag/attributes). If the element itself needed to change, the component should be restructured (e.g. conditional rendering with `@if` / `@switch`).
- `[innerHTML]` runs Angular's built-in sanitization; standard XSS/sanitization concerns still apply when rendering untrusted content.

## Notes

- Type: `problem` (this is a latent runtime error, not a style preference).
- No configuration options.
- Not added to the `recommended` config — opt-in via `all` only.
- No autofix, because `outerHTML` -> `innerHTML` is not semantically equivalent.
- Docs, rule lists, and configs were regenerated with the standard `update-*` scripts.

## Checklist

- [x] Rule implementation + tests (`pnpm nx test eslint-plugin-template`)
- [x] `pnpm format-check`
- [x] `pnpm nx sync:check`
- [x] `pnpm nx run-many -t check-rule-docs check-rule-lists check-rule-configs`
- [x] Lint + typecheck pass


JamesHenry (approved):
Thank you!

PR #3111, opened 2026-07-14
feat(eslint-plugin): add `prefer-queries` rule

SchroederSteffen:
Direct DOM querying via `document.getElementById()`, `document.querySelector()`, and similar methods bypasses Angular's component encapsulation and change detection.
These calls create tight coupling between component logic and the DOM structure, making the code fragile to template changes.

Angular provides first-class [query APIs](https://angular.dev/guide/components/queries) — `viewChild()`, `viewChildren()`, `contentChild()`, and `contentChildren()` — that integrate with the component lifecycle, work with Angular's rendering pipeline, and are automatically updated when the view changes.
Using Angular queries also improves testability because they can be resolved in Angular's testing environment without requiring a real DOM.

Resolves https://github.com/angular-eslint/angular-eslint/discussions/2702.


packages/eslint-plugin/docs/rules/prefer-queries.md
```diff
@@ -0,0 +1,739 @@
...
+The rule accepts an options object with the following properties:
+
+```ts
+interface Options {
+  /**
+   * Also check @Injectable(), @Pipe() & @Service() in addition to @Component() & @Directive().
+   *
+   * Default: `false`
+   */
+  checkServices?: boolean;
```
SchroederSteffen, line 41: I'd even turn this on by default, but it might be a matter of taste.

JamesHenry (changes requested):
Thanks for this PR @SchroederSteffen and I'm sorry for the delay.

Having thought about it, this PR kind of mixes two different concerns right now. I think we need to separate them into two different rules.

**Rule 1: no global `document`**

This rule must report any use of the global `document` inside an Angular class. The fix is `inject(DOCUMENT)`. This covers all uses, not only query methods. Examples: `document.title`, `document.body.classList`, `document.addEventListener()`. The reason is platform safety. The global does not exist on the server. This rule can be exact, and there is no guesswork.

**Rule 2: prefer Angular queries**

This rule must report DOM lookups for elements in the component's own template. The fix is `viewChild()`, `viewChildren()`, `contentChild()`, or `contentChildren()`. The reason is encapsulation. This rule applies to any document, including an injected `DOCUMENT`. It also applies to `this.elementRef.nativeElement.querySelector()`. A lint rule cannot know if a selector points into the template. So this rule is always a guess. It will have false positives.

The current PR does parts of both. It matches any object named `document`, which is Rule 1 logic. It only checks query methods and suggests `viewChild()`, which is Rule 2 logic. The result is not really correct for either rule right now.

**Request**

Please change this PR to implement Rule 1 only. Rule 2 can be a separate proposal later. For Rule 2, we would need an explicit heuristic design first. For example, report only `getElementById()` and `querySelector()` with a literal argument.

**Problems found in the current code**

I got my agent to test the rule with extra cases. These are the results.

- `this.document.getElementById('x')` where `document = inject(DOCUMENT)` is reported. This is wrong. It is the exact pattern from the Angular SSR guide. The rule must never report a class member.
- The same happens with `@Inject(DOCUMENT) private document: Document` in a constructor.
- Renaming the field to `doc` stops the report. This shows the rule checks a property name, not the global.
- `document.body.querySelector('x')` is not reported. `document.documentElement.querySelectorAll('x')` is not reported. Any property between `document` and the method call bypasses the rule.
- `iframe.contentWindow.document.querySelector('x')` is reported. Angular queries cannot reach an iframe.
- A method parameter named `document` is reported. The rule does not use scope analysis.
- A plain class declared inside a component method is reported.
- `document.title` is accepted. Under Rule 1, this must be reported.

**Changes needed for Rule 1**

1. Use scope analysis to find the real global `document` identifier. Report only that identifier. Use `context.sourceCode.getScope()` and check that the variable is not declared in any local scope.
2. Also report `window.document` and `globalThis.document`.
3. Report every use of the global, not only the query methods. Remove the method list.
4. Never report `this.document` or any other member expression.
5. Rename the rule. `prefer-queries` does not describe Rule 1. Suggested names: `no-global-document` or `prefer-document-token`.
6. Change the message. Suggested text: "Do not use the global `document`. It is not available during server-side rendering. Inject the `DOCUMENT` token instead: `private readonly document = inject(DOCUMENT);`"
7. Rewrite the rationale. The current text says direct queries bypass change detection. That is not correct. It says Angular queries do not need a real DOM in tests. That is also not correct. The real reasons are server-side rendering, hydration, and encapsulation. Link to the SSR guide section above.
8. Keep the `checkServices` option. Services and pipes also run on the server, so Rule 1 applies to them. Consider making it `true` by default for this new rule.
9. As you already did, keep the rule out of the `recommended` config. Put it in `ts-all` only. It will be noisy on existing code.

**Tests to add**

- Valid: `inject(DOCUMENT)` imported from `@angular/core`.
- Valid: `inject(DOCUMENT)` imported from `@angular/common`.
- Valid: `@Inject(DOCUMENT)` constructor parameter.
- Valid: a parameter or local variable named `document`.
- Valid: `this.document.title`.
- Invalid: `document.title`.
- Invalid: `document.body.querySelector('x')`.
- Invalid: `document.addEventListener('click', fn)`.
- Invalid: `window.document.getElementById('x')`.
- Invalid: `globalThis.document.getElementById('x')`.
- Invalid: `document?.querySelector('x')`.

SchroederSteffen:
Thank you for the response.
I have been on vacation for the past few weeks and will need some time to incorporate your feedback.

PR #3112, opened 2026-07-14
feat(eslint-plugin-template): [no-inline-styles] allow [style] bindings via allowBindToStyle

### LINKED ISSUE #3016: [@angular-eslint/template/no-inline-styles] conflicts with Angular style guide for [style] object bindings
Body
### Description
The `@angular-eslint/template/no-inline-styles` rule currently reports `[style]` bindings even when the value is an object / style map.
At the same time, Angular's official style guide recommends preferring native `class` and `style` bindings over `ngClass` and `ngStyle`, including object/map binding syntax for `[style]`.
This creates an awkward situation:
- `ngStyle` can be allowed with `allowNgStyle: true`
- `[style.prop]` can be allowed with `allowBindToStyle: true`
- but `[style]="styleObject"` is still reported by the rule
That means teams following the Angular style guide still get lint errors for the recommended native `[style]` syntax.
### Angular Style Guide
Angular recommends:
> Prefer `class` and `style` bindings over using the `NgClass` and `NgStyle` directives.
Example from the style guide:
```html
<div [style]="{'color': textColor, 'background-color': backgroundColor}"></div>
```
https://angular.dev/style-guide#prefer-class-and-style-over-ngclass-and-ngstyle
#### Current behavior
With this config:
```
{
  "rules": {
    "@angular-eslint/template/no-inline-styles": [
      "error",
      {
        "allowBindToStyle": true
      }
    ]
  }
}
```
This is allowed:
```html
<div [style.color]="textColor"></div>
```
But this is still reported:
```html
<div [style]="{ 'color': textColor, 'background-color': backgroundColor }"></div>
```
#### Expected behavior
It would be useful to support one of the following:
1. Extend allowBindToStyle so it also allows [style] map/object bindings
2. Add a separate option specifically for [style] map/object bindings

For example, something like:
```
{
  "rules": {
    "@angular-eslint/template/no-inline-styles": [
      "error",
      {
        "allowBindToStyle": true,
        "allowStyleMapBinding": true
      }
    ]
  }
}
```
#### Why this would help
Using [style] with an object is:
- native Angular syntax
- recommended by the Angular style guide over ngStyle
- often more ergonomic than expanding everything into multiple [style.prop] bindings
Right now the main workaround is switching back to ngStyle, but that goes against Angular's own recommendation.

#### Reproduction
```html
<div [style]="{ 'color': textColor, 'background-color': backgroundColor }"></div>
```
```
{
  "rules": {
    "@angular-eslint/template/no-inline-styles": [
      "error",
      {
        "allowBindToStyle": true
      }
    ]
  }
}
```
#### Environment
- `@angular-eslint/eslint-plugin-template`: `21.3.1`

xonaib:
## Summary
- extend `allowBindToStyle` so it covers `[style]`, `bind-style`, interpolated `style`, and `[style.prop]` bindings
- keep static `style` attributes and `[attr.style]` invalid in all configurations
- classify style inputs using Angular binding types instead of matching `keySpan.details`
- add regression coverage for `[attr.stylesheet]` and the requested style binding cases

Fixes #3016

## Checks
- `node ..\..\node_modules\vitest\vitest.mjs run tests\rules\no-inline-styles\spec.ts --config vitest.config.mts`
- `node node_modules\typescript\bin\tsc --build packages\eslint-plugin-template\tsconfig.spec.json --pretty false`
- `node node_modules\prettier\bin\prettier.cjs --check packages\eslint-plugin-template\src\rules\no-inline-styles.ts packages\eslint-plugin-template\tests\rules\no-inline-styles\cases.ts packages\eslint-plugin-template\docs\rules\no-inline-styles.md`

### DIFF
--- a/packages/eslint-plugin-template/src/rules/no-inline-styles.ts
+++ b/packages/eslint-plugin-template/src/rules/no-inline-styles.ts
@@ -9,11 +9,13 @@ export type Options = [
   {
     readonly allowNgStyle?: boolean;
     readonly allowBindToStyle?: boolean;
+    readonly allowStyleObjectBinding?: boolean;
   },
 ];
 const DEFAULT_OPTIONS: Options[number] = {
   allowNgStyle: false,
   allowBindToStyle: false,
+  allowStyleObjectBinding: false,
 };
 export type MessageIds = 'noInlineStyles';
 export const RULE_NAME = 'no-inline-styles';
@@ -37,6 +39,10 @@ export default createESLintRule<Options, MessageIds>({
             type: 'boolean',
             default: DEFAULT_OPTIONS.allowBindToStyle,
           },
+          allowStyleObjectBinding: {
+            type: 'boolean',
+            default: DEFAULT_OPTIONS.allowStyleObjectBinding,
+          },
         },
         additionalProperties: false,
       },
@@ -47,28 +53,25 @@ export default createESLintRule<Options, MessageIds>({
     },
   },
   defaultOptions: [DEFAULT_OPTIONS],
-  create(context, [{ allowNgStyle, allowBindToStyle }]) {
+  create(
+    context,
+    [{ allowNgStyle, allowBindToStyle, allowStyleObjectBinding }],
+  ) {
     const parserServices = getTemplateParserServices(context);
 
     return {
       Element(node: TmplAstElement) {
-        let isInvalid = false;
-
-        if (!allowNgStyle && !allowBindToStyle) {
-          isInvalid =
-            isNodeHasStyleAttribute(node) ||
-            isNodeHasNgStyleAttribute(node) ||
-            isNodeHasBindingToStyleAttribute(node);
-        } else {
-          const ngStyle = allowNgStyle
-            ? false
-            : isNodeHasNgStyleAttribute(node);
-          const bindToStyle = allowBindToStyle
-            ? false
-            : isNodeHasBindingToStyleAttribute(node);
-
-          isInvalid = isNodeHasStyleAttribute(node) || ngStyle || bindToStyle;
-        }
+        const hasDisallowedNgStyle =
+          !allowNgStyle && isNodeHasNgStyleAttribute(node);
+        const hasDisallowedBindToStyle =
+          !allowBindToStyle && isNodeHasBindingToStyleAttribute(node);
+        const hasDisallowedStyleObjectBinding =
+          !allowStyleObjectBinding && isNodeHasStyleObjectBinding(node);
+        const isInvalid =
+          isNodeHasStyleAttribute(node) ||
+          hasDisallowedNgStyle ||
+          hasDisallowedBindToStyle ||
+          hasDisallowedStyleObjectBinding;
 
         if (isInvalid) {
           const loc = parserServices.convertElementSourceSpanToLoc(
@@ -95,7 +98,7 @@ export default createESLintRule<Options, MessageIds>({
 function isNodeHasStyleAttribute(node: TmplAstElement): boolean {
   return (
     node.attributes.some(({ name }) => isStyle(name)) ||
-    node.inputs.some(({ name }) => isStyle(name))
+    node.inputs.some(({ keySpan }) => isAttributeStyleBinding(keySpan))
   );
 }
 /**
@@ -113,6 +116,15 @@ function isNodeHasBindingToStyleAttribute(node: TmplAstElement): boolean {
 }
 
 /**
+ *  Check that an element (for example `<img>`) has a `[style]` object/map binding.
+ */
+function isNodeHasStyleObjectBinding(node: TmplAstElement): boolean {
+  return node.inputs.some(
+    ({ name, keySpan }) => isStyle(name) && !isAttributeStyleBinding(keySpan),
+  );
+}
+
+/**
  *  Check element is style
  */
 function isStyle(name: string): name is 'style' {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-template/src/rules/no-inline-styles.ts
```diff
@@ -112,6 +115,15 @@ function isNodeHasBindingToStyleAttribute(node: TmplAstElement): boolean {
...
 }
 
+/**
+ *  Check that an element (for example `<img>`) has a `[style]` object/map binding.
+ */
+function isNodeHasStyleObjectBinding(node: TmplAstElement): boolean {
+  return node.inputs.some(
+    ({ name, keySpan }) => isStyle(name) && !isAttributeStyleBinding(keySpan),
+  );
+}
```
MariaMikhnevich: [According to the Angular documentation](https://angular.dev/guide/templates/binding#css-style-properties), `[style]` can also accept a string. However, the option's name suggests (`allowStyleObjectBinding`) that enabling it in `[style]` will only allow passing objects. However, if I understand your logic correctly, passing a string will also be allowed, and this won't be an invalid case. Perhaps, in this case, it would be worth renaming the option to make it more descriptive, or tightening the validation by also checking whether the value is an object (if this is feasible)? What do you think about this?

MariaMikhnevich: If tightening validation is possible and won't significantly impact performance and speed, how about adding two options, one that would only allow `[style]` objects, and one that would just allow `[style]`? Or one option that would accept a union of string literal types rather than a boolean value. This would allow for more precise customization.

xonaib: Good point. The current implementation effectively allows `[style]` bindings generally, not only object/map values, because Angular also permits strings and identifiers like `[style]="styleObject"` cannot be reliably classified as object-only without type information.

I can rename the option to something broader like `allowStyleBinding` so the option name matches the behavior. Does that sound good, or maybe if you another suggestion?

MariaMikhnevich: Since, judging by your words, tightening validation is impossible, then yes, let's rename the option. Your suggestion sounds good. However, in this case, perhaps should also expand the documentation examples and tests.

xonaib: Thanks. I’ve renamed the option to `allowStyleBinding` and updated the docs/tests to cover the broader `[style]` behavior, including string, object literal, and identifier examples. 

Let me know if I should change something.

MariaMikhnevich (approved):

MariaMikhnevich (approved):

xonaib:
@MariaMikhnevich  How do i trigger these 4 workflows, so this can be merged?

MariaMikhnevich:
> How do i trigger these 4 workflows, so this can be merged?

@xonaib What do you mean?

xonaib:
@MariaMikhnevich, oh i just wanted to ask how do I merge this PR, its awaiting 4 workflows approval checks, Do i need to wait for some one else, do something else?
{9FF32D19-1073-4B9E-B759-F1D139DF473D}

MariaMikhnevich:
@xonaib
Unfortunately, I can't offer any advice on this matter, as I'm not part of this project's team. I tried to approve this PR, but it seems to be failing because I don't have access. It likely requires at least one approval from people with the necessary rights. Regarding the reviews, judging by the message "4 workflows awaiting approval," they also require permission from people with the necessary rights.
image

@JamesHenry, could you please look at the current PR or appoint people who could do this?

JamesHenry (changes requested):
Thanks for taking this on @xonaib, and I'm sorry for the delay in reviewing.

The motivation is spot on: the rule already lets teams opt into `ngStyle` and `[style.prop]`, so leaving `[style]` out was an inconsistency, especially now that Angular recommends it over `ngStyle`. I'd like to get this in, with a few changes first.

### 1. Extend `allowBindToStyle` rather than adding a new option

`[style]` and `[style.prop]` are the same Angular mechanism (both go through the styling runtime rather than setting a `style` attribute), and `allowBindToStyle` already reads as "allow style bindings". A third boolean next to it, especially one named `allowStyleBinding`, is going to be hard for users to tell apart. So let's make it that when  `allowBindToStyle: true` it covers `[style]` as well. It's a small loosening for people who already opted in, which I'm comfortable shipping in a non-breaking.

### 2. Classify bindings by type instead of matching on `keySpan.details`

The Angular parser already tells us what each binding is, so we can drop the string matching. This also fixes a regression in the current diff: `includes('attr.style')` matches any attribute starting with `style`, so `<div [attr.stylesheet]="x">` is now reported and no option can suppress it. Something like:

```ts
import { BindingType } from '@angular-eslint/bundled-angular-compiler';

const hasDisallowedInput = node.inputs.some((input) => {
  switch (input.type) {
    case BindingType.Attribute:
      return input.name === 'style'; // [attr.style] is never allowed
    case BindingType.Style:
      return !allowBindToStyle; // [style.prop]
    case BindingType.Property:
      if (input.name === 'style') return !allowBindToStyle; // [style]
      if (input.name === 'ngStyle') return !allowNgStyle;
      return false;
    default:
      return false;
  }
});
const isInvalid =
  node.attributes.some(({ name }) => name === 'style') || hasDisallowedInput;
```

That replaces all the `isNodeHas*` / `isStyleBound` helpers.

### 3. A few more tests

- `<div [attr.stylesheet]="x">` as a valid case with default options
- `<input [style]="x">` still failing with only `allowNgStyle: true`
- `<input [style.width.px]="w">` under `allowBindToStyle: true`
- One case each for `style="{{ x }}"` and `bind-style="x"`. These parse identically to `[style]`, so they'll now be allowed under the option; a test pins that down.
- Two of the invalid cases share the description "should fail when input element with style binding attribute exist", could you make them distinct?

### 4. Housekeeping

- Rebase onto `main`: #3170 moved `defaultOptions` into `meta`, so there's a conflict.
- Run `pnpm format` and commit the result (`format-check` currently fails on the `create(...)` signature).
- Retitle to something like `feat(eslint-plugin-template): [no-inline-styles] allow [style] bindings via allowBindToStyle`.

Separately, I'll open follow-up issues for a `prefer-style-binding` rule (mirroring `prefer-class-binding` for `ngStyle`) and for making these exceptions distinguish dynamic values from literal ones. Neither needs to hold this PR up.

Many thanks again

JamesHenry (approved):
Thank you! I've filed the follow up issues

PR #3125, opened 2026-08-12
chore: emove unused peer dependency @angular/cli

mrginglymus:
In my usage of angular, I have no direct need of `@angular/cli`, which is an incredibly chunky package (just adding it adds around 90 transitive dependencies). `angular-eslint` is the only package which claims a dependency on it, as a peer. However, I can't find any actual usage of it within `angular-eslint`.

This PR removes it - all tests seem to pass, and it certainly works in the wild on my machine without providing it

### DIFF
--- a/packages/angular-eslint/package.json
+++ b/packages/angular-eslint/package.json
@@ -19,7 +19,6 @@
     "LICENSE"
   ],
   "peerDependencies": {
-    "@angular/cli": "catalog:other-runtime-dependencies",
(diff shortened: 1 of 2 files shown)


JamesHenry:
Thanks for flagging this up @mrginglymus. This one wasn't entirely complete and I also wanted to preserve the user communication element, so I've gone ahead and completed this work in https://github.com/angular-eslint/angular-eslint/pull/3154

PR #3128, opened 2026-08-23
fix(eslint-plugin-template): [valid-aria] validate each token of a token list

### LINKED ISSUE #3124: [valid-aria] False positive on any multi-token tokenlist value (aria-relevant, aria-dropeffect)

`valid-aria` reports `validAriaValue` for any **multi-token** value of an ARIA attribute whose
`aria-query` type is `tokenlist`. A tokenlist is by definition a space-separated list of tokens, so
every legal multi-token value is rejected.

Two attributes are affected, because they are the only `tokenlist` entries in `aria-query@5.3.2`:

- `aria-relevant` — values `additions | all | removals | text`
- `aria-dropeffect` — values `copy | execute | link | move | none | popup`

Notably `aria-relevant="additions text"` is the **ARIA-specified default value** for the attribute,
and it is reported as invalid.

### Cause

`isValidAriaPropertyValue` in `dist/rules/valid-aria.js` handles `token` and `tokenlist` in the
**same** switch branch, and tests the whole attribute string against `values`:

```js
case 'token':
case 'tokenlist': {
  const parsedAttributeValue = isBooleanLike(attributeValue)
    ? JSON.parse(attributeValue)
    : attributeValue;
  return Boolean(values?.includes(parsedAttributeValue));
}
```

`values.includes('additions removals')` is `false`, because `values` holds the individual tokens.
The string is never split on whitespace. `token` is correct here; `tokenlist` needs each
whitespace-separated token validated against `values` instead.

### Reproduction

No StackBlitz link, sorry — the repro needs no Angular app, no template file and no config file, so
a self-contained Node script seemed more useful than a fork. Run this from any project that has the
plugin installed:

```js
// repro.mjs — node repro.mjs
import { Linter } from 'eslint';
import templatePlugin from '@angular-eslint/eslint-plugin-template';
import * as templateParser from '@angular-eslint/template-parser';

const linter = new Linter({ configType: 'flat' });

const cases = [
  ['aria-relevant="additions"', '<div aria-live="polite" aria-relevant="additions"></div>'],
  ['aria-relevant="additions removals"', '<div aria-live="polite" aria-relevant="additions removals"></div>'],
  ['aria-relevant="additions text"', '<div aria-live="polite" aria-relevant="additions text"></div>'],
  ['aria-dropeffect="copy move"', '<div aria-dropeffect="copy move"></div>'],
];

for (const [label, code] of cases) {
  const messages = linter.verify(code, {
    files: ['**/*.html'],
    plugins: { '@angular-eslint/template': templatePlugin },
    languageOptions: { parser: templateParser },
    rules: { '@angular-eslint/template/valid-aria': 'error' },
  }, 'repro.html');
  console.log(label.padEnd(36) + ' -> ' + (messages.length ? messages.map((m) => m.message).join('; ') : 'no error'));
}
```

**Actual output:**

```text
aria-relevant="additions"            -> no error
aria-relevant="additions removals"   -> The `aria-relevant` has an invalid value. Check the valid values at https://raw.githack.com/w3c/aria/stable/#roles
aria-relevant="additions text"       -> The `aria-relevant` has an invalid value. Check the valid values at https://raw.githack.com/w3c/aria/stable/#roles
aria-dropeffect="copy move"          -> The `aria-dropeffect` has an invalid value. Check the valid values at https://raw.githack.com/w3c/aria/stable/#roles
```

**Expected output:** no error for all four. The single-token case already passes, which isolates the
defect to the multi-token path rather than to the attribute being unrecognised.

### Suggested fix

Split `tokenlist` out of the shared branch:

```js
case 'token': {
  const parsed = isBooleanLike(attributeValue) ? JSON.parse(attributeValue) : attributeValue;
  return Boolean(values?.includes(parsed));
}
case 'tokenlist': {
  if (!isString(attributeValue)) return false;
  const tokens = attributeValue.trim().split(/\s+/);
  return tokens.length > 0 && tokens.every((token) => Boolean(values?.includes(token)));
}
```

Happy to raise a PR against that if it is a shape you would take.

### Workaround, for anyone who lands here first

There is no in-place suppression: the rule reports at the attribute's own source location, and an
HTML comment cannot sit between an element's attributes, so `eslint-disable-next-line` cannot reach
it. It takes an `eslint-disable` / `eslint-enable` pair wrapping the element, or dropping the
attribute. We dropped `aria-relevant` — the default `additions text` was what we wanted anyway.

**Versions**

| package                                  | version  |
| ---------------------------------------- | -------- |
| `@angular-eslint/eslint-plugin-template` | `21.4.0` |
| `@angular-eslint/template-parser`        | `21.4.0` |
| `@typescript-eslint/parser`              | `8.63.0` |
| `ESLint`                                 | `10.7.0` |
| `node`                                   | `24.13.0` |

`aria-query` resolves to `5.3.2`. `@angular/core` is `21.2.19`, TypeScript `5.9.3`.

**Also present on latest.** Our workspace is pinned to `21.4.0`, so before filing I installed
`@angular-eslint/eslint-plugin-template@22.1.0` into a scratch directory and checked
`dist/rules/valid-aria.js` — the shared `case 'token': case 'tokenlist':` branch is byte-identical
at lines 128-134, and it resolves the same `aria-query@5.3.2`. So this is not a stale-version
report; the quoted source above is what `22.1.0` ships.

oussamaLaribi:
Closes #3124.

`isValidAriaPropertyValue` handled `token` and `tokenlist` in the same `switch` branch and tested the whole attribute value against `values`. A token list is by definition a whitespace-separated list of tokens, so every legal multi-token value was rejected — including `aria-relevant="additions text"`, which is the ARIA-specified default value for that attribute.

`aria-relevant` and `aria-dropeffect` are the only `tokenlist` entries in `aria-query@5.3.2`, so both were affected.

Full credit to @dungnguyen-dev, who diagnosed this precisely in the issue, down to the offending branch.

### Change

`tokenlist` gets its own branch and each whitespace-separated token is validated on its own. `token` keeps its existing behaviour, including the boolean-like `JSON.parse` path.

I left out the `tokens.length > 0 &&` guard from the suggestion in the issue: `String.prototype.split` never returns an empty array, so `''.trim().split(/\s+/)` yields `['']` and the check is unreachable. Empty and whitespace-only values therefore keep reporting, exactly as before — there is a test pinning that.

### Tests

Added to `valid`:

- `aria-relevant="additions removals"`
- `aria-relevant="additions text"` (the ARIA default)
- `aria-relevant="  additions   text  "` (surrounding and repeated whitespace)
- `aria-dropeffect="copy move"`

Added an `invalid` case so the fix cannot silently become too permissive: a list holding an unknown token (`additions notAToken`, `copy notAToken`) still reports, and so does a value with no token at all.

All four new valid cases fail on `main` and pass with this change. `nx test eslint-plugin-template` is green (1261 tests), as are `lint`, `typecheck` and `format-check`. Rule docs were regenerated with `nx update-rule-docs eslint-plugin-template`.

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/valid-aria.md
+++ b/packages/eslint-plugin-template/docs/rules/valid-aria.md
@@ -110,6 +110,37 @@ The rule does not have any configuration options.
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ```
 
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/valid-aria": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```html
+<div aria-relevant="additions notAToken">Text</div>
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div aria-dropeffect="copy notAToken">Text</div>
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div aria-relevant="   ">Text</div>
+     ~~~~~~~~~~~~~~~~~~~
+```
+
 </details>
 
 <br>
@@ -375,6 +406,110 @@ The rule does not have any configuration options.
 
 ```html
 <div aria-relevant="additions">additions</div>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/valid-aria": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```html
+<div aria-relevant="additions removals">additions removals</div>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/valid-aria": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```html
+<div aria-relevant="additions text">the ARIA default value</div>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/valid-aria": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```html
+<div aria-relevant="  additions   text  ">surrounding and repeated whitespace</div>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/valid-aria": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ✅ Valid Code
+
+```html
+<div aria-dropeffect="copy move">copy move</div>
 ```
 
 <br>
--- a/packages/eslint-plugin-template/src/rules/valid-aria.ts
+++ b/packages/eslint-plugin-template/src/rules/valid-aria.ts
@@ -177,12 +177,17 @@ function isValidAriaPropertyValue(
       return isNumeric(attributeValue);
     case 'string':
       return isString(attributeValue);
-    case 'token':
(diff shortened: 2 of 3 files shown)


JamesHenry (approved):
Thanks @oussamaLaribi

PR #3135, opened 2026-08-23
fix(eslint-plugin-template): [prefer-static-string-properties] escape the fixed attribute value

### LINKED ISSUE #3083: [@angular-eslint/template/prefer-static-string-properties] regression
## Before run eslint --fix

```html
<pre><code [highlightAuto]="'<ng-template #ref=&quot;polymorpheus&quot; polymorpheus>...</ng-template>'"></code></pre>
```

## After run eslint --fix

```html
<pre><code highlightAuto="<ng-template #ref="polymorpheus" polymorpheus>...</ng-template>"></code></pre>
```

```
  27:10  error  Parsing error: Opening tag "code" not terminated

✖ 1 problem (1 error, 0 warnings)
```

oussamaLaribi:
Closes #3083.

The fixer moved the string literal into attribute text verbatim. Whatever the literal held was then re-read by the HTML parser, so the autofix could break the template — or, worse, leave it parsable with a different value.

Three distinct outcomes, all reproducible on `main`:

| Source | Fixed to | Result |
| --- | --- | --- |
| `[name]="'say &quot;hi&quot;'"` | `name="say "hi""` | template no longer parses |
| `[name]='"it&apos;s"'` | `name='it's'` | template no longer parses |
| `[name]="'a &amp;quot; b'"` | `name="a &quot; b"` | parses, but the value silently became `a " b` |

The reported case is the first one. The third is the one I would flag: there is no error to notice, the attribute simply holds something else afterwards.

### Change

Escape `&` and the surrounding quote before emitting the attribute. `&` is escaped first, otherwise it would escape the entities introduced right after it.

The existing quote heuristic is untouched — it still prefers `'` when the binding was written with `"` inside. Escaping makes that choice cosmetic rather than load-bearing, so the fixer is now correct whichever quote it lands on.

### Tests

Four `invalid` cases, one per failure mode plus the template from the issue. The rule tester reports `A fatal parsing error occurred in autofix` for three of them on `main`, and an incorrect output for the ampersand one, so they all fail before the change and pass after.

`nx test eslint-plugin-template` is green (1265 tests), as are `lint`, `typecheck` and `format-check`. Rule docs regenerated with `nx update-rule-docs eslint-plugin-template`.

I also checked the fixed output end to end against the built plugin: each case re-parses and the attribute value round-trips to exactly the original literal.


JamesHenry (approved):
Thanks @oussamaLaribi

PR #3152, opened 2026-08-28
feat(eslint-plugin): [use-injectable-provided-in] add allowProvidedInNull option

### LINKED ISSUE #3120: [use-injectable-provided-in] Allow nulls

I imagine `use-injectable-provided-in` is primarily added to encourage tree shaking through use of `providedIn: 'root'`. However, this feels too strict to cover all scenarios as sometimes the flexibility is needed to be more granular with services and how they are injected. With this rule in place, it is impossible to have a service instance that is scoped to a particular part of the application (after being added to a providers array) without needing to also configure `ignoreClassNamePattern` per usage.

Given the rule is named "use provided in" rather than "prefer provided in root", can we either:
- Allow nulls by default (which would be a non-breaking change but would change the default allowance)
- Have a configuration option added so that `null` is a valid choice so that the current functionality is retained by default

Therefore, if I'm providing `null`, I'm explicitly specifying something rather than leaving the injectable configuration as default (which is questionable). With it being specifically set to `null`, it's seen as a specific and intentional choice and still in line with the naming of the rule. For this reason, my prefer would be to allow null by default.

For info, the code I'm suggestion to be made conditional is here:
https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/src/rules/use-injectable-provided-in.ts#L41

omy13:
## Summary

- Add an `allowProvidedInNull` option (default `false`) to `use-injectable-provided-in`
- `providedIn: null` remains flagged unless users opt in
- `providedIn: undefined` continues to be flagged
- Simplified the `undefined` AST selector per review feedback

Fixes #3120

## Test plan

- [x] Added valid test case for `providedIn: null` with `allowProvidedInNull: true`
- [x] Kept `providedIn: null` as invalid by default
- [x] All rule tests pass (`pnpm nx test eslint-plugin -- use-injectable-provided-in`)
- [x] Rule docs regenerated

### DIFF
--- a/packages/eslint-plugin/src/rules/use-injectable-provided-in.ts
+++ b/packages/eslint-plugin/src/rules/use-injectable-provided-in.ts
@@ -38,10 +38,10 @@ export default createESLintRule<Options, MessageIds>({
       METADATA_PROPERTY_NAME,
     );
     const withoutProvidedInDecorator = `${injectableClassDecorator}:matches([expression.arguments.length=0], [expression.arguments.0.type='ObjectExpression']:not(:has(${providedInMetadataProperty})))`;
-    const nullableProvidedInProperty = `${injectableClassDecorator} ${providedInMetadataProperty}:matches([value.type='Identifier'][value.name='undefined'], [value.type='Literal'][value.raw='null'])`;
+    const undefinedProvidedInProperty = `${injectableClassDecorator} ${providedInMetadataProperty}:matches([value.type='Identifier'][value.name='undefined'], [value.type='Literal'][value.raw='undefined'])`;
     const selectors = [
       withoutProvidedInDecorator,
-      nullableProvidedInProperty,
+      undefinedProvidedInProperty,
     ].join(',');
 
     return {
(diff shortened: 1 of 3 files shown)


JamesHenry (changes requested):
Thanks for your contribution @omy13!

In its current form this PR takes the rule in a new direction that we wouldn't want to take.

`providedIn: null` does not make a service tree-shakeable, the Angular documentation says that null is equivalent to undefined.

- Before this PR, the rule encouraged and made sure that all Injectables were tree-shakeable (it's still in the description even in your PR). Each exception was a visible suppression comment which is a clear signal that cannot go silently unnoticed.
- After this PR, the rule only makes sure that you write a value. TypeScript already does most of that check. Only bare @Injectable() is not caught by TypeScript.

Please instead add an `allowProvidedInNull` option. Set the default to `false`. Users from #3120 can then set it to `true`.

packages/eslint-plugin/src/rules/use-injectable-provided-in.ts
```diff
@@ -38,10 +38,10 @@ export default createESLintRule<Options, MessageIds>({
       METADATA_PROPERTY_NAME,
     );
     const withoutProvidedInDecorator = `${injectableClassDecorator}:matches([expression.arguments.length=0], [expression.arguments.0.type='ObjectExpression']:not(:has(${providedInMetadataProperty})))`;
-    const nullableProvidedInProperty = `${injectableClassDecorator} ${providedInMetadataProperty}:matches([value.type='Identifier'][value.name='undefined'], [value.type='Literal'][value.raw='null'])`;
+    const undefinedProvidedInProperty = `${injectableClassDecorator} ${providedInMetadataProperty}:matches([value.type='Identifier'][value.name='undefined'], [value.type='Literal'][value.raw='undefined'])`;
```
JamesHenry, line 41: The clause `[value.type='Literal'][value.raw='undefined']` can never match. The parser always makes `undefined` an `Identifier` node, not a `Literal` node. The string 'undefined' has a raw value that contains quotes. Thus it also does not match. Recommendation: remove the `:matches()` and keep only `[value.type='Identifier'][value.name='undefined']`.

omy13: Thanks for the review @JamesHenry — that makes sense.

I've updated the PR to add an `allowProvidedInNull` option (default `false`), so `providedIn: null` is still flagged unless users opt in.

Also simplified the `undefined` selector to `[value.type='Identifier'][value.name='undefined']` as you suggested.

JamesHenry (approved):

PR #3169, opened 2026-09-04
feat(eslint-plugin): [inject-at-top] add additionalInjectFunctions option

g-drouard:
## Motivation

`inject-at-top` only knows about the built-in `inject()` call. Codebases commonly wrap it in their own helpers — e.g. an `injectTranslations()` that injects a translation service and derives a scoped accessor from it. Those helpers have exactly the same field-initialization ordering constraint as `inject()` itself, but the rule currently ignores them, so the very bug the rule exists to prevent can still slip through.

## What this PR does

Adds an `additionalInjectFunctions` option: a list of function names to treat like `inject()`.

Each entry is a **regular expression matched against the whole callee name**, so both plain names and patterns work:

```jsonc
{
  "rules": {
    // plain names
    "@angular-eslint/inject-at-top": [
      "error",
      { "additionalInjectFunctions": ["injectTranslations", "injectFeatureFlags"] }
    ]
  }
}
```

```jsonc
{
  "rules": {
    // or a pattern covering a whole family of helpers
    "@angular-eslint/inject-at-top": [
      "error",
      { "additionalInjectFunctions": ["inject[A-Z].*"] }
    ]
  }
}
```

```ts
// with { "additionalInjectFunctions": ["injectTranslations"] }
@Component({})
class MyComponent {
  private count = 0;
  private readonly t = injectTranslations(); // now reported
}
```

Listed functions go through the exact same detection as `inject()`, so they are also found when wrapped (`withScope(injectTranslations())`, IIFEs, arrays, …) and are still ignored inside factory functions.

The report message is now parameterised with the actual function name (`Move this injectTranslations() to the top of the class. …`) instead of hard-coding `inject()`.

## Notes

- Default behaviour is unchanged: with no option, only `inject()` is considered.
- The name must match in full, so `injectTranslationsLater()` is not matched by `injectTranslations`.
- Docs regenerated via `pnpm update-rule-docs`.

<br>

Tests: `pnpm nx test eslint-plugin tests/rules/inject-at-top/spec.ts` passes (54 cases).

### DIFF
--- a/packages/eslint-plugin/docs/rules/inject-at-top.md
+++ b/packages/eslint-plugin/docs/rules/inject-at-top.md
@@ -23,13 +23,25 @@ Requires inject() calls to be declared at the top of the class, before any other
 
 ## Rationale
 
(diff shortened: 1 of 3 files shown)


JamesHenry:
@g-drouard seems reasonable thank you, just has conflicts I imagine because of the default options change

https://github.com/angular-eslint/angular-eslint/commit/9d80d956040fbc6246554f0b42bad37a1a9f28f6

JamesHenry (changes requested):
See comment

JamesHenry:
Nvm, I've done it for you

JamesHenry (approved):

PR #3170, opened 2026-09-05
chore: refactor deprecated defaultOptions

undsoft:
`defaultOptions` in root were deprecated upstream, so I've removed them and moved to recommended `meta.defaultOptions`.
This allowed me to get rid of `patchedRuleCreator`.
However,  `component-selector` and `directive-selector` are special in that they provide meaningless placeholder default options. They also don't match the schema, so if I make the replacement there, eslint won't let it through. I kept them as is and added an exception in `general-rule-docs`.


JamesHenry (approved):
Nice, thanks!

PR #3189, opened 2026-09-07
fix(eslint-plugin-template): [prefer-at-else] avoid unsafe content projection fixes

### LINKED ISSUE #3076: [@angular-eslint/template/prefer-at-else] regression for `Prefer using `@else` instead of a second `@if` clause`
I have https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/src/pages/components/action-bar/examples/1/index.html

```html
@if (!isMobile()) {
        <button
            iconStart="@tui.send"
            tuiButton
            type="button"
        >
            Send
        </button>
    }
    @if (!isMobile()) {
        <button
            iconStart="@tui.trash"
            tuiButton
            type="button"
            (click)="close()"
        >
            Delete
        </button>
    }

    @if (isMobile()) {
        <button
            iconStart="@tui.send"
            tuiIconButton
            type="button"
        >
            Send
        </button>
    }
    @if (isMobile()) {
        <button
            iconStart="@tui.ellipsis"
            tuiIconButton
            type="button"
            (click)="expanded = !expanded"
        >
            More
        </button>
    }

    @if (!isMobile()) {
        <button
            appearance="icon"
            iconStart="@tui.x"
            tuiIconButton
            type="button"
            (click)="close()"
        >
            Close
        </button>
    }
```

After eslint lint fix

```html
@if (isMobile()) {
        <button
            iconStart="@tui.send"
            tuiIconButton
            type="button"
        >
            Send
        </button>

        <button
            iconStart="@tui.ellipsis"
            tuiIconButton
            type="button"
            (click)="expanded = !expanded"
        >
            More
        </button>
    } @else {
        <button
            iconStart="@tui.send"
            tuiButton
            type="button"
        >
            Send
        </button>

        <button
            iconStart="@tui.trash"
            tuiButton
            type="button"
            (click)="close()"
        >
            Delete
        </button>

        <button
            appearance="icon"
            iconStart="@tui.x"
            tuiIconButton
            type="button"
            (click)="close()"
        >
            Close
        </button>
    }
```

Then after build

```bash
 [ERROR] NG8011: Node matches the "a,button,[tuiAction]" slot of the "TuiActionBar" component, but will not be projected into the specific slot because the surrounding @else has more than one node at its root. To project the node in the right slot, you can:

1. Wrap the content of the @else block in an <ng-container/> that matches the "a,button,[tuiAction]" selector.
2. Split the content of the @else block across multiple @else blocks such that each one only has a single projectable node at its root.
3. Remove all content from the @else block, except for the node being projected.
```

This looks like an unsafe autofix.

The original template compiles, but after applying the autofix Angular reports `NG8011: controlFlowPreventingContentProjection`.

The problem is that the fixer groups several projectable sibling nodes into a single `@if/@else` branch with multiple root nodes. This is not equivalent for components that use content projection slots, for example a slot like `a, button, [tuiAction]`.

I think the rule may still report this case, but it should not provide an automatic fix when the generated control-flow block would contain multiple projectable root nodes. In such cases it should either provide a suggestion only, or generate separate `@if` blocks per projected node.

Same problem: 

https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/src/pages/components/table/examples/11/index.html#L66

https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/src/pages/components/select/examples/8/index.html#L15

splincode:
## Summary
- keep reporting `prefer-at-else` violations when opposite control-flow blocks can be combined
- skip the autofix when merging branch contents would create multiple root nodes that can interfere with Angular content projection
- add regression coverage for both unsafe merge directions while preserving existing text-only autofixes

Fixes #3076

## Rationale
Angular control-flow content projection can fail when a projected element is inside a control-flow branch with multiple root nodes. The existing fixer can create exactly that structure when it merges the contents of an existing `@else` or a current `@else` into another branch.

The simple `@if (...)` -> `@else` rewrite remains fixable because it does not merge branch contents or change the number of roots in either branch.

### DIFF
--- a/packages/eslint-plugin-template/docs/rules/prefer-at-else.md
+++ b/packages/eslint-plugin-template/docs/rules/prefer-at-else.md
@@ -726,6 +726,113 @@ The rule does not have any configuration options.
 #### ❌ Invalid Code
 
 ```html
+<my-component>
+  @if (a) {
+    <button>1</button>
+  } @else {
+    <button>2</button>
+  }
+  @if (!a) {
+  ~~~~
+    <button>3</button>
+  }
+</my-component>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/prefer-at-else": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```html
+<my-component>
+  @if (a) {
+    <button>1</button>
+  }
+  @if (!a) {
+  ~~~~
+    <button>2</button>
+  } @else {
+    <button>3</button>
+  }
+</my-component>
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/prefer-at-else": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
+#### ❌ Invalid Code
+
+```html
+@if (a) {
+  <button>1</button>
+}
+@if (!a) {
+~~~~
+  <button>2</button>
+}
+@if (!a) {
+  <button>3</button>
+}
+```
+
+<br>
+
+---
+
+<br>
+
+#### Default Config
+
+```json
+{
+  "rules": {
+    "@angular-eslint/template/prefer-at-else": [
+      "error"
+    ]
+  }
+}
+```
+
+<br>
+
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks @splincode I've added the missing coverage and regenerated the rules docs (a requirement every time tests change), I'll have an agent keep an eye on CI as well and fix it up

JamesHenry (approved):

PR #3193, opened 2026-09-07
feat(eslint-plugin): [sort-keys-in-type-decorator] disallow unconfigured properties

### LINKED ISSUE #2494: [sort-keys-in-type-decorator] add config option to error on unexpected/unconfigured properties
The `sort-keys-in-type-decorator` rule doesn't force you to configure the order of all possible properties. I think that's great, as it allows for a lot of flexibility and not all projects will use all properties. However, in order to enforce true consistency, you must configure all properties that are used in your project. I would like to have a way to enforce that.

### Solution proposal

In my project, we had a custom lint rule that did the same thing as the `sort-keys-in-type-decorator` rule, but with the addition that if it encountered a property that wasn't listed in the configuration it would _also_ trigger a lint error for that property, indicating that it would need to be configured. Since we probably don't want this to be the default behavior, having a configuration option for triggering this error would be great.

### Alternative

There are other means of achieving this "true consistency". With the changes in https://github.com/angular-eslint/angular-eslint/commit/5e8c7ded02b5563b33d70b16234c5e5081b5ba58, you could fall back to the order in `DEFAULT_ORDER` for properties not listed in the custom configuration. That is less flexible than adding the config option, as order will then be enforced even when a property is not configured, but that may be acceptable (how likely is it actually that someone enables this rule with the intent to only enforce order of a subset of properties?).

CC'ing @bschaeublin as the original author of the lint rule (hope you don't mind)

splincode:
## Summary

- add an opt-in `allowUnconfiguredProperties` option to `sort-keys-in-type-decorator`
- default `allowUnconfiguredProperties` to `true` for backwards compatibility
- report decorator properties that are not listed in the configured order when `allowUnconfiguredProperties` is `false`
- add `Injectable` support with the built-in order `providedIn`, `useClass`, `useExisting`, `useFactory`, `useValue`, `deps`
- keep sorting configured properties while reporting unconfigured ones when they are disallowed

Fixes #2494

## Example

```js
'@angular-eslint/sort-keys-in-type-decorator': [
  'error',
  {
    Component: ['selector', 'imports', 'template', 'styles'],
    allowUnconfiguredProperties: false,
  },
]
```

With `allowUnconfiguredProperties: false`, any `@Component` property not listed above is reported so the configured order remains exhaustive and intentional. Decorators that are not explicitly configured still use their built-in default order.

## Injectable default behavior

This PR also enables default ordering for `@Injectable`. For example, `@Injectable({ useFactory: f, providedIn: 'root' })` is reported and autofixed to place `providedIn` before `useFactory`.


JamesHenry (changes requested):
strict is not a clear name, let's go with `allowUnconfiguredProperties` defaulting to `true` for compatibility

JamesHenry (changes requested):
Thanks for the update, the option name looks good now. A few things before this can be merged:

**1. `Injectable` default list is incomplete**

Angular's `@Injectable` accepts more than `providedIn`. It also accepts the provider keys: `useClass`, `useExisting`, `useFactory`, `useValue` and `deps`.

Because ESLint deep-merges user options with the defaults, a user who sets `allowUnconfiguredProperties: false` and only configures `Component` will still get the default `Injectable: ['providedIn']`. That means every `@Injectable({ providedIn: 'root', useFactory: ... })` in their project will be reported as an error.

Please change the default to include all six keys, with `providedIn` first.

**2. Test cases must live in `cases.ts`, not `spec.ts`**

All rules in this repo keep their test cases in `cases.ts`. The rule docs are generated from `cases.ts`, so right now the docs show the new option in the interface block but have no example of it in use.

Please move the new cases back into `cases.ts` and run `pnpm nx run-many -t update-rule-docs`, then commit the result.

**3. Missing test scenarios**

Please add tests for:

- `allowUnconfiguredProperties: false` where an unconfigured property exists **and** the configured properties are out of order. This should report both messages and still autofix the order.
- `allowUnconfiguredProperties: false` where the user only configures `Component`, and the file contains an `@Injectable` using `useFactory` (this covers point 1).
- Default options with `@Injectable({ useFactory: ..., providedIn: 'root' })`, to pin down the new default behavior (see point 4).

**4. Note the new default behavior in the PR description and docs**

On `main`, the `@Injectable` listener already exists, but it does nothing because there is no default order for it. With this PR, `@Injectable` is checked by default. For example, `@Injectable({ useFactory: f, providedIn: 'root' })` will now report and autofix `providedIn` to the front. That is fine, but please mention it in the PR description.

Also, please add one sentence to the rule docs explaining that when `allowUnconfiguredProperties` is `false`, decorators you did not configure are still checked against the built-in default order.

**5. Small cleanups (optional)**

- The `?? true` fallback in `create()` is not needed, `defaultOptions` already guarantees a value.
- `decoratorName` can be typed as `keyof OrderConfig` instead of casting at the lookup, since every call site passes a literal.
- Message wording suggestion: `Property "{{property}}" is not in the configured key order for @{{decorator}}.`

JamesHenry (approved):
Thank you!

PR #3197, opened 2026-09-08
feat(eslint-plugin-template): [prefer-style-binding] add bindUnits option

g-drouard:
## Overview

Adds an opt-in `bindUnits` option to `@angular-eslint/template/prefer-style-binding`.

When enabled, the rule additionally reports `[style.property]` bindings whose bound value embeds the CSS unit, and fixes them to Angular's `[style.property.unit]` form:

| Before | After (auto-fixed) |
| --- | --- |
| `[style.width]="'30px'"` | `[style.width.px]="30"` |
| `[style.margin-top]="'-12.5%'"` | `[style.margin-top.%]="-12.5"` |
| ``[style.height]="`${height()}em`"`` | `[style.height.em]="height()"` |

This keeps the unit out of the bound expression, so Angular no longer has to build a new string on every change detection cycle, and it lines up with the existing rationale of this rule, which already points users at `[style.prop.unit]` bindings when it reports `[ngStyle]` object keys with unit suffixes.

The option defaults to `false`, so existing users see no new reports.

## Details

- Style bindings are identified through `__originalType === BindingType.Style`, the same approach as `no-inline-styles`. `[style]`, `[class.x]` and other attributes are therefore out of scope, and the canonical `bind-style.width` form is handled too.
- Only two shapes are reported: a string literal made of a number immediately followed by a CSS unit, and a template literal interpolating exactly one expression immediately before a CSS unit. Anything else (`'auto'`, `'0 10px'`, ``` `${top}px ${left}px` ```, ``` `calc(${width}px)` ```, `width + 'px'`) is left alone.
- The fixer inserts `.unit` after the key span and replaces the value span, so the original attribute quoting is preserved.
- Bindings that already carry a unit (`[style.height.px]="h"`) are skipped.

The rule is now `fixable: 'code'`; rule docs and the README rule list have been regenerated.

## Test plan

- 42 cases in `tests/rules/prefer-style-binding/cases.ts`, covering both messages on the same element, the `bind-style` canonical form, and the negative cases listed above.
- `pnpm nx run-many -t lint typecheck` and the full `eslint-plugin-template` test suite pass.


JamesHenry (changes requested):
Thanks a lot, the option is a nice addition and the tests and docs look good. 

I got an agent to run the rule with `bindUnits: true` against a set of edge-case templates and found a few issues that need fixing before this can merge.

**1. The autofix breaks templates that contain HTML entities**

Angular parses the attribute value after decoding entities like `&quot;`, so the expression's source span offsets point into the decoded string, not the raw template. The fixer slices the raw template with those offsets and produces broken output.

Example:

```html
<div [style.width]="`${sizes[&quot;md&quot;]}px`"></div>
```

is fixed to:

```html
<div [style.width.px]="sizes[&quot"></div>
```

Suggested fix: before applying the fix, compare the raw slice of the value span with the expression's `source` text. If they differ, skip the fix (or skip the report).

**2. The rule never fires on elements with a structural directive**

```html
<div *ngIf="x" [style.width]="'30px'"></div>
```

produces no report, while the same element without `*ngIf` does.

Cause: Angular hoists the element's inputs onto the wrapping `Template` node using the same object instances. The template parser then visits each `BoundAttribute` twice and, on the second visit, overwrites `__originalType` with the string `'BoundAttribute'`. So the check `originalType === BindingType.Style` fails. `no-inline-styles` has the same blind spot.

Suggested fix: either guard the assignment in `packages/template-parser/src/index.ts` so `__originalType` is only set when the type is still a number, or detect style bindings from the key span text (`style.` / `bind-style.` prefix) instead of relying on `__originalType`.

**3. Duplicate reports on structural-directive elements**

Because of the same hoisting, the existing `ngStyle` handler already reports twice on `<div *ngIf="x" [ngStyle]="s">`. Once issue 2 is fixed, the new handler will do the same. Please dedupe, for example by tracking reported start offsets in a `Set`, or by scoping handlers to `Element > BoundAttribute`.

**4. Wrong property name in the message for CSS custom properties**

Angular rewrites `--gap` internally to `--%NS%gap`, and the message uses `node.name`, so users see:

```
Consider using [style.--%NS%gap.px] instead of embedding the unit in the bound value.
```

The fix output is correct because it uses the key span. Please derive the property name in the message from the key span too.

**5. Please reuse the existing literal helpers**

The new code checks `instanceof LiteralPrimitive` directly. The repo has `isStringLiteralPrimitive()` in `src/utils/literal-primitive.ts`, which was hardened in #3182 to work when more than one copy of `@angular-eslint/bundled-angular-compiler` is installed. With the direct `instanceof` check, `bindUnits` silently reports nothing in that setup.

It also skips `unwrapParenthesizedExpression()`, so `[style.width]="('30px')"` is not reported. Please use `isStringLiteralPrimitive(unwrapParenthesizedExpression(value))`.

**Things I checked that work fine:** `bind-style.x` syntax, single and double quotes, CRLF files, `@if` blocks, `ng-template`, multiple bindings on one element, pipes inside the interpolation, and the `ms` vs `s` unit matching. Docs and rule lists are correctly regenerated.

Please add test cases for issues 1 through 4 alongside the fixes.

g-drouard:
Thanks for the thorough review — all five points reproduced locally, and all five are fixed in 0e8d769. Test cases added for 1 through 4.

**1. HTML entities** — fixed. The fixer now compares the raw slice of the value span with the `ASTWithSource` source before using the expression's span; when they differ the report is kept but carries no fix. I went with dropping the fix rather than the report, since the binding is still worth flagging and the user can rewrite it by hand.

**2. Structural directives** — fixed, by taking the second option you offered: style bindings are now detected from `keySpan.details` (`style.` prefix) rather than `__originalType`. Note that this leaves the underlying parser bug in place, so `no-inline-styles` keeps the same blind spot. I'd rather not fold the `packages/template-parser` guard into this PR, since it changes what `no-inline-styles` reports and deserves its own tests — happy to open a separate PR for it if you'd like.

**3. Duplicate reports** — fixed with a `Set` of `messageId` + start offset rather than `Element > BoundAttribute`, so bindings on `<ng-template>` keep being reported. This also fixes the pre-existing double report on `<div *ngIf="x" [ngStyle]="s">`, which now has a test.

**4. CSS custom properties** — fixed by the same change as 2: the message derives the property from `keySpan.details`, so it reads `[style.--gap.px]` instead of `[style.--%NS%gap.px]`.

**5. Literal helpers** — now using `isStringLiteralPrimitive(unwrapParenthesizedExpression(value))`, so `[style.width]="('30px')"` is reported. I added an `isTemplateLiteral()` alongside it in `src/utils/literal-primitive.ts`, following the same `type` / `instanceof` pattern, because the template-literal branch had the identical multi-copy problem.

Also merged `main` in. `pnpm nx run-many -t lint typecheck` and the full `eslint-plugin-template` suite (1382 tests) pass locally.

JamesHenry:
@g-drouard Thanks for the updates, the earlier issues are all resolved. A few remaining items before this can merge.

**1. The fix is lost when the bound value has leading whitespace**

The verbatim-source check and the fixer both slice the template from `valueSpan.start.offset`. Angular computes expression offsets from `valueSpan.fullStart.offset`, which includes leading whitespace. When they differ, the check fails and the fix is dropped.

```html
<div [style.width]="  `${w}px`  "></div>   <!-- reported, but no fix -->
<div [style.width]="  '30px'  "></div>     <!-- fixed to "  30", trailing space lost -->
```

Please use `valueSpan.fullStart.offset` in both places (around lines 246 and 271 of `prefer-style-binding.ts`) and add a test for a value with leading and trailing whitespace.

**2. Interpolation syntax is not detected**

These two lines are the same style binding, but only the first is reported:

```html
<div [style.width]="`${w}px`"></div>   <!-- reported -->
<div style.width="{{w}}px"></div>      <!-- not reported -->
```

Add an `Interpolation` branch to `getUnitBinding` so both forms are handled.

**3. The unit regex is case-sensitive and rejects some valid numbers**

CSS units are case-insensitive. These are valid but not reported:

```html
<div [style.width]="'30PX'"></div>
<div [style.width]="'.5em'"></div>
<div [style.width]="'+30px'"></div>
```

Add the `i` flag to `NUMBER_WITH_UNIT`, lowercase the unit in the fix output, and allow a leading `.` or `+`.

**4. Run the dedupe check before the analysis**

`getUnitBinding` runs before `isFirstReportFor`, so bindings on elements with a structural directive are analysed twice. Move the dedupe check first. Also, the `BoundAttribute` listener is registered even when `bindUnits` is false. Please only register it when the option is enabled.

**Not required in this PR**

The `alreadyReported` Set works around a parser-wide double visit that affects every rule with a `BoundAttribute` handler. That belongs in a separate `template-parser` fix. Leave the Set in for now and I will follow up.

Please add appropriate test cases for the above points

g-drouard:
Thanks — all four are fixed in 3665b0b, with tests for each.

**1. Leading whitespace** — fixed. Both the verbatim check and the fixer now use `valueSpan.fullStart.offset`. I confirmed the asymmetry you described: for `[style.width]="  '30px'  "`, `start` is 22 and `fullStart` is 20, and `source` is `  '30px'  `, so only the slice from `fullStart` matches it. Replacing `[fullStart, end]` also drops the padding rather than half of it, so the output is `[style.width.px]="30"`. Tests added for the static and template-literal forms.

**2. Interpolation** — added. `Interpolation` exposes `strings`/`expressions` with the same shape as `TemplateLiteral`'s `elements`/`expressions`, so the suffix check is now shared in `getTrailingUnit()`. The fix keeps the interpolation rather than converting the binding syntax, so `style.width="{{w}}px"` becomes `style.width.px="{{w}}"`. It is produced by trimming the unit off the verbatim value, which preserves inner spacing — `style.height="{{ height() }}em"` becomes `style.height.em="{{ height() }}"`. Valid cases added for two expressions and for text before the interpolation.

**3. Unit regex** — fixed. The `i` flag is on `NUMBER_WITH_UNIT` and on `UNIT_ONLY`, so uppercase units are matched in the template-literal and interpolation forms too, and the unit is lowercased once at extraction so the message and the fix agree. The number pattern is now `[+-]?(?:\d+(?:\.\d+)?|\.\d+)`. I deliberately left the value itself verbatim, so `'+30px'` gives `[style.width.px]="+30"` rather than normalising to `30`.

**4. Ordering and registration** — fixed. `isFirstReportFor()` is the first statement in the handler, and the handler is only spread into the returned listeners when `bindUnits` is true.

Noted on the parser-wide double visit — the `Set` stays, and I'll leave that follow-up to you.

One thing worth flagging: `'2kHz'` now fixes to `[style.x.khz]="2"`. That follows your "lowercase the unit in the fix output" note and is fine for CSS, but say the word if you would rather keep the authored casing.

PR #3201, opened 2026-09-09
fix(eslint-plugin): [reactive-context-must-read-signal] ignore inherited primitive names

### LINKED ISSUE #3198: [reactive-context-must-read-signal] Crash "config.args is not iterable" when a destructured callback param shadows an Object.prototype property (e.g. valueOf)

This only reproduces when using the experimental **Signal Forms** APIs (`@angular/forms/signals`, i.e. `form()` / `disabled()` / `required()` / etc.).

The rule crashes with `TypeError: config.args is not iterable` when linting a file that contains a function call whose callee is a bare identifier matching an inherited `Object.prototype` property name (e.g. `valueOf`, `toString`, `constructor`, `hasOwnProperty`) — for example a destructured parameter named `valueOf` that is later invoked as a function inside a Signal Forms schema:

```ts
import { Component } from '@angular/core';
import { disabled, form } from '@angular/forms/signals';

interface Task {
 id: number;
 archived: boolean;
}

@Component({ selector: 'app-repro', template: '' })
export class ReproComponent {
 private readonly taskSchema = (path: any) => {
   disabled(path.archived, { when: ({ valueOf }) => valueOf(path.id) > 0 });
 };
 protected readonly taskForm = form({} as Task, this.taskSchema);
}
```

I don't have an online reproduction (e.g. StackBlitz) yet — the snippet above is a minimal, self-contained repro extracted from my codebase and should be enough to trigger the crash when linted with `@angular-eslint/reactive-context-must-read-signal` enabled.

**Root cause:** in `reactive-context-must-read-signal.js`, the primitives lookup is a plain object property access:

```js
const config = primitives[node.callee.name];
if (!config) return;
```

`primitives` is a plain `{}` object literal, so it inherits `Object.prototype`. When `node.callee.name` is `"valueOf"` (or `toString`/`constructor`/etc.), `primitives['valueOf']` resolves to the inherited `Object.prototype.valueOf` function instead of `undefined`. That value is truthy, so the `!config` guard is bypassed, and `registerTrackedFunctions` then does `for (const index of config.args)` where `config.args` is `undefined`, throwing `TypeError: config.args is not iterable`.

**Suggested fix:** use `Object.hasOwn(primitives, node.callee.name)` (or build `primitives` via `Object.create(null)` / a `Map`) before indexing.

**Versions**

| package                         | version    |
| -------------------------------- | ---------- |
| `@angular-eslint/eslint-plugin`  | `22.5.0`   |
| `@typescript-eslint/parser`      | `8.70.0`   |
| `ESLint`                         | `10.10.0`  |
| `node`                           | `v24.19.0` |

```sh


```

splincode:
## Summary

- guard reactive primitive lookup with `Object.hasOwn()`
- avoid treating inherited `Object.prototype` members such as `valueOf`, `toString`, `constructor`, and `hasOwnProperty` as configured primitives
- add a regression test covering bare calls with those names

Fixes #3198


packages/eslint-plugin/tests/rules/reactive-context-must-read-signal/spec.ts
```diff
@@ -15,6 +15,27 @@ const ruleTester = new RuleTester({
 });
 
 ruleTester.run(RULE_NAME, rule, {
-  valid,
+  valid: [
+    ...valid,
+    // Regression for #3198: inherited Object.prototype names are not primitives.
```
JamesHenry: As mentioned on your previous PR, please stop inlining tests into this file, there is no precedent for this in the entire codebase, we generate docs from our tests intentionally

JamesHenry (changes requested):

packages/eslint-plugin/tests/rules/reactive-context-must-read-signal/spec.ts
splincode: Thanks, fixed. I moved the regression case into `cases.ts` and regenerated the rule docs.

JamesHenry (approved):

JamesHenry:
Thank you!

PR #3221, opened 2026-09-17
feat: add support for extracting inline styles in Angular components

### LINKED ISSUE #3215: Feature: Add plugin to support linting Angular (S)CSS with eslint
I'd like to be able to lint my SASS in my Angular projects without needing to install/manage stylelint separate from ESLint.

ESLint supports:
* css linting: `@eslint/css`
* scss syntax: `@humanwhocodes/scsstree`

However, this configuration still misses Angular-specific linting of:
* `styles` defined on the component in the TS file
* `style` attributes in the inline `template` on the component in the TS file
* `style` attributes in the component's HTML file

Is a preset that preconfigures all of this plumbing something like this "in scope" for the `angular-eslint` project or is it better suited for a separate package?

NateRadebaugh:
_**AI disclosure:** authored using GPT-5.6-Luna (xhigh) and code reviewed with GPT-6 Astra (low) for fixes._

---
Closes: #3215


JamesHenry (changes requested):
Thanks for picking this up so quickly, the overall shape is exactly what I had in mind: a processor only, no `@eslint/css` wiring, exposed as `processInlineStyles`.

A few things need to change before this can land.

**1. Performance: every `.ts` file gets fully parsed**

`preprocessInlineStyles` calls `preprocessComponentFile` (which is protected by the `isFileLikelyToContainComponentDeclarations` heuristic), but then does a second, unconditional `ts.createSourceFile` for style extraction. The README tells users to apply the processor to `**/*.ts`, so every utility and service file in a project pays for a full TypeScript parse.

On a 20k-line non-Angular file this is roughly 150 ms per file versus effectively zero for the existing `extract-inline-html` processor.

Fix: check the heuristic before parsing, and parse once. I pushed a working version of this on top of your branch here: https://github.com/angular-eslint/angular-eslint/commit/1733c72. It adds an optional `parsedSourceFile` parameter to `preprocessComponentFile` so the two extractors share one `SourceFile`. Feel free to cherry-pick it or rework it as you see fit.

**2. `postprocess` can throw and abort linting of the file**

`mapStyleMessage` calls `virtualSource.getPositionOfLineAndCharacter`, which hard-fails on an out-of-range line. If a CSS rule or the CSS parser reports a location past the end of the virtual file, the user gets `Debug Failure. Bad line number` instead of lint output. The commit above wraps this in a try/catch and leaves the location unmapped instead.

**3. The `.html` dedupe depends on an ESLint internal**

The logic that strips the `\d+_` prefix and uses `dirname(filename)` to find the parent `.ts` file mirrors how `processor-service.js` names code blocks internally (`${i}_${block.filename}`). I checked and the recursion does happen, so the dedupe is needed and works, but please add a comment explaining what it relies on. If ESLint changes that naming, this silently turns into double reports.

**4. Interpolations in style attributes must be skipped, not replaced with 0**

extractTemplateStyles substitutes 0 for each {{ }} interpolation. That only produces valid CSS when the interpolation is a bare number inside a longer value. For style="color: {{ c }}" it yields color: 0 and triggers invalid-property rules. For style="{{ dynamicStyles }}" or style="{{ prop }}: 1px" it yields a hard CSS parse error. None of these can be suppressed, because a disable comment cannot be placed inside a style attribute and comments in the surrounding HTML or TypeScript are not carried into the virtual CSS file. Any fix touching the substituted region is also dropped by mapStyleMessage.

Please skip any style attribute whose value contains an interpolation, the same way [style] and [style.x] bindings are already skipped because they are expressions rather than static CSS. State in the README that only static style attributes and static styles entries are linted.

**5. Docs**

`docs/CONFIGURING_ESLINT.md` is where `processInlineTemplates` is explained to users. The new processor, and especially the "use this instead of `processInlineTemplates`" guidance, should be documented there as well, not only in `packages/angular-eslint/src/configs/README.md`.

**6. Hard-coded .scss for component styles is a blocker**

Every styles entry is emitted as inline-style-N.scss. Two problems:

It is the wrong default. Angular's inlineStyleLanguage option defaults to css, so for most projects inline styles are plain CSS. Labelling them .scss means a user with a normal files: ['**/*.css'] block for @eslint/css gets no component style linting at all, with no error or warning. A processor that silently does nothing is the worst failure mode for a lint tool.

It cannot be configured. The processor is a static object, so there is no way for a project that genuinely uses SCSS inline styles to opt in, and no way for a CSS project to opt out.

The language of inline styles is a per-project setting, so the processor needs to take it as an option. Suggested shape:

// default: css, matching Angular's inlineStyleLanguage default
processor: angular.processInlineStyles,

// explicit
processor: angular.processInlineStyles.withOptions({ inlineStyleLanguage: 'scss' }),

Implementation-wise that is a small factory in processors.ts that closes over the extension, with the default export being the css instance. style="..." attributes extracted from templates are always CSS and should stay .css regardless of this option. The README example should then show both the default and the SCSS variant.

Misc others:

- Virtual style filenames (`inline-style-0.scss`) do not include the component filename, unlike `inline-template-<file>-<n>.component.html`.
- Fixes containing newlines or quotes are dropped even inside backtick literals where they would be safe, and it is undocumented.
- The unit test uses a stub parser and a fake rule. That is fine for the mapping logic, but nothing exercises real `@eslint/css` output. An e2e fixture would be the right place for this.

NateRadebaugh:
Thank you for this detailed review and updated guidance. I'll regroup on this next week and propose a followup then.

